PR #1, opened 2019-12-02
Take readiness probes into account

ytsarev:
Liveness probes might not be enough for GSLB scenario, we need
to be sure that the application is ready to accept the traffic after
some time after the start(java warm up is a classic example of that).

Thus, for reliable load balancing we should check for readiness
probe status and also strongly advise application team to configure one
for their workload(s)


ytsarev:
@donovanmuller can't assign reviewers so pinging you here
@chriskieser could you please tweak access rights a bit ?

PR #13, opened 2019-12-18
Make OhMyGlb operator watch all namespaces for Gslb CRs

ytsarev:
* Operator is cluster scope
* Gslb CRDs are namespaced

### DIFF
--- a/deploy/coredns/README.md
+++ b/deploy/coredns/README.md
@@ -1,5 +1,5 @@
 # CoreDNS helm install to act as resolver for GSLB
 
 ```
-helm upgrade gslb-coredns -f deploy/coredns/values.yaml
+helm -n ohmyglb upgrade -i gslb-coredns stable/coredns -f values.yaml
 ```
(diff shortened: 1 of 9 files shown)


ytsarev:
Before we operated default namespaced scenario of operator-sdk: operator was managing resources in the same namespace as own deployment. OhMyGlb needs to watch all namespaces as there can be multiple application teams in different namespaces managing own `Gslb` resources.
Excerpt from operator log after this change:
```
{"level":"info","ts":1576684481.7743516,"logger":"controller-runtime.controller","msg":"Starting workers","controller":"gslb-controller","worker count":1}
{"level":"info","ts":1576684481.77468,"logger":"controller_gslb","msg":"Reconciling Gslb","Request.Namespace":"default","Request.Name":"example-gslb"}
{"level":"info","ts":1576684482.086346,"logger":"controller_gslb","msg":"Reconciling Gslb","Request.Namespace":"test","Request.Name":"example-gslb"}
```
It's visible that `example-gslb` got reconciled in both `default` and `test` namespaces

PR #19, opened 2020-01-03
Deprecate coreDNS hosts config and worker health checks

ytsarev:
* Deprecate direct CoreDNS configmap modification and `hosts` plugin
* Deprecate k8s worker nodes health checks as we are relying in Ingress
  Address status now


donovanmuller:
Always nice to see more code deleted than added 👍

PR #26, opened 2020-01-14
Ohmyglb operator chart

ytsarev:
* Create ohmyglb chart including dependencies

Motivation

* Control operator and subchart config variables in
  a single place
* Handle use cases like image location in private registry
* More streamlined installation


ytsarev:
@donovanmuller please provide early feedback. The PR requires more testing and love before the merge.

chart/ohmyglb/Chart.yaml
```diff
@@ -0,0 +1,29 @@
...
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application.
+appVersion: 1.16.0
```
donovanmuller: Would this be reset to reflect the release version when released?

chart/ohmyglb/values.yaml
```diff
@@ -0,0 +1,65 @@
+global:
+  ## Reference to one or more secrets to be used when pulling images
+  ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/
+  ##
+  imagePullSecrets: []
+  # - name: "image-pull-secret"
+
+ohmyglb:
+  image: ytsarev/ohmyglb:v0.2.1
```
donovanmuller: Similar to Helm version, reset on first release?

chart/ohmyglb/values.yaml
```diff
@@ -0,0 +1,65 @@
...
+      parameters: cluster.local
+    - name: loadbalance
+      parameters: round_robin
+    # Serves a /metrics endpoint on :9153, required for serviceMonitor
+    - name: prometheus
+      parameters: 0.0.0.0:9153
+    - name: forward
+      parameters: . /etc/resolv.conf
+    - name: etcd
+      parameters: absa.internal
```
donovanmuller: For future perhaps but refrain from using Absa in any configuration defaults

deploy/coredns/external-dns.yaml
```diff
@@ -41,6 +41,8 @@ kind: ServiceAccount
 metadata:
   name: external-dns
   namespace: ohmyglb
+imagePullSecrets:
+- name: regcred
```
donovanmuller: Would a pull secret be a Chart option that defaults to none. I.e. In general use perhaps image pull secrets would not be necessary

ytsarev: @donovanmuller that's clearly a bug, thanks for the catch

chart/ohmyglb/values.yaml
ytsarev: @donovanmuller `example.com` ?

chart/ohmyglb/Chart.yaml
ytsarev: @donovanmuller argh, this number was automatically generated by `helm create..`

chart/ohmyglb/values.yaml
ytsarev: @donovanmuller we can, but reset to what ? `v0.2.1` btw reflects semantics of actual development

chart/ohmyglb/values.yaml
donovanmuller: Yeah 👍

PR #36, opened 2020-01-27
Initial Edge DNS support 

ytsarev:
* Dedicated DNSEndpoint to manage A records for NS servers pointing
  to GSLB
* Initial Infoblox support


ytsarev:
@donovanmuller this is work in progress. I'm using term 'Edge' for the stuff like Infoblox, Route53 or any dns solution that will externally host delegation to gslb controlled zone and associated A records. Not sure if there is a better established term for that, open to suggestions

PR #56, opened 2020-03-01
Config and helpers for local multicluster setup

ytsarev:
* Make targets to create/destroy 2 local kind clusters
* hostAlias support to enable overriding of external cluster
  fqdn resolution and point to local ip address within docker network
* Enables testing ohmyglb locally with multicluster setup including
  communication between ohmyglb instances
* Especially useful to test and develop load balancing strategies
  avoiding heavy external setup

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,11 +35,22 @@ dns-smoke-test:
 
 .PHONY: deploy-local-cluster
 deploy-local-cluster:
-	kind create cluster --config=deploy/kind/cluster.yaml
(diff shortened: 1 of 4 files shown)


ytsarev:
@donovanmuller btw was seriously evaluating https://github.com/rancher/k3d/ today as alternative to kind. It is working well and much faster than kind, especially for the topic of local multicluster thing.
Unfortunately https://github.com/rancher/k3d/issues/111 and we need easy cross-cluster communication

PR #58, opened 2020-03-06
Infoblox update

ytsarev:
It's a thousand of unpredictable from the vendor. With capital V


ytsarev:
@donovanmuller yeah, initially I was trying to make it as generic as possible.
But it crashed over two points:
* NS and associated A records are getting created automatically by infoblox
* TXT record just does not work with externaldns-infoblox

So I decided to get rid of externaldns-infoblox completely and make direct api kick with ibclient for TXT record ( and it works reliably) . 

Looks like for every edgeDNS we will have to treat special API call and it is unfortunately unavoidable. Obviously , happy to discuss the thing deeper

PR #79, opened 2020-04-07
Terratest e2e for Failover strategy

ytsarev:
* Utilizes 2 local clusters deployed by `make deploy-full-local-setup`
* Creates `Gslb` CRs in both clusters with Failover strategy
* Checks expected DNS response on the main cluster
* Makes workload unhealthy on the main cluster
* Checks that DNS response contains IP list from secondary cluster
  proving that Failover works

### DIFF
--- a/terratest/examples/failover.yaml
+++ b/terratest/examples/failover.yaml
@@ -1,0 +1,17 @@
+apiVersion: ohmyglb.absa.oss/v1beta1
+kind: Gslb
+metadata:
+  name: test-gslb
+spec:
+  ingress:
+    rules:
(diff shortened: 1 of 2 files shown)


donovanmuller:
@ytsarev Correct me if I'm wrong but these e2e tests are not yet added to the build pipeline yet?

ytsarev:
@donovanmuller not yet... I'm slowly crawling to prove https://github.com/AbsaOSS/ohmyglb/issues/67 programmable way :) We will include all of them into the pipeline as part of https://github.com/AbsaOSS/ohmyglb/issues/48 and 0.7 release

PR #84, opened 2020-04-08
Test mutual failover setup

### LINKED ISSUE #67: When using the failover load balancing strategy, investigate and validate how resolution will be handled effectively when clusters are configured for mutual failover
Related to this note on #46:

> The existence of multiple "secondary" failover clusters should also be considered. For example, if there were 3 clusters (X, Y and Z) in the scenario 2 above, could the Ingress node IPs for both clusters (X and Z) be resolved and if so, how (in terms of "load balancing") would the Ingress node IPs across both those secondary/failover clusters be resolved? Would they use the default round robin strategy, if any strategy at all?

we need to validate that the IPs resolved from the secondary/failover _clusters_ are handled correctlty and consistently in line with a load balancing strategy of their own.

ytsarev:
* Resolves #67
* Proves expected behavior of primary-primary failover setup
* Move common test function to `helpers.go`

```
--- PASS: TestOhmyglbSplitFailoverExample (230.45s)
    --- PASS: TestOhmyglbSplitFailoverExample/Each_cluster_resolves_its_own_set_of_IP_addresses (30.53s)
    --- PASS: TestOhmyglbSplitFailoverExample/serviceHealth_becomes_Unhealthy_after_scaling_down_to_0 (10.17s)
    --- PASS: TestOhmyglbSplitFailoverExample/Cluster_1_failovers_to_Cluster_2 (61.28s)
    --- PASS: TestOhmyglbSplitFailoverExample/Cluster_2_still_returns_own_entries (0.07s)
```

### DIFF
--- a/terratest/test/ohmyglb_basic_split_failover_test.go
+++ b/terratest/test/ohmyglb_basic_split_failover_test.go
@@ -1,0 +1,115 @@
+package test
+
+import (
+	"fmt"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/gruntwork-io/terratest/modules/k8s"
+	"github.com/gruntwork-io/terratest/modules/random"
+)
+
+// Basic ohmyglb deployment test that is verifying that associated ingress is getting created
+// Relies on two local clusters deployed by `$make deploy-two-local-clusters`
+// Tests expected behavior for https://github.com/AbsaOSS/ohmyglb/issues/67
+func TestOhmyglbSplitFailoverExample(t *testing.T) {
+	t.Parallel()
+
+	// Path to the Kubernetes resource config we will test
+	kubeResourcePath1, err := filepath.Abs("../examples/failover.yaml")
+	require.NoError(t, err)
+
+	kubeResourcePath2, err := filepath.Abs("../examples/failover2.yaml")
+	require.NoError(t, err)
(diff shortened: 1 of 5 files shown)


terratest/test/ohmyglb_basic_split_failover_test.go
```diff
@@ -0,0 +1,115 @@
...
+			300,
+			1*time.Second,
+			func() ([]string, error) { return Dig(t, "localhost", 53, "terratest-failover.cloud.example.com") },
+			expectedIPsCluster2)
+		require.NoError(t, err)
+
+		assert.Equal(t, afterFailoverResponse, expectedIPsCluster2)
+	})
+
+	t.Run("Cluster 2 still returns own entries", func(t *testing.T) {
```
donovanmuller, line 101: @ytsarev would it be worth adding an additional step. Test if the service from Cluster 1 becomes healthy once more, that  Cluster 1 should then continue to resolve its own IP addresses? I.e. it fails back again.

So:

1. Cluster 1 resolves it's own addresses
2. Service becomes unhealthy on Cluster 1
3. Cluster 2 addresses are now resolved only
4. Service becomes healthy again on Cluster 1
5. Cluster 1 once again resolves it's own addresses

ytsarev: totally, will add this step, thanks!

ytsarev: @donovanmuller implemented, please review

PR #91, opened 2020-04-17
Terratest based end-to-end pipeline 

### LINKED ISSUE #48: Add full end to end integration tests to build pipeline
Existing build pipelines should be enhanced with full end to end integration tests running in managed Kubernetes clusters.

The goal of these tests would be to simulate:

* real world scenarios deploying to multiple managed Kubernete clusters (potentially across different clouds)
* exercise tests of the various load balancing strategies
* simulate the various failure scenarios including split brain examples
* verify the edge DNS integrations across the supported providers

These tests should be triggered at strategic and meaningful points in the development flow, considering that this test suite could take considerable time to run to completion.

ytsarev:
* Resolves #48

* Utilizes kind action from helm org

* Rebind host dns ports to 5053 and 5054
  to avoid conflicts with running DNS server
  Must have for Github Actions, more safe
  for local scenarion as well.

* Include stable repo to nginx-ingress install

* Adjust hostAlias to non-standard dns port scenario

* Reduce number of kind workers for terratest

* Make hostAlias ip resolution dynamic

* Optimize assertGslbStatus with smarter waiting

* Fix parallel terratests for parallel execution

* Avoid ingress same name conflic for failover tests

* Enhance app deployment function

* Enable all parallel terratest run in the pipeline

### DIFF
--- a/deploy/kind/cluster2.yaml
+++ b/deploy/kind/cluster2.yaml
@@ -13,5 +13,5 @@ nodes:
     hostPort: 444
     protocol: TCP
   - containerPort: 53
-    hostPort: 54
+    hostPort: 5054
     protocol: UDP
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -1,5 +1,5 @@
-on: push
-name: build
+name: Golang lint and test
+on: [push, pull_request]
 jobs:
   go-test:
     name: Go Test
--- a/.github/workflows/terratest.yaml
+++ b/.github/workflows/terratest.yaml
@@ -1,6 +1,6 @@
 name: Terratest
 
-on: pull_request
+on: [push, pull_request]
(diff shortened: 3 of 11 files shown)


deploy/kind/cluster2.yaml
```diff
@@ -13,5 +13,5 @@ nodes:
     hostPort: 444
     protocol: TCP
   - containerPort: 53
-    hostPort: 54
+    hostPort: 5054
```
kuritka, line 16: I must update this in Readme

```shell script
dig @localhost localtargets.app3.cloud.example.com -p 5053 && dig -p 5054 @localhost localtargets.app3.cloud.example.com
```

ytsarev: yeah, good point

Issue #47, opened 2020-02-24
Expose metrics and tracing


donovanmuller:
Implement relevant metrics that provide insight into:

* Edge DNS integrations
* Split brain scenarios / heartbeat
* Reconciliation
* Load balancing strategies
* etc.

Metrics should be expressed as Prometheus [text exposition format](https://github.com/prometheus/docs/blob/master/content/docs/instrumenting/exposition_formats.md).

Consider the following CoreDNS plugins as reference:

* https://coredns.io/plugins/metrics/
* https://coredns.io/plugins/trace/

somaritane:
https://github.com/AbsaOSS/ohmyglb/pull/97 addresses generation of initial scope of prometheus metrics for the project.
Metrics are exposed at prometheus endpoints, and can scraped with 3rd party tools, as well as [Prometheus Operator](https://github.com/coreos/prometheus-operator).
Since the scope of this issue is really wide, I suggest to chop it to smaller pieces (additional metrics, tracing support) and resolve the original, considering #97 is merged.

PR #97, opened 2020-04-23
Initial operator metrics

somaritane:
- Partially addresses #47 with initial creation of OhMyGLB specific metrics.
- Following custom metrics are added: 
   - healthy_records
   - managed_hosts
- Documentation updated in order to cover metric types exposed by operator (extended `controller-runtime` metrics and `kube-state-metrics`)


donovanmuller:
@somaritane Would it be possible to provide a simple sample Grafana dashboard showing off the existing metrics with a complimentary visualisation?

Makefile
```diff
@@ -11,8 +11,8 @@ OHMYGLB_IMAGE ?= absaoss/ohmyglb:$(VERSION)
 .PHONY: up-local
 up-local: create-test-ns
 	kubectl apply -f ./deploy/crds/ohmyglb.absa.oss_gslbs_crd.yaml
-	kubectl apply -f deploy/crds/ohmyglb.absa.oss_v1beta1_gslb_cr.yaml
-	operator-sdk up local --namespace=test-gslb
+	kubectl apply -f ./deploy/crds/ohmyglb.absa.oss_v1beta1_gslb_cr.yaml
+	operator-sdk run --local --namespace=test-gslb --enable-delve
```
ytsarev, line 15: @somaritane how was the experience with Delve enablement ?

Makefile
```diff
@@ -137,11 +137,11 @@ clean-test-apps:
 
 .PHONY: build
 build:
-	operator-sdk build $(REPO):$(VERSION)
+	operator-sdk build $(OHMYGLB_IMAGE)
```
ytsarev, line 140: Good catch

Makefile
somaritane: > @somaritane Would it be possible to provide a simple sample Grafana dashboard showing off the existing metrics with a complimentary visualisation?

@donovanmuller Good point, I'll try to publish one, maybe with sample local snapshot.

pkg/controller/gslb/status.go
```diff
@@ -128,3 +140,31 @@ func (r *ReconcileGslb) getHealthyRecords(gslb *ohmyglbv1beta1.Gslb) (map[string
 
 	return healthyRecords, nil
 }
+
+func (r *ReconcileGslb) updateManagedHostsTotalMetric(gslb *ohmyglbv1beta1.Gslb, serviceHealth map[string]string) error {
```
ytsarev, line 144: @somaritane there is quite amount of logic in metric functions, what about unit tests ?

Makefile
somaritane: > @somaritane how was the experience with Delve enablement ?

@ytsarev, no issues, helped me to debug some corner cases.
Actually, you need to have delve installed for that, so I'll probably make it optional, and update the docs for that.

pkg/controller/gslb/status.go
somaritane: @ytsarev 
What is our take on unit-tests for internal (unexported) functions, which are not the part of API/contract? 
I usually take pragmatic approach and trying to avoid this, unless the logic gets really complex (which on other hand is the sign, that there’s a need of some refactoring), and usually unit testing for non-public functions/methods is considered unwanted practice.
So I see following options here:
Make controller metrics a separate package, and export the functions, then cover them by tests. In the future, as we get more complex instrumentation added for metrics in different packages, there will be more satellite metric packages involved.
Cover metrics by integration/e2e tests instead of unit tests
Write unit tests for internal package functions, but decide about balance: should we cover all internal functions  with tests, how much complex is complex etc :slightly_smiling_face:

kuritka: Hi, personally I'd like to go with metrics in separate package as you proposed in 1. , to keep separation of contrac,  better testability and maintainibility.

ytsarev: @somaritane 
Not testing unexported functions is arguable. Some people say that the fact that it's the power of golang testing package to be able to test internal functions.
Anyhow in our specific use case most of the functions we use inside Reconcile loop are unexported (for a good reason) . Not testing them would affect quality quite dramatically .
In the meantime the approach to testing is something in between of point 2,3 you provided.

If you looks at our existing unit test suite at https://github.com/AbsaOSS/ohmyglb/blob/master/pkg/controller/gslb/gslb_controller_test.go we stick to roughly following pattern:

* Create special condition under test
* Invoke Gslb reconsiliation loop
* Check the result from outside perspective ( e.g. controlled Ingress was modified, Gslb status got changed in expected way etc)

So we usually do not fine-grain unit test this functions ( until it is unavoidable) rather checking overall result of reconciliation loop with conditions associated with the function under test.

So I would suggest to extend above mentioned test suite with this scenario:

* Create Gslb resource(s) with conditions that about to be reflected in metrics
* Invoke reconciliation loop with `reconcileAndUpdateGslb(t, r, req, cl, gslb)`
* Check if associated metrics are getting exposed as expected

This way we can have a confidence in the metric functions we develop meanwhile avoiding reliance exclusively on very heavy e2e tests.

somaritane: @ytsarev I've added the missing tests (had to remove tight dependency on previous fixture state for some other tests, as I had to restore test conditions there). I also renamed metrics in order to comply to metrics linter, provided with prometheus testutils package.

README.md
```diff
@@ -249,3 +250,74 @@ Optionally you can switch GLSB back to round-robin mode
...
+Number of managed hosts observed by OhMyGLB.
+
+Example:
+
+```yaml
+# HELP ohmyglb_gslb_managed_hosts Number of managed hosts observed by OhMyGLB.
+# TYPE ohmyglb_gslb_managed_hosts gauge
+ohmyglb_gslb_managed_hosts{name="test-gslb",namespace="test-gslb",status="Healthy"} 1
+ohmyglb_gslb_managed_hosts{name="test-gslb",namespace="test-gslb",status="NotFound"} 1
+ohmyglb_gslb_managed_hosts{name="test-gslb",namespace="test-gslb",status="Unhealthy"} 2
```
ytsarev, line 301: @somaritane a little bit confused here. Number of managed hosts is orthogonal to status(unless managed hosts are down to 0)  or do I miss something ?

somaritane: The metric is derived from ServiceHealth, it represents the amount of services with particular status, probably the naming is not correct, happy to consider better options.

ytsarev: So currently Gslb Status looks like
```
  status:
    healthyRecords:
      app3.cloud.example.com:
      - 172.17.0.4
      - 172.17.0.5
      - 172.17.0.6
    managedHosts:
    - app1.cloud.example.com
    - app2.cloud.example.com
    - app3.cloud.example.com
    serviceHealth:
      app1.cloud.example.com: NotFound
      app2.cloud.example.com: Unhealthy
      app3.cloud.example.com: Healthy
```
So managed hosts == ingress host ohmyglb operates and serviceHealth is associated status... I will think of combined depicting name, seems to be not easy :)

somaritane: Yes, so we calculate how many ingress hosts are with certain status, and record this number  with "status" label accordingly. This can be aggregated in query to give amount of all ingress hosts within gslb etc.

somaritane: Just a side not: Looking at serviceHealth and managedHosts status properties, managedHosts looks kind of redundant, as from serviceHealth it is clear how many managed ingress hosts we have.

ytsarev: @somaritane `ohmyglb_gslb_ingress_hosts_per_status_total` ? Inspired by https://prometheus.io/docs/practices/naming/

re `Status` content it needs to be streamlined I have it in mind for a while but never created an Issue - will do

somaritane: Looks good, just without `_total` suffix, as this metric is not accumulating count, it is [gauge](https://prometheus.io/docs/concepts/metric_types/#gauge), whose value can arbitrarily go up and down, metrics linter fails in case if `_total` suffix is used when naming Gauge and derived metric types.

somaritane: The metric is renamed to `ohmyglb_gslb_ingress_hosts_per_status`

pkg/controller/gslb/metrics.go
```diff
@@ -0,0 +1,72 @@
...
+		[]string{"namespace", "name", "status"},
+	)
+)
+
+func (r *ReconcileGslb) updateManagedHostsTotalMetric(gslb *ohmyglbv1beta1.Gslb, serviceHealth map[string]string) error {
+	var healthyTotal, unhealthyTotal, notFoundTotal int
+	for _, hs := range serviceHealth {
+		switch hs {
+		case healthyStatus:
+			healthyTotal++
```
ytsarev, line 46: @somaritane just a note that internal variable do not comply with upstream Total suffix semantics, do I get it right ?

somaritane: Well spotted, @ytsarev! Makes sense to rename method as well.

pkg/controller/gslb/gslb_controller_test.go
```diff
@@ -525,6 +627,126 @@ func TestGslbController(t *testing.T) {
 	})
 }
 
+func createHealthyService(t *testing.T, serviceName string, cl client.Client, gslb *ohmyglbv1beta1.Gslb) {
+
```
ytsarev, line 631: Please add `t.Helper()` so the stacktrace will report specific failing test context, not just function limited scope.

See https://golang.org/pkg/testing/#T.Helper

PR #125, opened 2020-05-16
Extend release pipeline with docker build and push

ytsarev:
* Build and push using version for `version.go`
* Forgotten v0.6.0 version bump

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -10,6 +10,18 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
+      - name: Operator image build and push
+        env:
+          OPERATOR_SDK_VERSION: v0.16.0
+          DOCKER_USER: ${{ secrets.DOCKER_USER }}
+          DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
+        run: |
+          wget https://github.com/operator-framework/operator-sdk/releases/download/${OPERATOR_SDK_VERSION}/operator-sdk-${OPERATOR_SDK_VERSION}-x86_64-linux-gnu -O ./operator-sdk
+          chmod +x ./operator-sdk
+          export PATH=.:$PATH
+          docker login -u "${DOCKER_USER}" -p "${DOCKER_PASSWORD}"
+          make build
+          make push
       - uses: dave-mcconnell/helm-gh-pages-microservices@master
         with:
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yaml
```diff
@@ -10,6 +10,18 @@ jobs:
...
+      - name: Operator image build and push
+        env:
+          OPERATOR_SDK_VERSION: v0.16.0
+          DOCKER_USER: ${{ secrets.DOCKER_USER }}
+          DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
+        run: |
+          wget https://github.com/operator-framework/operator-sdk/releases/download/${OPERATOR_SDK_VERSION}/operator-sdk-${OPERATOR_SDK_VERSION}-x86_64-linux-gnu -O ./operator-sdk
+          chmod +x ./operator-sdk
+          export PATH=.:$PATH
+          docker login -u "${DOCKER_USER}" -p "${DOCKER_PASSWORD}"
```
somaritane, line 22: Do GH Actions handle the redaction of these credentials from build logs?
I wonder if `docker/build-push-action@v1` can be a safer and more native option for docker image publishing.

ytsarev: @somaritane yep, i thought about it. make targets are convenient and consistent with regards to build version build/tag/push.
Also https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#example-using-cmdexe
```
Warning: Be careful that your secrets do not get printed when your action runs. When using this workaround, GitHub does not redact secrets that are printed in logs.
```
which kinda of indicates that in standard case it redacts the secrets properly. 
Anyhow, will double check on my fork

ytsarev: Checking on the fork was in any case good idea. 

See https://github.com/ytsarev/ohmyglb/runs/681281779?check_suite_focus=true

secrets are properly redacted, but I had to change VERSION retrieval style to be helm based. It is to avoid dependency on preexisting running cluster in case of `operator-sdk run -local` as in https://github.com/AbsaOSS/ohmyglb/pull/125/commits/e4b81d179b2c45bd09537d51fb3bcbd6646fdc7b

somaritane: Ah, good, I think I had issues before with that dependency running locally. So I'm happy with both confirmation of secrets reduction and that version retrieval fix :)

PR #135, opened 2020-05-29
Switch to safe geotag propagation with depresolver

ytsarev:
* Consistent cluster geo tag propagation with new deprespolver
* Should get rid us of `status.geoTag` required issues during upgrade

### DIFF
--- a/pkg/controller/gslb/gslb_controller_test.go
+++ b/pkg/controller/gslb/gslb_controller_test.go
@@ -626,6 +626,16 @@ func TestGslbController(t *testing.T) {
 		}
 	})
 
+	t.Run("Reflect GeoTag in the Status as unset by default", func(t *testing.T) {
+		reconcileAndUpdateGslb(t, r, req, cl, gslb)
+		got := gslb.Status.GeoTag
+		want := "unset"
+
+		if got != want {
+			t.Errorf("got: '%s' GeoTag status, want:'%s'", got, want)
+		}
+	})
+
 	t.Run("Reflect GeoTag in the Status", func(t *testing.T) {
 		defer func() {
 			err = os.Unsetenv("CLUSTER_GEO_TAG")
--- a/pkg/controller/gslb/gslb_controller_test.go
+++ b/pkg/controller/gslb/gslb_controller_test.go
@@ -648,6 +648,11 @@ func TestGslbController(t *testing.T) {
 		if err != nil {
 			t.Fatalf("Can't setup env var: (%v)", err)
 		}
+		resolver := depresolver.NewDependencyResolver(context.TODO(), cl)
+		r.config, err = resolver.ResolveOperatorConfig()
+		if err != nil {
+			t.Fatalf("config error: (%v)", err)
+		}
 		reconcileAndUpdateGslb(t, r, req, cl, gslb)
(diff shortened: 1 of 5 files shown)


pkg/controller/gslb/internal/depresolver/depresolver_config.go
```diff
@@ -17,6 +18,7 @@ func (dr *DependencyResolver) ResolveOperatorConfig() (*Config, error) {
 		dr.config = &Config{}
 		// set predefined values when not read from the environment variables
 		dr.config.ReconcileRequeueSeconds, _ = env.GetEnvAsIntOrFallback(reconcileRequeueSecondsKey, 30)
+		dr.config.ClusterGeoTag = env.GetEnvAsStringOrFallback(clusterGeoTagKey, "unset")
```
kuritka, line 21: just a comment: geotag don't need to be validated against values i.e. 

- crazy symbols ".?:\\!",  
- whitespace(s) "    "
- any size limit criteria - "dksjdkdskjsdjsdksdkjk sdlsfhldfh kdjfh kjhdkjfhdk hgdjk ghkjgh . fhjhfhlkhjhkjhkhgjklhjg....

pkg/controller/gslb/gslb_controller_test.go
```diff
@@ -626,6 +626,16 @@ func TestGslbController(t *testing.T) {
 		}
 	})
 
+	t.Run("Reflect GeoTag in the Status as unset by default", func(t *testing.T) {
```
kuritka, line 629: Not sure if this should be here. Test itself is needed for sure  but on controller level we should check reconciliation context. This should be in depresolver tests as it is just testing default GeoTag value when no input given.

ytsarev: @kuritka it is testing exactly subtle bug. Not just the config construction but also the Gslb.Status propagation with default value. The place where we hit the API errors during latest deployments.

pkg/controller/gslb/internal/depresolver/depresolver_config.go
ytsarev: To be honest I do not really see a reason to prescribe limitation on geo tag. Its contents craziness level is up to the user imho. @somaritane thoughts ?

pkg/controller/gslb/gslb_controller_test.go
kuritka: ok, so than I have nothing against...

pkg/controller/gslb/internal/depresolver/depresolver_config.go
kuritka: I mean, world wide users are really crazy and when invalid value is entered it fails during runtime on place where nobody expects with message which is not bounded to an invalid value (i.e. If I put -1 into TTL it could fail during runtime with random network error instead of clear "invalid TTL configuration" on the start ).

Should be this allowed `Central Europe http://cz.eu/Prague?env=local` ?

ytsarev: it's actually valid point as we construct ns dns names using the geo tag. Will look into it

PR #150, opened 2020-07-19
Fix code markup in the readme

ytsarev:


ytsarev:
Merging it in, will monitor unexpected failures of terratest pipe in master ( as it is definitely not related to readme :) )

PR #159, opened 2020-08-02
Detect mismatch of Ingress hostname and EdgeDNSZone

### LINKED ISSUE #149: Report on dnsZone and Gslb Ingress host mismatch
When we have mismatch between https://github.com/AbsaOSS/k8gb/blob/master/chart/k8gb/values.yaml#L11 and Gslb Ingress host like https://github.com/AbsaOSS/k8gb/blob/master/deploy/crds/k8gb.absa.oss_v1beta1_gslb_cr_failover.yaml#L9 we are going to send non-valid configuration to Infoblox/EdgeDNS

We probably want to:
* Detect the mismatch by k8gb operator
* Circuit break Infoblox API calls
* Clearly reflect the fact of misconfiguration in the main k8gb log stream

ytsarev:
* Log the error and requeue reconciliation loop in case of mismatch
* Resolves #149

### DIFF
--- a/pkg/controller/gslb/dnsupdate.go
+++ b/pkg/controller/gslb/dnsupdate.go
@@ -143,7 +143,7 @@ func (r *ReconcileGslb) gslbDNSEndpoint(gslb *k8gbv1beta1.Gslb) (*externaldns.DN
 		var finalTargets []string
 
 		if !strings.Contains(host, edgeDNSZone) {
-			return nil, fmt.Errorf("ingress host %s is not matching delegated zone %s", host, edgeDNSZone)
+			return nil, fmt.Errorf("ingress host %s does not match delegated zone %s", host, edgeDNSZone)
 		}
 
 		if health == "Healthy" {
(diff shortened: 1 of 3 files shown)


pkg/controller/gslb/dnsupdate.go
```diff
@@ -137,9 +137,15 @@ func (r *ReconcileGslb) gslbDNSEndpoint(gslb *k8gbv1beta1.Gslb) (*externaldns.DN
 		return nil, err
 	}
 
+	edgeDNSZone := os.Getenv("EDGE_DNS_ZONE")
+
 	for host, health := range serviceHealth {
 		var finalTargets []string
 
+		if !strings.Contains(host, edgeDNSZone) {
+			return nil, fmt.Errorf("ingress host %s is not matching delegated zone %s", host, edgeDNSZone)
```
donovanmuller, line 146: Grammar

```suggestion
			return nil, fmt.Errorf("ingress host %s does not match delegated zone %s", host, edgeDNSZone)
```

ytsarev: @donovanmuller thanks a lot for fixing my Runglish, comrade ! :)

ytsarev:
@donovanmuller added chart version bump, please reapprove

PR #169, opened 2020-10-07
Initial configuration layout for Route53 support

ytsarev:
It's not yet working solution but important foundation for further full support.

* Separate `external-dns-route53` deployment which exclusively handles route53 updates
* CoreDNS service type LoadBalancer which exposes 53/udp with Network Load Balancer without the need of special upd config of Nginx Ingress controller also avoiding 53/udp exposure on each of the worker nodes as in on-prem scenario
* K8gb configuration options for route53 enablement and associated depresolver modifications
* Helper makefile targets to ease the testing of temporary k8gb images in AWS
* Creation of DNSEndpoint CRD with NS record as a start of automated Zone delegation configuration in Route53
* Associated tests

Unfortunately external-dns currently does not pick up NS records which are described as CRD.

We will have to fix https://github.com/kubernetes-sigs/external-dns/issues/1647 upstream to proceed further with the implementation

### DIFF
--- a/controllers/dnsupdate.go
+++ b/controllers/dnsupdate.go
@@ -378,6 +378,9 @@ func filterOutDelegateTo(delegateTo []ibclient.NameServer, fqdn string) []ibclie
 func (r *GslbReconciler) configureZoneDelegation(gslb *k8gbv1beta1.Gslb) (*reconcile.Result, error) {
 	clusterGeoTag := os.Getenv("CLUSTER_GEO_TAG")
 	infobloxGridHost := os.Getenv("INFOBLOX_GRID_HOST")
+	if r.Config.Route53Enabled {
+		log.Info("Gonna rule route53")
+	}
 	if len(infobloxGridHost) > 0 {
 
 		objMgr, err := infobloxConnection()
--- a/controllers/dnsupdate.go
+++ b/controllers/dnsupdate.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"sort"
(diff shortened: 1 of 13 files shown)


controllers/dnsupdate.go
```diff
@@ -377,7 +389,38 @@ func filterOutDelegateTo(delegateTo []ibclient.NameServer, fqdn string) []ibclie
 
 func (r *GslbReconciler) configureZoneDelegation(gslb *k8gbv1beta1.Gslb) (*reconcile.Result, error) {
 	clusterGeoTag := os.Getenv("CLUSTER_GEO_TAG")
+	extClusterGeoTags := os.Getenv("EXT_GSLB_CLUSTERS_GEO_TAGS")
 	infobloxGridHost := os.Getenv("INFOBLOX_GRID_HOST")
+	if r.Config.Route53Enabled {
```
kuritka, line 394: After we implement #170 (all configuration goes to one place) we can refactor this into the factory and implementations:
i.e.
```go
//DNS factory contains only logic which based on configuration returns Route53 or infoblocks or ... or error 
dnsImplementation, err := NewDnsFactory(r.config).Get()  

//somewhere in the code we just execute
//dnsImplementation contains concrete DNS logic returned by factory
err := dnsImplementation.Execute()
```
At the end we will have simpler code because we will split logic into smaller parts which we can test independently

kuritka (approved):
I put two comments for things should be  done, otherwise LGTM

ytsarev:
@kuritka thanks a lot for thorough review 👍

PR #171, opened 2020-10-07
Fix external-dns SA definition

ytsarev:
* Externaldns Service Account definition was duplicated between coredns
  and route53 externaldns manifests
* Consolidate it in the main one with route53 keeping IRSA annotation

### DIFF
--- a/chart/k8gb/templates/external-dns/external-dns-route53.yaml
+++ b/chart/k8gb/templates/external-dns/external-dns-route53.yaml
@@ -1,14 +1,4 @@
 {{ if .Values.route53.enabled }}
-apiVersion: v1
-kind: ServiceAccount
-metadata:
-  name: external-dns
-  # If you're using Amazon EKS with IAM Roles for Service Accounts, specify the following annotation.
-  # Otherwise, you may safely omit it.
-  annotations:
-    # Substitute your account ID and IAM service role name below.
-    eks.amazonaws.com/role-arn: {{ .Values.route53.irsaRole }}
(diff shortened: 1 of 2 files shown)


chart/k8gb/templates/external-dns/external-dns-route53.yaml
```diff
@@ -36,6 +26,7 @@ spec:
         - --aws-zone-type=public # only look at public hosted zones (valid values are public, private or no value for both)
         - --registry=txt
         - --txt-owner-id= {{ .Values.route53.hostedZoneID }}
+        - --log-level=debug # debug only
```
k0da, line 29: debug by default?

k0da: Looks like leftover from debugging

kuritka (approved):

chart/k8gb/templates/external-dns/external-dns-route53.yaml
ytsarev: @k0da yes, by default. It's not super verbose still useful. We are using the same for coredns external-dns instance.

k0da: @ytsarev worth to mention it in a commit message then

ytsarev: @k0da sorry, missed this one, will document properly during next iteration

PR #172, opened 2020-10-15
Route53 support

ytsarev:
This PR adds support for AWS Route53 as EdgeDNS.

Notable differences comparing on-prem setup with Infoblox:

* GSLB Zone Delegation is fully controlled by external-dns through DNSEndpoint CRDs which are dynamically created by k8gb controller. 
* Upstream PR with NS Record support https://github.com/kubernetes-sigs/external-dns/pull/1813 is pending so we are running our own external-dns build for a time being
* Single DNSEndpoint for NS record in k8gb namespace and also single NS record per k8gb multicluster setup(also simplifies Infoblox based setup)
* CoreDNS 53/udp is exposed via NLB instead of nginx-controller based exposure on each worker node
* Workloads/workers in the reference setup are also behind NLB so k8gb now capable to resolve LoadBalancer Hostname in Ingress status and populate Gslb enabled DNS record accordingly

Note:
TXT based split brain control is skipped for now - the mechanism might be rethought given we are moving from direct API calls to edgeDNS to full utilization of external-dns capabilities ( Custom TXT creation is also lacking there as of now )


k0da (approved):
LGTM

kuritka (approved):

chart/k8gb/values.yaml
```diff
@@ -17,12 +17,11 @@ k8gb:
...
-     - "test-gslb-ns-us.example.com"
-     - "test-gslb-failover-ns-us.example.com"
+     - "gslb-ns-us.example.com"
   reconcileRequeueSeconds: 30
 
 externaldns:
-  image: k8s.gcr.io/external-dns/external-dns:v0.7.4
+  image: absaoss/external-dns:v0.7.1-401-gede9767c # temporary image containing NS record support before the merge of https://github.com/kubernetes-sigs/external-dns/pull/1813
   interval: "20s"
   expose53onWorkers: true # open 53/udp on workers nodes with nginx controller
```
somaritane, line 26: probably can use known port name instead of number, like `exposeUdpOnWorkers`

ytsarev:
```
 cat /etc/services |grep 53/udp|head
domain           53/udp     # Domain Name Server
```
53 is more clear and explicit imho. Like in Route53 naming itself lol

somaritane (approved):

Issue #170, opened 2020-10-07
Move input environment variables into depresolver


kuritka:
Historically, we load multiple environment variables in different places:  CLUSTER_GEO_TAG,  EXT_GSLB_CLUSTERS_GEO_TAGS, INFOBLOX_GRID_HOST, EDGE_DNS_ZONE... 
There is a potential risk of unset variable or invalid variable value caused by misconfiguration. Move loading environment variables into depresolver and return proper error message in case of invalid or unset value. 
Implement validations and cover by tests.

PR #173, opened 2020-10-15
Extending DepResolver

kuritka:
related to (#170)

The primary purpose of this improvement is to read all k8gb input at one place, validate it and provide to GSLB in typed, well-formatted way. Depresolver is a single point of k8gb inputs now and fail-fast with proper error message if inputs are incorrect.

Inputs:

spec
 - DNSTtlSeconds
 - SplitBrainThresholdSeconds

config
 - ReconcileRequeueSeconds
 - ClusterGeoTag 
 - Route53Enabled
 - **ExtClustersGeoTags**
 - **EdgeDNSType**
 - **EdgeDNSServer**
 - **EdgeDNSZone**
 - **DNSZone**
-  **Infoblox.Host**
- **Infoblox.Version**
- **Infoblox.Port**
- **Infoblox.Username**
- **Infoblox.Password**

 
#### Validations
We implement the new validator, applying various rules to each input. 

### Tests
75% coverage checking depresolver works as expected. Extensive coverage allow us to make safe refactoring. 
Because I introduced new validation rules, I had to update controller_tests. 

### EdgeDNSType
Providing new computed property EdgeDNSType telling which configuration(s) loaded to depresolver.  This allows us to write conditions like 
```
if config.EdgeDNSType == DNSTypeRoute53 || config.EdgeDNSType == DNSTypeInfo
```
which overall simplify logic and helps in case we provide EdgeDNS implementations by Factory  

## depresolver as single point of environment variables names 
So we are not using names like "DNS_ZONE"  directly. Instead, we use depresolver.DnsZone constant.  I replaced variables in controller_tests while the DNS code will not access inputs directly. Will be part of following PR. 

## TODO:
Will follow in next PR (s). Thanks to validator, we can remove validations from GSLB logic and simplify code. Also, accessing depresolver instead of  direct access environment variables will enhance code readability. We can more concentrate on DNS logic rather than checking inputs. It will require changes in the code that I postpone to the next PR, because this PR is too large.

### DIFF
--- a/controllers/depresolver/depresolver.go
+++ b/controllers/depresolver/depresolver.go
@@ -4,7 +4,6 @@
 // - provides predefined values when configuration is missing
 // - validates configuration
 // - executes once
-// TODO: Add the rest of configuration to be resolved
 package depresolver
 
 import (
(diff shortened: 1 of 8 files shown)


controllers/depresolver/depresolver.go
```diff
@@ -14,17 +13,58 @@ import (
...
+	// EdgeDNSType is READONLY and is set automatically by configuration
+	EdgeDNSType EdgeDNSType
+	// EdgeDNSServer
+	EdgeDNSServer string
+	// EdgeDNSZone main zone which would contain gslb zone to delegate; e.g. example.com
+	EdgeDNSZone string
+	// DNSZone controlled by gslb; e.g. cloud.example.com
+	DNSZone string
+	// DNSTypeRoute53 switch
+	//TODO: hide for depresolver subscriber as depresolver retrieves EdgeDNSType. Maybe we can change configuration and set EdgeDNSType directly instead of DNSTypeRoute53 boolean
```
somaritane, line 61: Let's make comments consistent (i.e `// TODO` instead of `//TODO`)

kuritka: THX Tim, I fixed TODOs. I use TODO now as points that will need to resolve in following PR in order to not touch live code during super large PR.

somaritane (changes requested):

controllers/depresolver/depresolver.go
```diff
@@ -14,17 +13,58 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-//Config holds operator configuration
+// EdgeDNSType specifies to which edge DNS is k8gb connecting
+type EdgeDNSType int
+
+const (
+	// DNSTypeNoEdgeDNS is default DNSType. Is used during integration testing when no infoblox or route53 exists.
```
ytsarev, line 20:
```suggestion
	// DNSTypeNoEdgeDNS is default DNSType. Is used during integration testing when no edgeDNS provider exists.
```
Let's relax the explicit names here as we are about to add more providers in future

controllers/gslb_controller_test.go
```diff
@@ -38,15 +40,29 @@ func TestGslbController(t *testing.T) {
...
 	if err != nil {
 		t.Fatalf("Can't setup env var: (%v)", err)
 	}
 
-	err = os.Setenv("FAKE_INFOBLOX", "true")
-	if err != nil {
-		t.Fatalf("Can't setup env var: (%v)", err)
+	predefinedConfig := depresolver.Config{
+		ReconcileRequeueSeconds: 30,
+		ClusterGeoTag:           "us-west1",
```
ytsarev, line 51: Should be `us-west-1` format

controllers/depresolver/depresolver_validator.go
```diff
@@ -0,0 +1,159 @@
...
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+const (
+	//hostNameRegex allows cloud region formats; e.g. af-south-1
+	geoTagRegex = "^[a-zA-Z\\-\\d]*$"
```
ytsarev, line 11: Are you sure we want to force it ? Also, how did you test the regex, didn't succeed to match region format for me, but I might be missing something

controllers/gslb_controller_test.go
```diff
@@ -38,15 +40,29 @@ func TestGslbController(t *testing.T) {
...
+		ExtClustersGeoTags:      []string{"us-east1"},
```
ytsarev, line 52: same as above

ytsarev (changes requested):

controllers/depresolver/depresolver.go
kuritka: @somaritane comments fixed

controllers/depresolver/depresolver_validator.go
kuritka: @ytsarev good point. Here is regex test:
[expected](https://github.com/AbsaOSS/k8gb/blob/170-include-envvars-into-depresolver/controllers/depresolver/depresolver_test.go#L228)
[corner case](https://github.com/AbsaOSS/k8gb/blob/170-include-envvars-into-depresolver/controllers/depresolver/depresolver_test.go#L217)

ytsarev: perfect

ytsarev (approved):
Great stuff, thanks!

somaritane (approved):

PR #177, opened 2020-10-22
Simplify values.yaml

ytsarev:
* Get rid of anchors
* No need to explicitly specify gslb dnsZone
  in coredns etcd plugin config - we can just serve the dot
* It breaks the anchoring dependency in values yaml and
  enables more simple values override on consumer side


donovanmuller (approved):

kuritka (approved):
👍

PR #182, opened 2020-10-26
Zone delegation garbage collection for Route53

ytsarev:
* Remove Zone delegation config during Gslb finalization process
* Switch external dns to sync mode and use different geo based
  ownership tags per k8gb instance
* Associated tests
* Logging bugfix inline
* Include linting as dependency to standard test make target
* Minor version bump

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ endif
 all: manager
 
 # Run tests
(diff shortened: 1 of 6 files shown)


kuritka (approved):
perfect 👍

PR #183, opened 2020-10-27
Include GSLB dns zone into NS server names

ytsarev:
* Makes NS server FQDNs unique per k8gb cluster pair
* Without this change distinct k8gb cluster deployments
  which are responsible for different gslb DNS zones
  would clash with writing to the same NS records
  if operating in the same edgeDNSZone
* Switch from env vars to depresolver config on the way


kuritka (dismissed):
I'll handle merge

kuritka (approved):

PR #187, opened 2020-10-28
Commit 'offline' Changelog

ytsarev:
There is a notion of 'best practice'
in OSS projects to keep Changelog right in git
next to Github releases.

Which kind of makes sense, so let's do that.


donovanmuller (approved):
Like this 👍

PR #191, opened 2020-10-29
Fix contributing section rendering

ytsarev:
* For some reason jekyll does not render anything that is
  named like CONTRIBUTING.md and shows it as a plain text
* Workaround it by renamed symlink
* Inline Changelog link fix

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,7 +85,7 @@ The first line is the subject and should be no longer than 70 characters, the se
 
 ## Documentation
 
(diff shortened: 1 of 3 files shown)


k0da:
looks like there is a [way](https://stackoverflow.com/questions/45899614/contributing-md-does-not-render-as-a-web-page-on-github-pages/45900131) to enable contibuting.md into render list.

k0da (changes requested):
see above comment

ytsarev:
@k0da thanks a ton for finding, fixed in `gh-pages` - https://www.k8gb.io/CONTRIBUTING.html . Closing this one

PR #194, opened 2020-10-29
Use absolute URLs in case of file reference

ytsarev:
Unfortunately relative paths are not working in Github Pages
leading to 404 in cases when we reference file/dir in the github
repository.

So unless the referenced file is md or a graphic file like png/svg
we have to stick with the absolute url to the master branch


k0da (approved):
LGTM

PR #204, opened 2020-11-18
Contexts complient with kube-builder

kuritka:
In order to migration operator-sdk 1.0 we should use contexts to be kube-builder compliant. To achieve this, I [reuse default Background context](https://github.com/AbsaOSS/k8gb/blob/master/controllers/gslb_controller.go#L60) from reconciliation loop.  see: https://book.kubebuilder.io/cronjob-tutorial/controller-implementation.html#implementing-a-controller; 
```go
func (r *CronJobReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
    ctx := context.Background()
    log := r.Log.WithValues("cronjob", req.NamespacedName)

```

I changed
 - depresolver doesn't hold context internally
 - update tests
 - use default context instead of context.TODO() within reconciliation queue

### DIFF
--- a/controllers/depresolver/depresolver_test.go
+++ b/controllers/depresolver/depresolver_test.go
@@ -44,9 +44,9 @@ var predefinedConfig = Config{
 func TestResolveSpecWithFilledFields(t *testing.T) {
 	// arrange
 	cl, gslb := getTestContext("./testdata/filled_omitempty.yaml")
-	resolver := NewDependencyResolver(context.TODO(), cl)
+	resolver := NewDependencyResolver(cl)
 	// act
-	err := resolver.ResolveGslbSpec(gslb)
+	err := resolver.ResolveGslbSpec(context.TODO(), gslb)
 	// assert
 	assert.NoError(t, err)
 	assert.Equal(t, 35, gslb.Spec.Strategy.DNSTtlSeconds)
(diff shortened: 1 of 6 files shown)


controllers/depresolver/depresolver_test.go
```diff
@@ -92,21 +92,22 @@ func TestResolveSpecWithEmptyFields(t *testing.T) {
 func TestResolveSpecWithNegativeFields(t *testing.T) {
 	// arrange
 	cl, gslb := getTestContext("./testdata/invalid_omitempty_negative.yaml")
-	resolver := NewDependencyResolver(context.TODO(), cl)
+	resolver := NewDependencyResolver(cl)
 	// act
-	err := resolver.ResolveGslbSpec(gslb)
+	err := resolver.ResolveGslbSpec(context.TODO(), gslb)
```
ytsarev, line 97: Why do we still use TODO() there

ytsarev (commented):
Btw I was always wondering of a real difference between `TODO()` vs `Background()`

I guess I found the answer https://essential-go.programming-books.io/context-todo-vs-context-background-d5224e27ff724a33a79cb4e03a5eb333#:~:text=TODO%20returns%20a%20non%2Dnil%2C%20empty%20Context%20.&text=Background%20returns%20a%20non%2Dnil,level%20Context%20for%20incoming%20requests.

The difference is NONE, lol

config/crd/bases/k8gb.absa.oss_gslbs.yaml
```diff
@@ -289,16 +289,16 @@ spec:
...
           type: object
       type: object
   version: v1beta1
   versions:
-  - name: v1beta1
-    served: true
-    storage: true
+    - name: v1beta1
+      served: true
+      storage: true
```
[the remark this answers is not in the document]
kuritka: 👍 , I pushed generated file by mistake and I removed as it is not part of PR.

controllers/depresolver/depresolver_test.go
kuritka: @ytsarev good question: The point is that you are creating i.e. WithCancel context from existing context.

 If you use Background context, you explicitly say: Hey, I'm using context from the top of the tree, usually `main` and this context will instantiate another context(s) afterwards `WithCancel`, `WithDeadline`... 

If you use context.TODO() you explicitly say: function just requires some context argument so I will pass TODO which is doing nothing, just to compile. For instance, our tests..

ytsarev: @kuritka cool, makes total sense

ytsarev (approved):

PR #205, opened 2020-11-20
Enable Gslb with Ingress Annotation

### LINKED ISSUE #200: Add ability to reuse existing Ingress
Currently, k8gb creates its own controlled Ingress from Gslb CR spec.
It is the desired scenario for cases when Gslb enabled Ingress should co-exist next to a standard one.

We have received community feedback from Rancher and Admiralty guys to add an ability to enable Gslb for already existing Ingress.

From end user perspective it might look like setting `k8gb.io/strategy: failover` annotation on existing Ingress.
It might also help adoption as instead of creating Gslb kind CR end users would need to just set annotation(s) on standard Ingress object.

Here we also have couple of k8gb desired behaviors we will have to choose to impement:

1. k8gb observes annotation on Ingress object and creates associated Gslb CR and populating its status ( kind of reversed activity of what is happening now when Gslb creates Ingress )
2. Ditch Gslb CRD completely and populate health status through annotation. Managed FQDNs and IP addresses will be reflected in DNSEndpoint objects.

ytsarev:
* Resolves #200
* Reacts to Ingresses with specific annotations
* Dynamically creates Gslb resources according to annotation contents
* Associated unit and terratests
* Brief documentation

### DIFF
--- a/controllers/gslb_controller.go
+++ b/controllers/gslb_controller.go
@@ -210,7 +210,7 @@ func (r *GslbReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			Name:      a.Meta.GetName(),
 		}, ingressToReuse)
 		if err != nil {
-			log.Info(fmt.Sprintf("Ingress(%s) does not exist anymore. Skipping...", a.Meta.GetName()))
+			log.Info(fmt.Sprintf("Ingress(%s) does not exist anymore. Skipping Glsb creation...", a.Meta.GetName()))
 			return
 		}
 		gslbExist := &k8gbv1beta1.Gslb{}
--- a/controllers/gslb_controller.go
+++ b/controllers/gslb_controller.go
@@ -240,6 +240,10 @@ func (r *GslbReconciler) SetupWithManager(mgr ctrl.Manager) error {
 					gslb.Spec.Strategy.PrimaryGeoTag = annotationValue
 				}
 			}
+			if gslb.Spec.Strategy.PrimaryGeoTag == "" {
+				log.Info("k8gb.io/primarygeotag annotation is missing, skipping Gslb creation...")
+				return
+			}
 		}
 
 		log.Info(fmt.Sprintf("Creating new Gslb(%s) out of Ingress annotation", gslb.Name))
--- a/controllers/gslb_controller.go
+++ b/controllers/gslb_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/AbsaOSS/k8gb/controllers/depresolver"
@@ -25,6 +26,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	v1beta1 "k8s.io/api/extensions/v1beta1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	types "k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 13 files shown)


controllers/gslb_controller.go
```diff
@@ -195,13 +198,78 @@ func (r *GslbReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+				Ingress: ingressToReuse.Spec,
+				Strategy: k8gbv1beta1.Strategy{
+					Type: strategy,
+				},
+			},
+		}
+
+		if strategy == "failover" {
+			for annotationKey, annotationValue := range a.Meta.GetAnnotations() {
+				if annotationKey == "k8gb.io/primarygeotag" {
```
donovanmuller, line 239: What happens in the case where `k8gb.io/primarygeotag` annotation is missing? Should we log an error and ignore?

ytsarev: @donovanmuller good catch, thanks a lot. Implemented in https://github.com/AbsaOSS/k8gb/pull/205/commits/7fb78f6a5358f680a262584c5d8bb5b89f5b5899

donovanmuller (dismissed):

controllers/gslb_controller.go
```diff
@@ -195,13 +198,82 @@ func (r *GslbReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		_ = c.Create(context.Background(), gslb)
+	}
+	ingressMapFn := handler.ToRequestsFunc(
+		func(a handler.MapObject) []reconcile.Request {
+			for annotationKey, annotationValue := range a.Meta.GetAnnotations() {
+				if annotationKey == "k8gb.io/strategy" {
+					switch annotationValue {
+					case "roundRobin":
+						createGslbFromIngress(annotationKey, annotationKey, a, "roundRobin")
+					case "failover":
```
kuritka, line 259: consider `"failover","roundRobin"` as constants

kuritka (approved):

controllers/gslb_controller.go
ytsarev: @kuritka I thought about it but given I can't use constants as a case value(`case roundRobinConstant` will not work afaik) the constants introduction here looks a bit artificial

ytsarev: @kuritka yeah, actually they work. Implemented constants in https://github.com/AbsaOSS/k8gb/pull/205/commits/86260ff365145f5e98c3a457280e0a9fadec0477

kuritka (changes requested):
please reuse also in `dnsupdate.go`

controllers/gslb_controller.go
```diff
@@ -195,13 +198,82 @@ func (r *GslbReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		log.Info(fmt.Sprintf("Detected strategy annotation(%s:%s) on Ingress(%s)",
+			annotationKey, annotationValue, a.Meta.GetName()))
+		c := mgr.GetClient()
+		ingressToReuse := &v1beta1.Ingress{}
+		err := c.Get(context.Background(), client.ObjectKey{
+			Namespace: a.Meta.GetNamespace(),
+			Name:      a.Meta.GetName(),
+		}, ingressToReuse)
+		if err != nil {
+			log.Info(fmt.Sprintf("Ingress(%s) does not exist anymore. Skipping...", a.Meta.GetName()))
```
somaritane, line 213: Consider update for better context:
```suggestion
			log.Info(fmt.Sprintf("Ingress(%s) does not exist anymore. Skipping Glsb creation...", a.Meta.GetName()))
```

controllers/gslb_controller.go
```diff
@@ -195,13 +198,82 @@ func (r *GslbReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			return
+		}
+		gslbExist := &k8gbv1beta1.Gslb{}
+		err = c.Get(context.Background(), client.ObjectKey{
+			Namespace: a.Meta.GetNamespace(),
+			Name:      a.Meta.GetName(),
+		}, gslbExist)
+		if err == nil {
+			log.Info(fmt.Sprintf("Gslb(%s) already exists. Skipping...", gslbExist.Name))
```
somaritane, line 222: Consider update for better context:
```suggestion
			log.Info(fmt.Sprintf("Gslb(%s) already exists. Skipping Gslb creation...", gslbExist.Name))
```

somaritane (approved):
lgtm

kuritka (approved):

PR #207, opened 2020-11-23
Fix helm args

kuritka:
quick fix on demand


k0da:
@kuritka pls rebase

kuritka:
invalid

PR #212, opened 2020-12-03
Primary geotag annotation fix

### LINKED ISSUE #210: Replace k8gb.io/primarygeotag annotation with k8gb.io/primary-geotag
I know this is minor issue, and there are no strict conventions on [k8s annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/).
But `primary` and `geotag` in `k8gb.io/primarygeotag` are two separate words, so it's better to separate them with `dash` in order to make annotation more readable, avoid authoring errors, and also align with the style used by existing Ingress Controllers known to community (e.g. https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/, https://doc.traefik.io/traefik/v1.7/configuration/backends/kubernetes/#general-annotations): 
```diff
--- k8gb.io/primarygeotag
+++ k8gb.io/primary-geotag
```

somaritane:
This PR renames `k8gb.io/primarygeotag` annotation to `k8gb.io/primary-geotag` 
Closes #210

### DIFF
--- a/controllers/gslb_controller.go
+++ b/controllers/gslb_controller.go
@@ -51,10 +51,14 @@ type GslbReconciler struct {
 	DepResolver *depresolver.DependencyResolver
 }
 
-const k8gbNamespace = "k8gb"
-const gslbFinalizer = "finalizer.k8gb.absa.oss"
-const roundRobinStrategy = "roundRobin"
-const failoverStrategy = "failover"
+const (
+	k8gbNamespace           = "k8gb"
+	gslbFinalizer           = "finalizer.k8gb.absa.oss"
+	roundRobinStrategy      = "roundRobin"
+	failoverStrategy        = "failover"
+	primaryGeoTagAnnotation = "k8gb.io/primary-geotag"
+	strategyAnnotation      = "k8gb.io/strategy"
+)
 
 // +kubebuilder:rbac:groups=k8gb.absa.oss,resources=gslbs,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=k8gb.absa.oss,resources=gslbs/status,verbs=get;update;patch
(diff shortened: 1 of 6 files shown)


controllers/gslb_controller.go
```diff
@@ -238,12 +242,12 @@ func (r *GslbReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 		if strategy == failoverStrategy {
 			for annotationKey, annotationValue := range a.Meta.GetAnnotations() {
-				if annotationKey == "k8gb.io/primarygeotag" {
+				if annotationKey == primaryGeoTagAnnotation {
 					gslb.Spec.Strategy.PrimaryGeoTag = annotationValue
 				}
 			}
 			if gslb.Spec.Strategy.PrimaryGeoTag == "" {
-				log.Info("k8gb.io/primarygeotag annotation is missing, skipping Gslb creation...")
+				log.Info(fmt.Sprintf("%s annotation is missing, skipping Gslb creation...", primaryGeoTagAnnotation))
```
k0da, line 250: Isn't it better to use Infof() here?

k0da (changes requested):
see above

kuritka (approved):
lgtm

controllers/gslb_controller.go
somaritane: Makes sense, @k0da, but we also should maintain consistency, and rest of the code uses `Info(fmt.Sprintf())`, not `Infof()`.
If you're OK, I can make the proposed replacement (35 changes in 5 files) as separate PR.

kuritka: @somaritane, good point for sure! If you decide to reactor, please can you check log.Error, log.Debug ,etc... We can do all at once.

kuritka: Maybe we can do in separate PR.

k0da (approved):

controllers/gslb_controller.go
k0da: @somaritane ack

somaritane: Also, after carefully looking, this `log` is of `logr.Logger` type, [logr](https://pkg.go.dev/github.com/go-logr/logr) package  doesn't have functions like `Infof()` or `Errorf()`, instead you should use `Info()` and `Error()` with structured logging format e.g.:
```diff
--- log.Errorf("unable to reconcile object %s/%s: %v", object.Namespace, object.Name, err)
+++ log.Error(err, "unable to reconcile object", "object", object)
```

PR #215, opened 2020-12-11
Relax requirement on k8gb namespace name

### LINKED ISSUE #129: Relax dependency on specific 'k8gb' namespace name
* We are relying on operator namespace to be named 'k8gb' in the helm chart templates and in the codebase
* It is not really flexible and goes against https://github.com/operator-framework/community-operators/blob/master/docs/best-practices.md
* To fix that we might want to use the approach similar to https://github.com/helm/charts/pull/22378/files

ytsarev:
* Do not rely on hardcoded `k8gb` namespace anymore
* Helm chart template fix
* Dynamically get current k8gb controller namespace
  with the help of Downward API
  https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#use-pod-fields-as-values-for-environment-variables
* Resolves #129

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -104,6 +104,8 @@ tags
 ### JetBrains IDE ###
 .idea
 # End of idea tools
- 
 # Exclude chart tar.gz files
 chart/k8gb/charts/*.tgz
+
+# Ignore testing kubeconfigs
+kubeconfig*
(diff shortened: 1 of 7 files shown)


k0da (approved):
LGTM

PR #216, opened 2020-12-12
Fix cluster communication in full local setup

ytsarev:
* Fix `hostAlias` for test-gslb2 cluster respecting
  new style of common single ns server name per geo tag


somaritane (approved):
lgtm

PR #217, opened 2020-12-13
NS1 support

ytsarev:
* Chart extension for external-dns based ns1 provider
* Move common code between route53 and NS1(and potentially other providers) to separate function
* Depresolver extension with NS1 provider type
* Depresolver and controller tests
* Chart version bump

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -201,6 +201,12 @@ infoblox-secret:
 		--from-literal=EXTERNAL_DNS_INFOBLOX_WAPI_USERNAME=$${WAPI_USERNAME} \
 		--from-literal=EXTERNAL_DNS_INFOBLOX_WAPI_PASSWORD=$${WAPI_PASSWORD}
 
+# creates ns1 secret in current cluster
+.PHONY: ns1-secret
+ns1-secret:
+	kubectl -n k8gb create secret generic ns1 \
+		--from-literal=apikey=$${NS1_APIKEY}
+
 # install CRDs into a cluster
 .PHONY: install
 install:
--- a/Makefile
+++ b/Makefile
@@ -205,7 +205,7 @@ infoblox-secret:
 .PHONY: ns1-secret
 ns1-secret:
 	kubectl -n k8gb create secret generic ns1 \
-		--from-literal=apikey=$${NS1_APIKEY}
+		--from-literal=apiKey=$${NS1_APIKEY}
 
 # install CRDs into a cluster
 .PHONY: install
--- a/chart/k8gb/templates/external-dns/external-dns-ns1.yaml
+++ b/chart/k8gb/templates/external-dns/external-dns-ns1.yaml
@@ -31,7 +31,7 @@ spec:
           valueFrom:
             secretKeyRef:
               name: ns1
-              key: apikey
+              key: apiKey
       securityContext:
         fsGroup: 65534 # For ExternalDNS to be able to read Kubernetes and AWS token files
 {{ end }}
--- a/chart/k8gb/templates/external-dns/external-dns-ns1.yaml
+++ b/chart/k8gb/templates/external-dns/external-dns-ns1.yaml
@@ -1,0 +1,37 @@
+{{ if .Values.ns1.enabled }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: external-dns-ns1
+spec:
+  strategy:
+    type: Recreate
+  selector:
+    matchLabels:
+      app: external-dns-ns1
+  template:
+    metadata:
+      labels:
+        app: external-dns-ns1
+    spec:
+      serviceAccountName: external-dns
+      containers:
+      - name: external-dns
+        image: {{ .Values.externaldns.image }}
+        args:
+        - --source=crd
+        - --domain-filter={{ .Values.k8gb.edgeDNSZone }} # will make ExternalDNS see only the hosted zones matching provided domain, omit to process all available hosted zones
+        - --annotation-filter=k8gb.absa.oss/dnstype=ns1 # filter out only relevant DNSEntrypoints
+        - --provider=ns1
+        - --txt-owner-id=k8gb-{{ .Values.k8gb.clusterGeoTag }}
+        - --policy=sync # enable full synchronization including record removal
+        - --log-level=debug # debug only
+        env:
+        - name: NS1_APIKEY
+          valueFrom:
+            secretKeyRef:
+              name: ns1
+              key: apikey
+      securityContext:
+        fsGroup: 65534 # For ExternalDNS to be able to read Kubernetes and AWS token files
+{{ end }}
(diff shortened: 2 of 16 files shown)


k0da:
LGTM

kuritka (dismissed):

chart/k8gb/values.yaml
```diff
@@ -92,3 +92,6 @@ route53:
   enabled: false
   hostedZoneID: ZXXXSSS
   irsaRole: arn:aws:iam::111111:role/external-dns
+
+ns1:
+  enabled: true
```
somaritane, line 97: Should this value be set to `false` by default?
Also, my understanding is that we support one EdgeDNS provider at runtime.
It therefore makes sense to introduce general `edgeDNSType` parameter in confifuration and update logic in  depresolver's `getEdgeDNSType()` accordingly, wdyt?

chart/k8gb/templates/external-dns/external-dns-ns1.yaml
```diff
@@ -0,0 +1,37 @@
...
+        - --provider=ns1
+        - --txt-owner-id=k8gb-{{ .Values.k8gb.clusterGeoTag }}
+        - --policy=sync # enable full synchronization including record removal
+        - --log-level=debug # debug only
+        env:
+        - name: NS1_APIKEY
+          valueFrom:
+            secretKeyRef:
+              name: ns1
+              key: apikey
```
somaritane, line 34: Consider renaming in analogy with `apiVersion`
```suggestion
              key: apiKey
```

Makefile
```diff
@@ -201,6 +201,12 @@ infoblox-secret:
 		--from-literal=EXTERNAL_DNS_INFOBLOX_WAPI_USERNAME=$${WAPI_USERNAME} \
 		--from-literal=EXTERNAL_DNS_INFOBLOX_WAPI_PASSWORD=$${WAPI_PASSWORD}
 
+# creates ns1 secret in current cluster
+.PHONY: ns1-secret
+ns1-secret:
+	kubectl -n k8gb create secret generic ns1 \
+		--from-literal=apikey=$${NS1_APIKEY}
```
somaritane, line 208: Consider renaming in analogy with `apiVersion`
```suggestion
		--from-literal=apiKey=$${NS1_APIKEY}
```

somaritane (changes requested):

chart/k8gb/values.yaml
ytsarev: it should be, that's why it is labeled WIP :)

ytsarev: not sure about `edgeDNSType`. Options heavily deviate from provider to provider and frequently not compatible

somaritane: You're right, it can be `provider` or `providerType`. 
My understanding is that options `infoblox.enabled`, `route53.enabled`,  `ns1.enabled` are mutually exclusive, and can be replaced with that one option in order to avoid confusion in the future, when more providers are added.

kuritka: Hi, agree with @somaritane. For now, depresolver acts as [facade](https://en.wikipedia.org/wiki/Facade_pattern), so transforms to `edgeDNSType` internally. To make it cleaner, we should extract parameter directly into configuration.

ytsarev: @somaritane @kuritka it means mixing up totally unrelated configuration options under the same hierarchy level. I would be very careful with that. We can attack it in a separate issue

ytsarev:
image
Just for illustration - populated NS1 web console after testing Gslb creation

chart/k8gb/values.yaml
ytsarev: The Issue for the above https://github.com/AbsaOSS/k8gb/issues/219

docs/deploy_ns1.md
```diff
@@ -0,0 +1,31 @@
...
+
+Two EKS clusters in `eu-west-1` and `us-east-1`.
+
+Terraform code for cluster reference setup can be found [here](https://github.com/AbsaOSS/k8gb/tree/master/docs/examples/route53)
+
+The EKS setup is identical to [Route53 tutorial](/docs/deploy_route53.md)
+
+## Deploy k8gb
+
+Example values.yaml override configs can be found [here](https://github.com/AbsaOSS/k8gb/tree/master/docs/examples/ns1/)
```
donovanmuller, line 15:
```suggestion
Example `values.yaml` override configs can be found [here](https://github.com/AbsaOSS/k8gb/tree/master/docs/examples/ns1/)
```

docs/deploy_ns1.md
```diff
@@ -0,0 +1,31 @@
...
+
+You can use `helm` to deploy stable release from helm repo
```
donovanmuller, line 17:
```suggestion
You can use `helm` to deploy stable release from Helm repo
```

donovanmuller (commented):
Minor suggestions

chart/k8gb/templates/operator.yaml
```diff
@@ -86,3 +86,7 @@ spec:
             - name: ROUTE53_ENABLED
               value: "true"
             {{ end }}
+            {{ if .Values.ns1.enabled }}
+            - name: NS1_ENABLED
+              value: "true"
```
kuritka, line 91: does helm support something like this ? 
```yaml
 - name: ROUTE53_ENABLED
    value:  {{ .Values.route53.enabled }}
 - name: NS1_ENABLED
    value:  {{ .Values.ns1.enabled }}

```

controllers/dnsupdate.go
```diff
@@ -513,6 +520,7 @@ func (r *GslbReconciler) configureZoneDelegation(gslb *k8gbv1beta1.Gslb) (*recon
 				return &reconcile.Result{}, err
 			}
 		}
+
 	}
 	return nil, nil
```
kuritka, line 525: Thx to refactor into  `configureZoneDelegation` factory, we are just little step from factory - strategy patterns. 

Should we rather manage `case DNSTypeNoEdgeDNS:`  as well and return `nil, error(unhandled DNS type...)` by default ?

kuritka (changes requested):
two open questions from my side, otherwise looks great!

chart/k8gb/templates/operator.yaml
ytsarev: @kuritka maybe but it will be declared but empty variables, do not think we want to pollute the environment with them

controllers/dnsupdate.go
ytsarev: @kuritka done in https://github.com/AbsaOSS/k8gb/pull/217/commits/36e2c34f3a0d81a8c9ea5c05c34e5224de561cb6

kuritka (approved):
lgtm

somaritane (approved):
lgtm

PR #224, opened 2020-12-16
Switch to new chart repos for dependency charts

ytsarev:
https://kubernetes-charts.storage.googleapis.com chart mirror
is deprecated and doesn't work anymore.

So we have to switch:

* CoreDNS to the one from https://artifacthub.io/
* etcd-operator to read-only mirror of stable repo


k0da (approved):
LGTM

PR #242, opened 2020-12-27
README support table update

ytsarev:


README.md
```diff
@@ -82,13 +82,17 @@ k8gb is very well tested with the following environment options
...
-| EdgeDNS                          | Infoblox, Route53                                                       |
+| Kubernetes Version               | >= 1.15                                                                 |
+| Environment                      | On-prem(vSphere), AWS(EKS) [*](#clarify)                                |
+| Ingress Controller               | NGINX, AWS Load Balancer Controller [*](#clarify)                       |
+| EdgeDNS                          | Infoblox, Route53, NS1                                                  |
+
+<a name="clarify"></a>* We explicitly mention solutions where we practically tested k8gb installation.
+If your environment or Ingress controller does not match this table it does not mean that k8gb will not be
+a fit for you. The k8gb is architected to be run on top of any Kubernetes cluster and Ingress controller
+without tight coupling with specific k8s distro and/or Ingress controller.
```
donovanmuller, line 93:
```suggestion
<a name="clarify"></a>* We only mention solutions where we have tested and verified a k8gb installation.
If your Kubernetes version or Ingress controller is not included in the table above, it does not mean that k8gb will not work for you. k8gb is architected to run on top of any compliant Kubernetes cluster and Ingress controller.
```

README.md
```diff
@@ -82,13 +82,17 @@ k8gb is very well tested with the following environment options
 
 | Type                             | Implementation                                                          |
 |----------------------------------|-------------------------------------------------------------------------|
-| Kubernetes Version               | >= 1.14 (with [install workaround][install-workaround] >= 1.15 (Stable) |
-| Environment                      | On-prem(vSphere), AWS(EKS)                                              |
-| Ingress Controller               | NGINX                                                                   |
-| EdgeDNS                          | Infoblox, Route53                                                       |
+| Kubernetes Version               | >= 1.15                                                                 |
+| Environment                      | On-prem(vSphere), AWS(EKS) [*](#clarify)                                |
```
donovanmuller, line 86:
```suggestion
| Environment                      | Self-managed, AWS(EKS) [*](#clarify)                                |
```

@ytsarev wondering if we should just mention "Self-managed" Kubernetes or something instead of `On-prem(vSphere),`? The fact that it is Kubernetes on top of vSphere is not really relevant, just that it is not a _managed_ Kubernetes service like EKS/AKS/GKE etc. but rather self-managed.

Would rather not give the impression that there is anything reliant on vSphere in any way.

ytsarev: @donovanmuller great point, thanks.

donovanmuller (approved):

PR #245, opened 2021-01-06
fix terratests

### LINKED ISSUE #211: terratests - Possible race condition
The pictures below show the result of two consecutive test runs. Very often it happens that `TestK8gbIngressAnnotationFailover, TestK8gbIngressAnnotationRR` do not pass. Please investigate and fix the issue. Because the tests sometimes pass, and sometimes don't, I think the issue is race condition. I must add it that both kind and k3d errors occur.

Screenshot 2020-12-03 at 09 19 51

Screenshot 2020-12-03 at 09 24 03

kuritka:
closes #211 

The error was caused by `t.Errorf(...)` instead of `t.Logf(...)`. Both print the message but Errorf sets property `*testing.T.Failed` to `true` which causes test fails although it passed.
`t.Errorf(...)` was executed  in case the resource was not ready. That's why GHA is stable, while on local mac failed sometimes.

### DIFF
--- a/terratest/test/helpers.go
+++ b/terratest/test/helpers.go
@@ -56,8 +56,8 @@ func DoWithRetryWaitingForValueE(t *testing.T, actionDescription string, maxRetr
 
 		output, err = action()
 		if err != nil {
-			return output, nil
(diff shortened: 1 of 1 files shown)


donovanmuller (approved):

k0da (approved):

ytsarev:
@kuritka cool find, thanks a lot!

Issue #255, opened 2021-01-14
Split dnsupdate into provider pattern


kuritka:
- Extract functionality into [providers](https://medium.com/swlh/provider-model-in-go-and-why-you-should-use-it-clean-architecture-1d84cfe1b097)
- clean `dnsupdate`
- Extract metrics from controller
- consider moving `utility` into `AbsaOSS/gopg`

Issue #124, opened 2020-05-15
Expose advanced metrics


somaritane:
This issue is a spin off from #47 

Implement relevant metrics that provide insight into:
- Edge DNS integrations
- Split brain scenarios / heartbeat
- Failover scenarios
- Reconciliation
- Load balancing strategies used

https://github.com/AbsaOSS/ohmyglb/pull/97 can be used as reference implementation

kuritka:
ℹ️  note : Consider implementing against the `IMetrics` interface.
e.g. 

```golang
type IMetrics interface {
   Register() error
   Unregister() error
   UpdateIngressHostsPerStatusMetric(gslb *k8gbv1beta1.Gslb, serviceHealth map[string]string)
   UpdateHealthyRecordsMetric(gslb *k8gbv1beta1.Gslb, healthyRecords map[string][]string)

   ConcreteMetricA() (data, err)
   ConcreteMetricB() (data, err)
   ConcreteMetricC() (data, err)
...
}
```

PR #256, opened 2021-01-14
Extract prometheus metrics #1

kuritka:
Related to #255
This is the first part of [larger](https://github.com/AbsaOSS/k8gb/compare/147-refactoring) refactoring.

 -  ~~move `Dig` to utils~~
 -  ~~move `prettyPrint()` to utils and rename to `ToJSON()`~~
 -  ~~cover `ToJSON()` by tests~~
 -  move environment variable `POD_NAMESPACE` to `depresolver`
 -  extend depresolver tests
 -  extend depresolver validators for namespace validator
 -  extract Prometheus metrics into standalone module
 -  extend Prometheus metrics with `Register()`,`Unregister()`
 -  remove `controllers/metrics.go`
 -  fix controller tests
 -  remove dead registry/k3s.yaml file (pushed by mistake)

TODO:
  - `dnsupdate` refactor, to provider pattern, will come in the next PR
  - finally, some utils will go into [gopkg](https://github.com/AbsaOSS/gopkg)
  - `IMetrics` introduction (should be implemented with #124)


ytsarev (commented):
Look good, but not sure if `ToJSON` is properly reflecting semantics... Originally the function was created just for readable struct printing...

somaritane (changes requested):
I'd probably separate Dig and prettyPrint refactoring to different PRs, as they are out of context of this PR

kuritka:
> Look good, but not sure if `ToJSON` is properly reflecting semantics... Originally the function was created just for readable struct printing...

I see, but function accepts interface so can be used to any type. Of course I can limit to structures only, and return error if passed argument is not structure. 
Do you think rename may help ? What about `ToString(a interface{}) string` ?

```go
str{"Foo", 007} => 

{
	"Name": "Foo",
	"Value": 7
}

123 => "123"
true => "true"
nil => "null"  // marshalling implicitly returns this, but can be changed. 
```

kuritka:
> I'd probably separate Dig and prettyPrint refactoring to different PRs, as they are out of context of this PR

@somaritane, amended.

somaritane (approved):
lgtm 👍

ytsarev (approved):
looks good

PR #257, opened 2021-01-15
Refactor #2, Dig 

kuritka:
related to #255 

 - Dig implementation moved from controller to `utils`
 - Tests for Dig require internet connection. Skipping tests If no internet connection available.
 - Remove `log.Info()` from dig, instead return extended error
 - call `log.Info()` in caller functions

### DIFF
--- a/controllers/dnsupdate.go
+++ b/controllers/dnsupdate.go
@@ -9,13 +9,14 @@ import (
 	"strings"
 	"time"
 
+	"github.com/AbsaOSS/k8gb/controllers/internal/utils"
+
 	"github.com/AbsaOSS/k8gb/controllers/depresolver"
 
 	coreerrors "errors"
 
 	k8gbv1beta1 "github.com/AbsaOSS/k8gb/api/v1beta1"
 	ibclient "github.com/infobloxopen/infoblox-go-client"
-	"github.com/lixiangzhong/dnsutil"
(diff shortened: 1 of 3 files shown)


controllers/dnsupdate.go
```diff
@@ -52,7 +53,7 @@ func (r *GslbReconciler) getGslbIngressIPs(gslb *k8gbv1beta1.Gslb) ([]string, er
 			gslbIngressIPs = append(gslbIngressIPs, ip.IP)
 		}
 		if len(ip.Hostname) > 0 {
-			IPs, err := Dig(r.Config.EdgeDNSServer, ip.Hostname)
+			IPs, err := utils.Dig(r.Config.EdgeDNSServer, ip.Hostname)
 			if err != nil {
 				return nil, err
```
ytsarev, line 58: as `log.Info()` was removed from `Dig` -  we need to explicitly log it here, otherwise the failure of resolution will be hidden from the end user

ytsarev (changes requested):

controllers/dnsupdate.go
kuritka: @ytsarev  thx for finding 👍 . I forgot .  Will log in caller function(s) as dnsupdate keeps dependency to log.

ytsarev (approved):

PR #259, opened 2021-01-19
Make CoreDNS exposure controllable

ytsarev:
* Do not rely on specific environment but allow control over exposure
* Resolve NSServer IP addresses according to the exposure scenario


kuritka (approved):
LGTM

PR #284, opened 2021-02-05
Consolidate `v` part of version tag in the Chart metadata

ytsarev:
Instead of repeating `v` thing here and there - put in the single
place as it introduced multiple issues already


kuritka (dismissed):

donovanmuller (approved):
`v` for vendetta

somaritane (dismissed):

k0da (approved):

kuritka (approved):

PR #287, opened 2021-02-05
README: Replaced screenshot with code excerpt

somaritane:


kuritka (approved):
good idea!

ytsarev (approved):
It will not expose my fancy Vim color scheme setup but OK :D

PR #294, opened 2021-02-12
Switch sed to helm template

k0da:
Wrap gslb test resources with dummy helm, in oder to leverage helm
template engine and get rid of gnused requirement on Mac


somaritane (approved):
lgtm 👍 , we shouldn't expect users to use certain sed package.

ytsarev:
the integration tests are dependant on the location of the templates

ytsarev (changes requested):
As the comment above

k0da:
sure, I need to rethink it a little bit more.

PR #299, opened 2021-02-15
Validate spec.ingress.rules.host

k0da:
Since controller relies on it, we need to explicitly require
spec.ingress.rules.http

Fixes https://github.com/AbsaOSS/k8gb/issues/296

### DIFF
--- a/controllers/depresolver/depresolver_spec.go
+++ b/controllers/depresolver/depresolver_spec.go
@@ -25,7 +25,7 @@ func (dr *DependencyResolver) ResolveGslbSpec(ctx context.Context, gslb *k8gbv1b
 		if strategy.SplitBrainThresholdSeconds == 0 {
 			strategy.SplitBrainThresholdSeconds = predefinedStrategy.SplitBrainThresholdSeconds
 		}
-		dr.errorSpec = dr.validateSpec(strategy)
+		dr.errorSpec = dr.validateStrategy(strategy)
 		if dr.errorSpec == nil {
 			dr.errorSpec = dr.client.Update(ctx, gslb)
 		}
@@ -38,7 +38,7 @@ func (dr *DependencyResolver) ResolveGslbSpec(ctx context.Context, gslb *k8gbv1b
 	return dr.errorSpec
 }
 
-func (dr *DependencyResolver) validateSpec(strategy *k8gbv1beta1.Strategy) (err error) {
+func (dr *DependencyResolver) validateStrategy(strategy *k8gbv1beta1.Strategy) (err error) {
 	err = field("DNSTtlSeconds", strategy.DNSTtlSeconds).isHigherOrEqualToZero().err
 	if err != nil {
 		return
--- a/controllers/depresolver/depresolver_spec.go
+++ b/controllers/depresolver/depresolver_spec.go
@@ -2,6 +2,7 @@ package depresolver
 
 import (
 	"context"
+	"errors"
 
 	k8gbv1beta1 "github.com/AbsaOSS/k8gb/api/v1beta1"
 )
@@ -27,6 +28,11 @@ func (dr *DependencyResolver) ResolveGslbSpec(ctx context.Context, gslb *k8gbv1b
 		dr.errorSpec = dr.validateSpec(strategy)
 		if dr.errorSpec == nil {
 			dr.errorSpec = dr.client.Update(ctx, gslb)
+		}
+		for _, rule := range gslb.Spec.Ingress.Rules {
+			if rule.HTTP == nil {
+				dr.errorSpec = errors.New("missing .spec.ingress.rules.http")
+			}
 		}
 	})
 	return dr.errorSpec
--- a/controllers/gslb_controller.go
+++ b/controllers/gslb_controller.go
@@ -91,7 +91,7 @@ func (r *GslbReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 
 	err = r.DepResolver.ResolveGslbSpec(ctx, gslb)
 	if err != nil {
-		log.Error(err, "resolving spec.strategy")
+		log.Error(err, "resolving spec")
 		return ctrl.Result{}, err
 	}
 	// == Finalizer business ==
@@ -190,6 +190,10 @@ func (r *GslbReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			gslbName := ""
 			for _, gslb := range gslbList.Items {
 				for _, rule := range gslb.Spec.Ingress.Rules {
+					if rule.HTTP == nil {
+						log.Info(fmt.Sprintf("spec.ingress.rules.http is missing on %s", gslb.Name))
+						continue
+					}
 					for _, path := range rule.HTTP.Paths {
 						if path.Backend.ServiceName == a.Meta.GetName() {
 							gslbName = gslb.Name
--- a/terratest/test/k8gb_basic_failover_test.go
+++ b/terratest/test/k8gb_basic_failover_test.go
@@ -77,5 +77,4 @@ func TestK8gbBasicFailoverExample(t *testing.T) {
 
 		assert.Equal(t, afterFailoverResponse, expectedIPsAfterFailover)
 	})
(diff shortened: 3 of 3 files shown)


k0da:
replace http with https in Gslb object and submit it:
```shell 
18:01:17 ❯ kubectl apply -f deploy/crds/k8gb.absa.oss_v1beta1_gslb_cr.yaml
The Gslb "test-gslb" is invalid: spec.ingress.rules.http: Required value
```

somaritane (changes requested):
It would be good to cover this change with tests to simulate the malformed spec.

deploy/crds/k8gb.absa.oss_gslbs_crd.yaml
```diff
@@ -65,6 +65,8 @@ spec:
                       under a specified host to the related backend services. Incoming
                       requests are first evaluated for a host match, then routed to
                       the backend associated with the matching IngressRuleValue.
+                    required:
+                    - http
```
kuritka, line 69: Hi, love that.

The code looks like this:
```go
	for _, gslb := range gslbList.Items {
				for _, rule := range gslb.Spec.Ingress.Rules {
					for _, path := range rule.HTTP.Paths {
						if path.Backend.ServiceName == a.Meta.GetName() {
							gslbName = gslb.Name
						}
					}
				}
```
My question: The code will work if `HTTP.Paths`  is empty (for cycle will be skipped), but isn't `Path` mandatory as well?  @Tim, @k0da

k0da: @kuritka Thanks, definetely we need that too.

```shell
09:04:27 ❯ kubectl apply -f terratest/examples/invalid.yaml
The Gslb "invalid-gslb" is invalid: spec.ingress.rules.http.paths: Required value
```

somaritane: As I mentioned earlier, nor `http` neither `paths` are not enforced by Ingress API, and k8gb should stick to that, if it wants to resemble Ingress-like experience for end-users.
k8gb supports indirect `gslb` object creation Ingress object with k8gb-specific annotations, and you won't get any errors if `http` or `path` sections are missing in Ingress manifest.
In that sense it is better to be prepared for missing spec sessions in the code instead of imposing restrictions in CRD.

somaritane (changes requested):

kuritka (dismissed):

deploy/crds/k8gb.absa.oss_gslbs_crd.yaml
k0da: But it is required for GSLB IIUC.

somaritane: In this case gslb creation can be skipped with log message, like in case when no glsb manifests are found or when `glsb.Name` is empty

k0da: Ok, we can go this way instead. @ytsarev @kuritka @donovanmuller  what do you think about that?

kuritka: Direct CRD modification is an elegant way to validate on the other hand, there was a decision to follow standard ingress definition.
When I hastily approved, I haven't realized the only input point in our architecture is depresolver, which has two parts - config and spec. Unfortunately, I'm afraid we can't break the validation into more places. It would be difficult to find out where the various inputs are being validated / tested.

somaritane: Actually, I was not correct about `spec.ingress.rules.http.paths`. It mandatory field by Ingress spec:
```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    k8gb.io/strategy: roundRobin
  name: invalid-ingress
spec:
  rules:
  - host: invalid-ingress.example.com
    http:
      paths:
```
```sh
k apply -f invalid-ingress.yaml --validate=true
error: error validating "invalid-ingress.yaml": error validating data: ValidationError(Ingress.spec.rules[0].http): missing required field "paths" in io.k8s.api.extensions.v1beta1.HTTPIngressRuleValue; if you choose to ignore these errors, turn validation off with --validate=false
```
So it's OK to impose this rule on CRD level:
```yaml
required:
- paths
```

but `spec.ingress.rules.http` is still not required:
```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    k8gb.io/strategy: roundRobin
  name: invalid-ingress
spec:
  rules:
  - host: invalid-ingress.example.com
    http:
```
```sh
k apply -f invalid-ingress.yaml
ingress.extensions/invalid-ingress created
```

k0da:
I ended up validating .spec.ingress.rules.host as broken rule can sneak in through Ingress object.

controllers/depresolver/depresolver_spec.go
```diff
@@ -24,15 +25,21 @@ func (dr *DependencyResolver) ResolveGslbSpec(ctx context.Context, gslb *k8gbv1b
 		if strategy.SplitBrainThresholdSeconds == 0 {
 			strategy.SplitBrainThresholdSeconds = predefinedStrategy.SplitBrainThresholdSeconds
 		}
-		dr.errorSpec = dr.validateSpec(strategy)
+		dr.errorSpec = dr.validateStrategy(strategy)
```
somaritane, line 28: @kuritka , what was the initial intention of this method, spec validation in general or just gslb strategy parameters validation?

controllers/depresolver/depresolver_spec.go
```diff
@@ -24,15 +25,21 @@ func (dr *DependencyResolver) ResolveGslbSpec(ctx context.Context, gslb *k8gbv1b
...
 		if dr.errorSpec == nil {
 			dr.errorSpec = dr.client.Update(ctx, gslb)
 		}
+		// check for missing .spec.ingress.host.http
```
somaritane, line 32: Imo this code is self-explanatory, no comments needed.
Or you can consider extracting this code into something like `ValidateIngressSpec` method, in this case you definitely don't need this comment :)

main.go
```diff
@@ -63,7 +63,7 @@ func main() {
 			"Enabling this will ensure there is only one active controller manager.")
 	flag.Parse()
 
-	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
+	ctrl.SetLogger(zap.New(zap.UseDevMode(false)))
```
somaritane, line 66: 👍  This changes logging format in positive way (especially for use with logging solutions like ELK), but it would be good to have this change in a separate PR, in order to make it visible in release history, plus it's a bit out of scope of this PR.

somaritane (changes requested):

controllers/gslb_controller.go
```diff
@@ -190,6 +190,10 @@ func (r *GslbReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			gslbName := ""
 			for _, gslb := range gslbList.Items {
 				for _, rule := range gslb.Spec.Ingress.Rules {
+					if rule.HTTP == nil {
```
kuritka, line 193: depresolver is always providing **valid** instance. In the future `if rule.HTTP == nil` can be multiplied to many places in the code. To avoid additional checks in the code, we can handle in depresolver by using Empty instance pattern. HTTP has only property Paths, so in case of `HTTP==nil` we can instantiate `HTTP` with empty `Paths` within depresolver. This will be valid with no checks:

```
for _, path := range rule.HTTP.Paths {
```

Mentioned depresolver test would cover this functionality: if input `HTTP == nil`, then output has HTTP instance with empty Paths

kuritka (changes requested):
good start, please see my comments

ytsarev (changes requested):
Something conceptually wrong here. We should not validate the embedded Ingress object which is a standard k8s primitive. 
If k8gb fails on missing Ingress fields we should modify the controller to handle such cases gracefully, maybe providing the log message to the user.

PR #311, opened 2021-02-20
Sort externalTargets queried from DNS

k0da:
In case CoreDNS runs with loadbalancer round_robin plugin enabled, k8gb
controller gets shuffled results while computing externalTarget list.
This leads to DNSEndpoint object update despite the fact list is the
same. Lets make list sorted and reproducible to avoid unnescessary
reconcilation loops.

### DIFF
--- a/controllers/dnsupdate.go
+++ b/controllers/dnsupdate.go
@@ -2,6 +2,7 @@ package controllers
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


kuritka (commented):
please, sort IP address list in documentation 
 - https://github.com/AbsaOSS/k8gb/blob/master/docs/local.md#verify-installation

otherwise LGTM

k0da:
@kuritka done

kuritka (approved):

ytsarev (approved):

PR #312, opened 2021-02-22
Infoblox, heavy load fixed

kuritka:
Rancher includes annotation into Ingress object.
```json
  "field.cattle.io/publicEndpoints": "[{\"addresses\":[\"22.240.40.188\"],\"port\":80,\"protocol\":\"HTTP\",\"serviceName\":\"test-gslb:frontend-podinfo\",\"ingressName\":\"test-gslb:test-gslb-failover\",\"hostname\":\"failover.k8gb-test-nonprod.bcp.absa.co.za\",\"path\":\"/\",\"allNodes\":true}]",
```
Original code modified ingress object back into desired state, but when reconciliation ended, rancher included new annotation again.
As result was annotation included, removed, included, removed, etc...
Such behavior led to executing reconciliation loop immediately instead of specified delay (RECONCILE_REQUEUE_SECONDS).

 - fix reconciliation issue (see `ingress.go` changes)
 - rename function `ensureIngres` -> `saveIngress`
 - ReconcileResultHandler
   - instead of `return ctrl.Result{},err` we call `return r.RequeueError(err)`,`return r.Requeue()` etc..

 - Some functions like `saveIngress`, `SaveEndpoint` returned `(*ctrl.Result, error)`  as result, but they shouldn't be responsible for reconciliation status.
   Now they return only `error` and only `gslb_controller` handles Reconciliation.


controllers/ingress.go
```diff
@@ -51,26 +51,40 @@ func (r *GslbReconciler) ensureIngress(instance *k8gbv1beta1.Gslb, i *v1beta1.In
...
-	if err != nil {
-		// Update failed
-		log.Error(err, "Failed to update Ingress", "Ingress.Namespace", found.Namespace, "Ingress.Name", found.Name)
-		return &reconcile.Result{}, err
+	if !ingressCompare(found, i) {
+		found.Spec = i.Spec
+		found.Annotations = i.Annotations
+		err = r.Update(context.TODO(), found)
+		if errors.IsConflict(err) {
+			r.Log.Info("CONFLICT, Ingress updated by third-party")
```
ytsarev, line 70: It will still make the actual update, correct? If yes, then the log message is misleading and *looks* like an error

kuritka: Thx for review and good question:

The heartbeat of this pull-request is this:
```go
if !ingressCompare(found, i) {
		found.Spec = i.Spec
		found.Annotations = i.Annotations
		err = r.Update(context.TODO(), found)
		if err != nil {
			// Update failed
			log.Error(err, "Failed to update Ingress", "Ingress.Namespace", found.Namespace, "Ingress.Name", found.Name)
			return err
		}
	}
```
It says, *hey, if you change GSLB annotations, then we revert your changes to the desired state. If you change no GSLB annotations, we don't care.*.

The problem is, between the time that we get and attempt to update a resource, something else has updated the resource, and the version we have is "stale" and it refuses to let you update it, which obviously leads to `r.Update(context.TODO(), found)` returns error -  detected by `errors.IsConflict(err)`.

So in such case we can log info message and return nil. Reconciliation queue will be executed immediately, because and next run would be ok. 

- misleading error: yes I cand fix. What about `re-run reconciliation due to ingress conflict` ?

I just realized, to make it even cleaner we should keep third party annotations on ingress and update only ours if needed:


```go
	if !ingressCompare(found, i) {
		found.Spec = i.Spec
		found.Annotations = repairAnnotations(i.Annotations) // <<< keep original annotations but alter / update our annotations
		err = r.Update(context.TODO(), found)
		if errors.IsConflict(err) {
			r.Log.Info("CONFLICT, Ingress updated by third-party")
			return nil
		}
		if err != nil {
			// Update failed
			log.Error(err, "Failed to update Ingress", "Ingress.Namespace", found.Namespace, "Ingress.Name", found.Name)
			return err
		}
	}
```

somaritane: @kuritka , how about this message:
```golang
r.Log.Info("Ingress has been modified outside of controller, retrying reconciliation", "Ingress.Namespace", found.Namespace, "Ingress.Name", found.Name)
```

controllers/internal/utils/result.go
```diff
@@ -0,0 +1,45 @@
+package utils
```
somaritane, line 1: Consider renaming `controllers/internal/utils/result.go` to `controllers/internal/utils/reconciler_result.go` to better reflect the intention of this package.

controllers/internal/utils/annotations_test.go
```diff
@@ -0,0 +1,76 @@
+package utils
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+var a2 = map[string]string{"k8gb.io/primary-geotag": "270", "k8gb.io/strategy": "failover"}
```
somaritane, line 9: Consider using geotag names for broader community:
```suggestion
var a2 = map[string]string{"k8gb.io/primary-geotag": "eu", "k8gb.io/strategy": "failover"}
```

somaritane (changes requested):
Nice work! It should help to reduce load on edgeDNS providers in general, not only Infoblox.
lgtm, just few comments from me.

controllers/internal/utils/annotations_test.go
kuritka: @somaritane Good idea

somaritane (approved):

controllers/internal/utils/annotations.go
```diff
@@ -0,0 +1,17 @@
+package utils
+
+// SaveAnnotations adds or updates annotations from a2 from a1 and returns merge
+func SaveAnnotations(a1 map[string]string, a2 map[string]string) map[string]string {
```
somaritane, line 4: consider renaming to emphasize merge priority:
```suggestion
func MergeAnnotations(target map[string]string, source map[string]string) map[string]string {
```
or
```suggestion
func MergeAnnotations(to map[string]string, from map[string]string) map[string]string {
```

controllers/ingress.go
```diff
@@ -51,26 +53,41 @@ func (r *GslbReconciler) ensureIngress(instance *k8gbv1beta1.Gslb, i *v1beta1.In
...
+			log.Error(err, "Failed to update Ingress", "Ingress.Namespace", found.Namespace, "Ingress.Name", found.Name)
+			return err
+		}
 	}
 
-	return nil, nil
+	return nil
+}
+
+func ingressCompare(found *v1beta1.Ingress, i *v1beta1.Ingress) bool {
```
somaritane, line 86: Consider renaming
```suggestion
func ingressEqual(ing1 *v1beta1.Ingress, ing2 *v1beta1.Ingress) bool {
```
as we're testing for equality, not comparing entities.

somaritane (changes requested):

controllers/internal/utils/annotations.go
```diff
@@ -0,0 +1,17 @@
+package utils
+
+// MergeAnnotations adds or updates annotations from a2 from a1 and returns merge
```
somaritane, line 3: Just a little one :)
```suggestion
// MergeAnnotations adds or updates annotations from source to target and returns merge result
```

somaritane (changes requested):

controllers/ingress.go
kuritka: Just thinking about new package `metav1` next to the `utils` package and reuse ApiMachinery
https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#SetMetaDataAnnotation but in k8gb


Having `metav1.AnnotationsEqual(obj1 *ObjectMeta, obj2 *ObjectMeta)`, `metav1.MergeAnnotations(obj1 *ObjectMeta, obj2 *ObjectMeta)`

kuritka: Then we can reuse for any kind. (Maybe move to AbsaOSS/gopkg)

somaritane: Like here https://github.com/AbsaOSS/k8gb/pull/314 ?

kuritka: @somaritane yes

somaritane: Might be good in general along with reusing helpers from https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1.
I don't have strong opinion here, perhaps we can extract it later when we find more use in other projects.
@ytsarev, @k0da, wdyt?

ytsarev: Good idea, though I wouldn't block this specific PR with it. We should log the Issue for that though

kuritka: ok, will provide in later.

somaritane (approved):
lgtm!

PR #313, opened 2021-02-22
Validate spec.ingress.http.path

k0da:
Require .spec.ingress.http.path as it is required for gslb to function.
If broken Gslb submitted, it will be rejected by openapi v3 validation.
Same will happen when using Ingress annotations, we won't create broken
Gslb object but log an error message instead.

### DIFF
--- a/controllers/gslb_controller.go
+++ b/controllers/gslb_controller.go
@@ -257,7 +257,10 @@ func (r *GslbReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		}
 
 		log.Info(fmt.Sprintf("Creating new Gslb(%s) out of Ingress annotation", gslb.Name))
-		_ = c.Create(context.Background(), gslb)
+		err = c.Create(context.Background(), gslb)
+		if err != nil {
+			log.Error(err, "Glsb creation failed")
+		}
 	}
 	ingressMapFn := handler.ToRequestsFunc(
 		func(a handler.MapObject) []reconcile.Request {
--- a/terratest/examples/broken-gslb.yaml
+++ b/terratest/examples/broken-gslb.yaml
@@ -1,0 +1,17 @@
+apiVersion: k8gb.absa.oss/v1beta1
+kind: Gslb
+metadata:
+  name: broken-test-gslb1
+spec:
+  ingress:
+    rules:
+      - host: terratest-failover.cloud.example.com
+        https:
+          paths:
+          - backend:
+              serviceName: frontend-podinfo # Gslb should reflect Healthy status and create associated DNS records
+              servicePort: http
+            path: /
+  strategy:
+    type: failover
+    primaryGeoTag: eu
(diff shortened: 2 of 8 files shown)


k0da:
@ytsarev @somaritane - as we discussed.

@kuritka It is cheap to cut off broken boject by api validation and not coplicate depresolver with essential data requirement. As for Ingress: we'll catch errors by same api validator, and throw an error, so broken object won't be created and controller won't observe it.

kuritka:
If app doesn't start because `http` and `http.paths` are required, then all is ok.

kuritka (approved):

k0da:
fixes https://github.com/AbsaOSS/k8gb/issues/296

controllers/gslb_controller.go
```diff
@@ -257,7 +257,10 @@ func (r *GslbReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		}
 
 		log.Info(fmt.Sprintf("Creating new Gslb(%s) out of Ingress annotation", gslb.Name))
-		_ = c.Create(context.Background(), gslb)
+		err = c.Create(context.Background(), gslb)
+		if err != nil {
+			log.Error(err, "Glsb create failed:")
```
somaritane, line 262:
```suggestion
			log.Error(err, "Glsb creation failed")
```

terratest/examples/broken-gslb.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: k8gb.absa.oss/v1beta1
+kind: Gslb
+metadata:
+  name: broke-test-gslb
```
somaritane, line 4:
```suggestion
  name: broken-test-gslb
```

terratest/examples/broken-gslb.yaml
```diff
@@ -0,0 +1,17 @@
...
+spec:
+  ingress:
+    rules:
+      - host: terratest-failover.cloud.example.com
+        https:
```
somaritane, line 9: should we also test for missing `http` in addition to misspelled `http(s)`?

somaritane (changes requested):
Looking good, just few comments.

terratest/examples/broken-gslb.yaml
k0da: sure, I can add it

k0da: @somaritane please check

somaritane (approved):
lgtm!

terratest/examples/broken-ingress-annotation.yaml
```diff
@@ -0,0 +1,16 @@
...
+kind: Ingress
+metadata:
+  annotations:
+    k8gb.io/strategy: failover
+    k8gb.io/primary-geotag: eu
+  name: broken-test-gslb-annotation-failover
+spec:
+  rules:
+  - host: notfound.cloud.example.com
+    #http:
```
ytsarev, line 11: what is it for?

ytsarev (commented):
looks good just some comment on possible commented out config leftover

terratest/examples/broken-ingress-annotation.yaml
k0da: Just a quick way to make it invalid for gslb controller, but keeping it visible.

ytsarev (approved):

PR #315, opened 2021-02-23
disable CoreDNS cache

k0da:
cache plugin requires TTL to be set, which overrides TTL of a served
entry. In k8gb setup we allow dnsTTLSeconds to be set throguh gslb spec,
witch cache enabled, setting will be ignored.

See: https://coredns.io/plugins/cache/

### DIFF
--- a/chart/k8gb/values.yaml
+++ b/chart/k8gb/values.yaml
@@ -39,8 +39,6 @@ coredns:
     - zone: .
     port: 53
     plugins:
-    - name: cache
-      parameters: 30
     - name: errors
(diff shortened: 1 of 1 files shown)


ytsarev:
@k0da already change the description to 'disable cache'?

ytsarev (changes requested):
Please update the PR/commit message to reflect reality otherwise LGTM

k0da:
@ytsarev done

ytsarev (approved):
Perfect

PR #327, opened 2021-02-28
Solve fsGroup issue

### LINKED ISSUE #293: [helm chart] fsGroup not a valid securityContext field
this seems to be an invalid field: https://github.com/AbsaOSS/k8gb/blob/7c9e66221b8c74997da1b8a10a9df553879a4d42/chart/k8gb/templates/external-dns/external-dns-ns1.yaml#L47

based on https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#securitycontext-v1-core

perhaps it should have been `runAsGroup`

Either way I'd recommend setting this at the pod level https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#podsecuritycontext-v1-core and making the values parametric as for example OCP is very specific about these kinds of things.

ytsarev:
* Respect recent api changes and move most of securityContext to pod level,
  only readOnlyRootFilesystem is stayed on container one

* Solves
```
Error: UPGRADE FAILED: error validating "": error validating data: ValidationError(Deployment.spec.template.spec.containers[0].securityContext): unknown field "fsGroup" in io.k8s.api.core.v1.SecurityContext
```
on eks reference deployment

* Fixes #293

### DIFF
--- a/chart/k8gb/templates/external-dns/external-dns-ns1.yaml
+++ b/chart/k8gb/templates/external-dns/external-dns-ns1.yaml
@@ -15,6 +15,10 @@ spec:
         app: external-dns-ns1
     spec:
       serviceAccountName: external-dns
+      securityContext:
+        fsGroup: 65534 # For ExternalDNS to be able to read Kubernetes and AWS token files
+        runAsUser: 1000
+        runAsNonRoot: true
       containers:
       - name: external-dns
         image: {{ .Values.externaldns.image }}
@@ -41,8 +45,5 @@ spec:
             memory: "128Mi"
             cpu: "500m"
         securityContext:
-          runAsUser: 1000
-          runAsNonRoot: true
           readOnlyRootFilesystem: true
-          fsGroup: 65534 # For ExternalDNS to be able to read Kubernetes and AWS token files
 {{ end }}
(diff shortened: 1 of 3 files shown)


kuritka (approved):

chart/k8gb/templates/external-dns/external-dns-ns1.yaml
```diff
@@ -15,6 +15,10 @@ spec:
         app: external-dns-ns1
     spec:
       serviceAccountName: external-dns
+      securityContext:
```
somaritane, line 18: How about making securityContext parameterized though `values.yaml` file, so it can be configured and reused in all external-dns deployments?
```suggestion
      securityContext:
{{- toYaml .Values.externaldns.securityContext | nindent 8 }}
```

somaritane (changes requested):
What about operator deployment, do we also want to lift its security context to pod level too?
https://github.com/AbsaOSS/k8gb/blame/master/chart/k8gb/templates/operator.yaml#L31

chart/k8gb/templates/external-dns/external-dns-ns1.yaml
ytsarev: @somaritane good idea, implemented, please check

ytsarev:
@somaritane operator has 1 container, securityContext is there, no fsGroup around. So I see no real reason to move the context there

somaritane (approved):
lgtm

kuritka (approved):

PR #332, opened 2021-03-01
Make k8gb demo curl script ready for local invocation

ytsarev:
* Add `--local` param to omit nameserver injection as we doing
  it in a demo pod when no edgeDNS provider configured
* Useful for live demo sessions, e.g. against EKS k8gb reference
  setup


kuritka (approved):

somaritane (approved):
lgtm

PR #334, opened 2021-03-01
Automate releases

k0da:
* Adds separate Dockerfile for each architecture to keep base image
      unchanged.
    * Brings github_changelog_generator into a release pipeline
    * Add .goreleaser.yaml with go build and amd64/arm64 images with docker
      common manifest
    * Introduce pipeline for tag event, where docker artifacts will be
      built.
    * Rename helm pipeline (release) into helm_publish

### DIFF
--- a/.github/workflows/helm_publish.yaml
+++ b/.github/workflows/helm_publish.yaml
@@ -1,0 +1,55 @@
+name: Helm Publish
+
+on:
+  release:
+    types:
+      - published
+
+jobs:
+  publish:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: dave-mcconnell/helm-gh-pages-microservices@master
+        with:
+          access-token: ${{ secrets.CR_TOKEN }}
+          source-charts-folder: 'chart'
+          destination-repo: absaoss/k8gb
+          destination-branch: gh-pages
+      - name: Create single node k3s cluster
+        uses: AbsaOSS/k3d-action@v1.3.1
+        with:
+          cluster-name: "test-gslb1"
+          use-default-registry: false
+          args: >-
+            -p "80:80@agent[0]"
+            -p "443:443@agent[0]"
+            -p "5053:53/udp@agent[0]"
+            --agents 1
+            --no-lb
+            --network k3d-action-bridge-network
+            --k3s-server-arg "--no-deploy=traefik,servicelb,metrics-server"
+      - name: Smoke test helm installation
+        run: |
+          helm repo add k8gb https://absaoss.github.io/k8gb/
+          helm repo update
+          helm -n k8gb upgrade -i k8gb k8gb/k8gb --wait --create-namespace --version=$(make version)
+      - uses: heinrichreimer/github-changelog-generator-action@v2.1.1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          project: k8gb
+      - name: Commit changelog
+        run: |
+          git config user.name 'Github Action'
+          git config user.email 'action@users.noreply.github.com'
+          git add CHANGELOG.md
+          git commit -m "Update Offline Changelog"
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v3
+        with:
+          title: "Update Offline Changelog"
+          branch: offline_changelog
+          delete-branch: true
+          token: ${{ secrets.GITHUB_TOKEN }}
--- a/.gitignore
+++ b/.gitignore
@@ -109,3 +109,6 @@ chart/k8gb/charts/*.tgz
 
 # Ignore testing kubeconfigs
 kubeconfig*
+
+# Ignore changes from github_changelog_generator action
+changes
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,0 +1,39 @@
+before:
+  hooks:
+    - go mod download
+builds:
+  - env:
+      - CGO_ENABLED=0
+    id: k8gb
+    main: ./main.go
+    binary: bin/manager
+    goos:
+      - linux
+    goarch:
+      - amd64
+      - arm64
+dockers:
+- image_templates:
+  - "absaoss/k8gb:{{ .Tag }}-amd64"
+  use_buildx: false
+  dockerfile: Dockerfile-gr-amd64
+  build_flag_templates:
+  - "--platform=linux/amd64"
+- image_templates:
+  - "absaoss/k8gb:{{ .Tag }}-arm64"
+  use_buildx: false
+  goarch: arm64
+  dockerfile: Dockerfile-gr-arm64
+  build_flag_templates:
+  - "--platform=linux/arm64"
+docker_manifests:
+- name_template: absaoss/k8gb:{{ .Tag }}
+  image_templates:
+  - absaoss/k8gb:{{ .Tag }}-amd64
+  - absaoss/k8gb:{{ .Tag }}-arm64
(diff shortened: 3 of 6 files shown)


ytsarev:
@somaritane keen to release `v0.7.6` using this one?

.goreleaser.yml
```diff
@@ -0,0 +1,43 @@
...
+checksum:
+  name_template: 'checksums.txt'
+snapshot:
+  name_template: "{{ .Tag }}-next"
+changelog:
+  sort: asc
+  filters:
+    exclude:
+      - '^docs:'
+      - '^test:'
```
ytsarev, line 43: Why exclude them?

.goreleaser.yml
```diff
@@ -0,0 +1,43 @@
...
+  - "absaoss/k8gb:{{ .Tag }}-arm64"
+  use_buildx: false
+  goarch: arm64
+  dockerfile: Dockerfile-gr-arm64
+  build_flag_templates:
+  - "--platform=linux/arm64"
+docker_manifests:
+- name_template: absaoss/k8gb:{{ .Tag }}
+  image_templates:
+  - absaoss/k8gb:{{ .Tag }}-amd64
```
ytsarev, line 32: Will the final tag include `$arch` suffix?

If yes, it will break this logic https://github.com/AbsaOSS/k8gb/blob/master/chart/k8gb/templates/operator.yaml#L29

.goreleaser.yml
```diff
@@ -0,0 +1,43 @@
...
+  - absaoss/k8gb:{{ .Tag }}-arm64
+checksum:
+  name_template: 'checksums.txt'
+snapshot:
+  name_template: "{{ .Tag }}-next"
```
ytsarev, line 37: Can you elaborate on the logic here?

ytsarev (commented):
Looks super fancy, just some clarifications required

.goreleaser.yml
k0da: This is upstream behaviour and ignored by providing --release-notes flag, https://goreleaser.com/customization/release/#custom-release-notes

somaritane:
> @somaritane keen to release `v0.7.6` using this one? 

@ytsarev, Oops, I've just released `v0.7.6` now, but `v0.7.7` - definitely :)

.goreleaser.yml
k0da: It follows same strategy as of now. We build two images imag-{.Arch} then wrap it with single manifest name_template: absaoss/k8gb:{{ .Tag }}

.goreleaser.yml
k0da: you can run goreleaser --snapshot, it will build image: image:tag-next and publish it

k0da: I was wrong on publish: https://goreleaser.com/customization/snapshots/#how-it-works

.goreleaser.yml
k0da: So it won't break ``Deployment`` manifest. absaoss/k8gb:{{ .Tag }} will resolve into absaoss/k8gb:{{ .Tag }}-amd64 if kubernetes worker is amd64

.goreleaser.yml
ytsarev: I get it, just don't think we should exclude docs or tests from the changelog :)

.goreleaser.yml
ytsarev: Cool!

ytsarev:
@k0da will it create Draft Release or Release right away?

k0da:
# If set to true, will not auto-publish the release.
  # Default is false.
  # draft: false
Release straighaway. I think we should start with draft first

somaritane:
>  I think we should start with draft first
Makes sense. `github-changelog-generator` uses GH issue's tags for changelog generation, which might required 2nd pair of eyes to check that issues are correctly categorized

k0da:
amended

.goreleaser.yml
```diff
@@ -0,0 +1,41 @@
...
+changelog:
+  sort: asc
```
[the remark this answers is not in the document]
k0da: Yes, I'll drop it.

k0da:
> @k0da what about we create some fake releases and see how it works end-to-end?

I can tag local commit with v0.7.7-alpha1 and run releaser manually?

ytsarev (dismissed):
https://github.com/AbsaOSS/k8gb/releases/tag/untagged-d1d7bd1b09241abe6eb0 looks great, so let's accommodate the cool pipes :)

somaritane (dismissed):
lgtm

.gitignore
```diff
@@ -109,3 +109,6 @@ chart/k8gb/charts/*.tgz
 
 # Ignore testing kubeconfigs
 kubeconfig*
+
+# Ignore changes from github_changelog_generator action
+changes
```
ytsarev, line 114: How do we update `CHANGELOG.md` in this scenario?

k0da: same way as before, post release commit (post release github_changelog_generator on maintainers machine and open PR). 
During GH run only temporary diff is generated into gitingored file.

ytsarev: But before local github_changelog_generator diff was the input for release. Now it will be required to run it locally post release and commit? It might lead to deviations

k0da: If I look at last v0.7.6 release, offline changelog was commited at 18:00, and release was done on 15:06. I see no deviation.

ytsarev: yeas, because offline changelog is the source for release text

k0da: Is it worth to sacrafice automation in order to have "offline" change log?
you have two tags, and tool to work with those tags. ATM we run tool get output and paste it into "Release".
I propose to run tool as part of release draft pipeline to format a release. Then run tool with same input to commit post release offline changelog. Sorry I only see an improvement here.

Additionally draft can be changed the same way from "offline" changelog.

ytsarev: I am not talking about sticking to current scenario, more thinking of avoiding pitfalls with the new one. Can we can write to CHANGELOG.md in post-release pipeline and commit it automatically?

k0da: This option can be explored for sure.

k0da: @ytsarev ammended

ytsarev: @k0da very cool, thanks!

.github/workflows/helm_publish.yaml
```diff
@@ -0,0 +1,52 @@
...
+on:
+  release:
+    types:
+      - published
+
+jobs:
+  publish:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v1
```
ytsarev, line 12: Please switch to v2

ytsarev (changes requested):

.github/workflows/helm_publish.yaml
k0da: done

k0da:
so I tested pull-request action here https://github.com/k0da/terraform-controller/runs/2020895388?check_suite_focus=true
results into https://github.com/k0da/terraform-controller/pull/3

k0da:
IMO any way automated PR can be closed any time. So this is quite harmless improvement

ytsarev (approved):
lgtm

somaritane (approved):
lgtm, can't wait for new release :)🤞

k0da:
@somaritane @ytsarev I've ammended fetch-depth param to checkout action to preserve full history for github_changelog_generator

ytsarev (approved):

somaritane (approved):

PR #336, opened 2021-03-01
k8gb extensions for metav1ext

kuritka:
refactor annotations into `github.com/AbsaOSS/k8gb/controllers/internal/utils`.
 - MergeAnnotations
 - ContainAnnotations

### DIFF
--- a/controllers/ingress.go
+++ b/controllers/ingress.go
@@ -61,9 +61,9 @@ func (r *GslbReconciler) saveIngress(instance *k8gbv1beta1.Gslb, i *v1beta1.Ingr
 	}
 
 	// Update existing object with new spec and annotations
-	if !ingressEqual(found, i) {
+	if !(utils.ContainsAnnotations(&found.ObjectMeta, &i.ObjectMeta) && reflect.DeepEqual(found.Spec, i.Spec)) {
 		found.Spec = i.Spec
-		found.Annotations = utils.MergeAnnotations(found.Annotations, i.Annotations)
+		utils.MergeAnnotations(&found.ObjectMeta, &i.ObjectMeta)
 		err = r.Update(context.TODO(), found)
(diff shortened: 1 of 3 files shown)


controllers/ingress.go
```diff
@@ -4,9 +4,9 @@ import (
 	"context"
 	"reflect"
 
-	"github.com/AbsaOSS/k8gb/controllers/internal/utils"
-
 	k8gbv1beta1 "github.com/AbsaOSS/k8gb/api/v1beta1"
+	metav1ext "github.com/AbsaOSS/k8gb/controllers/internal/meta/v1"
```
ytsarev, line 8: I am a bit concerned to use clashing name with the upstream even if it is internal. 
@kuritka what is the motivation to do so?

ytsarev (commented):
At wich point do you plan to move it out to external package?

controllers/ingress.go
kuritka: The original idea was to have a same structure as `k8s.io/apimachinery/pkg/apis/meta/v1` which contains annotation helpers. I moved back to utils and amended changes.

kuritka:
> At which point do you plan to move it out to external package?
I wanted to make separate PR for that. First, move utils to gopkg, then update k8gb

controllers/ingress.go
somaritane: @kuritka can you please amend commit and PR message, so it reflect renaming to `internal/utils/annotations"?

ytsarev (changes requested):
@kuritka can we please update PR header and commit message reflecting the current state of things as it is staying under `utils` ?

kuritka:
@ytsarev, done, I will move to `gopkg` soon
thx

kuritka:
k8gb extensions for metav1ext,  amended

PR #343, opened 2021-03-03
Add Apache 2 license header to every Go file

ytsarev:
* Add Apache 2 headers
* Add incredible piece of automation for header injection


hack/header_file
```diff
@@ -0,0 +1,16 @@
+/*
```
k0da, line 1: Can't we sync it with ./hack/boilerplate.txt and use same header here and in controller-gen?

kuritka:
I dind't study deepply, but there are tools doing that:  https://github.com/cloudposse/copyright-header. Maybe we can manipulate with headers before GOLINT

hack/header_file
ytsarev: @k0da thanks, i've made boilerplate a single source of headers file

ytsarev:
@kuritka I definitely don't want to employ super sophisticated machinery with lots of rubygem dependencies here.

Also tried https://github.com/google/addlicense . It does pretty much the same stuff as the tiny shell script provided.

So I suggest we stick with the simplest version of it which does not bring in any new dependencies

kuritka (dismissed):

kuritka (approved):

somaritane (approved):

PR #353, opened 2021-03-05
Update apiextensions to v1

k0da:
This updates apiextensions.k8s.io/v1beta1 to apiextensions.k8s.io/v1


ytsarev (commented):
@k0da just to double-check, everything under `chart/k8gb/templates/k8gb.absa.oss_gslbs_crd.yaml` is purely a result of generation, no manual modifications were required?

k0da:
@ytsarev The flow is change controller-gen options to produce v1 into config/crd/bases/k8gb.absa.oss_gslbs.yaml, modify it to validate http and paths fields of an IngressSpec and copy it over into chart/k8gb/templates

ytsarev (changes requested):
@k0da Please make the generation and manual modification in form of separate commits so it is clear what is actually happening in the git history.

Apparently we also need to get rid of duplicate in `deploy/crds/k8gb.absa.oss_gslbs_crd.yaml`

Is it possible to symlink  `chart/k8gb/templates/k8gb.absa.oss_gslbs_crd.yaml` and `config/crd/bases/k8gb.absa.oss_gslbs.yaml` to avoid manual copy over? (I'm not sure if it will properly work during packaging)

@somaritane surprise surprise, we hit the case when we need a Merge commit containing multiple logically separated commits ;)

somaritane:
> @somaritane surprise surprise, we hit the case when we need a Merge commit containing multiple logically separated commits ;) 
> @ytsarev yet it still looks like exceptional case rather than common one :) 
This PR can be split into two separate. 
Or you can temporary lift "squash only" cap from merge trigger for this particular case.

k0da:
Reworked in a way:
* dropped config/crd
* generate straight into chart/k8gb/templates

Makefile
```diff
@@ -410,7 +410,7 @@ define generate
 endef
 
 define manifest
-	$(call controller-gen,crd:trivialVersions=true rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases)
+	$(call controller-gen,crd paths="./..." output:crd:artifacts:config=chart/k8gb/templates/)
```
ytsarev, line 413: is crd version v1 is the default in `sigs.k8s.io/controller-tools/cmd/controller-gen@v0.5.0` ? I recall explicit version statement in the first version of this PR

k0da: yes, it is. Since v1beta1 is deprecated quite a while ago. We can have it explicit though

ytsarev (changes requested):
Apparently
```
deploy/crds/k8gb.absa.oss_gslbs_crd.yaml
```
still should be removed

k0da:
done

ytsarev (changes requested):
Probably last bit, sorry, just remembered.
We should remove associated kubelinter exclusion https://github.com/AbsaOSS/k8gb/blob/master/.kube-linter.yaml#L3

k0da:
@ytsarev not yet. DNSEndpoint is still v1beta1

ytsarev:
@k0da indeed 👍

ytsarev (approved):
Looks great

k0da:
DNSEndpoint addresed in https://github.com/kubernetes-sigs/external-dns/pull/2001

PR #355, opened 2021-03-08
Reduce load on test setup

k0da:
This commit reduces number of k3d workers from 3 to 1 and number of
podinfo replicas from 2 to 1. That should increase requirements for test
setup and speed of tests.


kuritka (approved):
lgtm

somaritane (approved):

ytsarev (approved):
It will become less illustrative from the demo perspective but as we are mostly making presentations using aws reference setup the reduction makes sense

Issue #331, opened 2021-03-01
Logger Enhancements


kuritka:
Because k8gb is constantly expanding, the requirements for monitoring the application are also increasing.
I'd like to mention a few of them:

 - DBG: There is no need to log everything all the time. Some messages only interests us when we debug the system or check its behavior. Cleaning log from too detailed information makes monitoring easier. On the other hand, when debugging system we need as detailed info as possible. I'd like to introduce boolean environment variable called `LOG_DEBUGGING`. When true, then prints `ERR + DBG + INFO + WRN`, otherwise `ERR + INFO + WRN`. 

- PROD / NONPROD : would be nice to distinguish between PROD and NONPROD systems. It could be Date format (for prod we want to show DateTime but only Time for nonprod). It could be format of message (plain print for NONPROD, json for PROD). Should distinguish by environment variable. For now bool variable `LOG_JSON`

- ERRORS handling: instead of printing entire stack trace including GO libraries, we should trace only stack related to k8gb

- please add your requirements for logging if any ?

To fulfill this purpose, I chose [zerolog](https://github.com/rs/zerolog) who has everything I mentioned earlier. In addition it has very high performance:

Library | Time | Bytes Allocated | Objects Allocated
-- | -- | -- | --
zerolog | 767 ns/op | 552 B/op | 6 allocs/op
⚡ zap | 848 ns/op | 704 B/op | 2 allocs/op
⚡ zap (sugared) | 1363 ns/op | 1610 B/op | 20 allocs/op
go-kit | 3614 ns/op | 2895 B/op | 66 allocs/op
lion | 5392 ns/op | 5807 B/op | 63 allocs/op
logrus | 5661 ns/op | 6092 B/op | 78 allocs/op
apex/log | 15332 ns/op | 3832 B/op | 65 allocs/op
log15 | 20657 ns/op | 5632 B/op | 93 allocs/op

Log a static string, without any context or printf-style templating:
Library | Time | Bytes Allocated | Objects Allocated
-- | -- | -- | --
zerolog | 50 ns/op | 0 B/op | 0 allocs/op
⚡ zap | 236 ns/op | 0 B/op | 0 allocs/op
standard library | 453 ns/op | 80 B/op | 2 allocs/op
⚡ zap (sugared) | 337 ns/op | 80 B/op | 2 allocs/op
go-kit | 508 ns/op | 656 B/op | 13 allocs/op
lion | 771 ns/op | 1224 B/op | 10 allocs/op
[...]

PR #356, opened 2021-03-08
Logger factory

kuritka:
- logger factory implementation
 - Move client attribute from depresolver NewDependencyResolver() to ResolveSpec(clien.Client).
   Thanks to that we can resolve configuration when main function starts instead of
   after manager is is instantiated. Now, immediately after obtaining the configuration,
   we can create a logger instance. Just on the beginning of the program.
   Together with moving attribute I had to fix tetsts
 - move resolving configuration on the top of main and initialize logger just after that
 - remove setupLog from main. new logger takes it's his role. the new logger will be used across whole app.
 - Change resolving `LOG_FORMAT`, in case is not given, default value is json instead of simple.

Related to #331

Screenshot 2021-03-08 at 18 12 58
*Logger is configured, third line is from logger which is still present in reconciliation*

Screenshot 2021-03-08 at 18 15 57
*Logger was incorrectly configured*

Screenshot 2021-03-08 at 18 17 35
*Logger was not configured so default logger was returned [json, info]*


k0da (dismissed):

controllers/depresolver/depresolver_config.go
```diff
@@ -76,7 +76,7 @@ func (dr *DependencyResolver) ResolveOperatorConfig() (*Config, error) {
 		dr.config.Override.FakeDNSEnabled = env.GetEnvAsBoolOrFallback(OverrideWithFakeDNSKey, false)
 		dr.config.Override.FakeInfobloxEnabled = env.GetEnvAsBoolOrFallback(OverrideFakeInfobloxKey, false)
 		dr.config.Log.Level, _ = zerolog.ParseLevel(strings.ToLower(env.GetEnvAsStringOrFallback(LogLevelKey, zerolog.InfoLevel.String())))
-		dr.config.Log.Format = parseLogOutputFormat(strings.ToLower(env.GetEnvAsStringOrFallback(LogFormatKey, simple)))
+		dr.config.Log.Format = parseLogOutputFormat(strings.ToLower(env.GetEnvAsStringOrFallback(LogFormatKey, JSON.String())))
```
somaritane, line 79: Just thinking: `JSON`, `depresolver.JSON`, `Simple`,`depresolver.Simple` looks a bit unusual in the code, hard to understand the purpose without knowing the context, probably need to add `Format` post-fix to these "enum"-like constants, e.g. `JSONFormat`, `SimpleFormat` etc., the way how zerolog does it (`zerolog.InfoLevel`)

somaritane (changes requested):
@kuritka looks good, just a small comment

controllers/depresolver/depresolver_config.go
kuritka: @somaritane amended

PR #360, opened 2021-03-10
Update DNSEndpoint CRD

k0da:
Sync DNSEndpoint CRD from upstream where it was updated to
apiextentions.k8s.io/v1

https://github.com/kubernetes-sigs/external-dns/pull/2001

### DIFF
--- a/.kube-linter.yaml
+++ b/.kube-linter.yaml
@@ -1,3 +1,0 @@
-checks:
-  exclude:
-  - "no-extensions-v1beta"
--- a/chart/k8gb/templates/dns-endpoint-crd-manifest.yaml
+++ b/chart/k8gb/templates/dns-endpoint-crd-manifest.yaml
@@ -1,66 +1,93 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+---
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.5.0
+    api-approved.kubernetes.io: "https://github.com/kubernetes-sigs/external-dns/pull/2007"
   creationTimestamp: null
-  labels:
-    api: externaldns
-    kubebuilder.k8s.io: 1.0.0
   name: dnsendpoints.externaldns.k8s.io
 spec:
   group: externaldns.k8s.io
   names:
     kind: DNSEndpoint
+    listKind: DNSEndpointList
     plural: dnsendpoints
+    singular: dnsendpoint
   scope: Namespaced
-  subresources:
-    status: {}
-  validation:
-    openAPIV3Schema:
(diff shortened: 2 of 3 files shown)


.kube-linter.yaml
```diff
@@ -1,3 +1,2 @@
 checks:
   exclude:
-  - "no-extensions-v1beta"
```
somaritane, line 3: do we still need `.kube-linter.yaml` after that?

ytsarev (changes requested):
Please link the upstream PR.

Shouldn't we update external-dns before crd version bump?

somaritane (changes requested):
lgtm, just a small question

.kube-linter.yaml
k0da: I'd like to keep it for later additions

k0da:
@ytsarev this is unrelated to external-dns version. There are no changes to the object itself, what changed is a representation of an object to kubernetes API

.kube-linter.yaml
somaritane: It goes slightly against YAGNI principle; when you don't need functionality anymore - you delete it, makes things simple and uncluttered. We can add `.kube-linter.yaml` later, once we really need it.

ytsarev (changes requested):
see terratest
```
Error: CustomResourceDefinition.apiextensions.k8s.io "dnsendpoints.externaldns.k8s.io" is invalid: metadata.annotations[api-approved.kubernetes.io]: Required value: protected groups must have approval annotation "api-approved.kubernetes.io", see https://github.com/kubernetes/enhancements/pull/1111
```

k0da:
Sent upstream for approval https://github.com/kubernetes-sigs/external-dns/pull/2007

k0da:
Upstream finally got merged, updated PR with new annotated CRD

chart/k8gb/templates/dns-endpoint-crd-manifest.yaml
```diff
@@ -1,66 +1,93 @@
...
+          metadata:
+            type: object
+          spec:
+            description: DNSEndpointSpec defines the desired state of DNSEndpoint
+            properties:
+              endpoints:
+                items:
+                  description: Endpoint is a high-level way of a connection between a service and an IP
+                  properties:
+                    dnsName:
```
kuritka, line 38: is this optional field ?

k0da: @kuritka CRD taken as is from upstream external-dns.  I don't think we can introduce deviation here. Yes, looks like DNSEndpoints aren't validated.

kuritka (approved):

ytsarev (approved):

somaritane (approved):

PR #362, opened 2021-03-10
Switch CoreDNS to LoadBalancer in local setup

k0da:
Simplify setup by spawning local clusters with LoadBalancer.

    * Drop exposeCoreDNS helm value and rely on coredns.serviceType
    * Drop internal coredns svc representation and instruct helm to install
      svc for CoreDNS based on coredns chart values
    * adjust documentation


Makefile
```diff
@@ -359,9 +359,7 @@ endef
 # get-host-alias-ip switch to second context ($2), search for IP and switch back to first context ($1)
 # function returns one IP address
 define get-host-alias-ip
-	kubectl config use-context $2 > /dev/null && \
-	kubectl get nodes $2-agent-0 -o custom-columns='IP:status.addresses[0].address' --no-headers && \
-	kubectl config use-context $1 > /dev/null
+	kubectl --context $2 get nodes $2-agent-0 -o custom-columns='IP:status.addresses[0].address' --no-headers
```
kuritka, line 362: Argument $1 dismissed while $2 is used. Remove second argument and use the first only.

Makefile
```diff
@@ -359,9 +359,7 @@ endef
 # get-host-alias-ip switch to second context ($2), search for IP and switch back to first context ($1)
 # function returns one IP address
```
kuritka, line 360: fix comment

kuritka (changes requested):

Makefile
kuritka: also, fix tests pls.

Makefile
k0da: done

chart/k8gb/templates/operator.yaml
```diff
@@ -111,7 +111,7 @@ spec:
             - name: NS1_ENABLED
               value: "true"
             {{ end }}
-            {{ if .Values.k8gb.exposeCoreDNS }}
+            {{ if eq "LoadBalancer" .Values.coredns.serviceType }}
             - name: COREDNS_EXPOSED
```
somaritane, line 115: @ytsarev, @k0da, is `COREDNS_EXPOSED` used for tests only? 
Name looks semantically confusing, as we always want associated CoreDNS to be exposed, regardless of serviceType.
Maybe it makes sense for tests to check if associated CoreDNS Service primitive is successfully created with "LoadBalancer" serviceType, and not created by default? In this case we don't have to wire this technical env variable.
Or at least it can be renamed to `COREDNS_EXTERNAL_LB` or similar.

k0da: @somaritane no.
k8gb controller based on this setting:
if true:
- will check CoreDNS service for LoadBalancer IP
if false:
- will check GSLB object for ingress IPs (node ips)

k0da: https://github.com/AbsaOSS/k8gb/blob/master/controllers/providers/dns/external.go#L64-L68

somaritane: @k0da I see, thanks, then I'd consider renaming it to `COREDNS_EXTERNAL_LB`.
Or, it can be renamed to `COREDNS_SERVICE_TYPE` and accept standard serviceType values ('LoadBalancer', 'ClusterIP'...)

k0da: IMO we should either get rid of it or keep it boolean

k0da: I agree,setting is confusing

docs/exposing_dns.md
```diff
@@ -38,20 +38,22 @@ kubectl patch -n ingress-nginx -p '{"data":{"53":"k8gb/k8gb-coredns:53"}}' --typ
...
 
-k8gb:
+coredns:
   ...
-  exposeCoreDNS: true # <== expose UDP DNS traffic via external load balancer
+  serviceType: LoadBalancer # <== expose CoreDNS service via external load balancer
+  service:
+    annotations:
+      # tell AWS to use NLB load balancer
+      service.beta.kubernetes.io/aws-load-balancer-type: nlb
```
somaritane, line 56: @k0da, am I getting this correctly that annotations are going to be sorted out automatically by upstream CoreDNS helm chart? (https://github.com/coredns/helm/blob/master/stable/coredns/values.yaml#L66).
Looks nice.

chart/k8gb/templates/operator.yaml
somaritane: > IMO we should either get rid of it or keep it boolean

Question is, which settings we want controller logic to rely on: settings from helm release vs settings from real CoreDNS Service primitive spec.

chart/k8gb/values.yaml
```diff
@@ -42,6 +42,7 @@ coredns:
   servers:
   - zones:
     - zone: .
+      scheme: dns://
```
ytsarev, line 45: What is that setting?

docs/exposing_dns.md
```diff
@@ -61,8 +63,8 @@ apiVersion: v1
 kind: Service
 metadata:
   annotations:
-    service.beta.kubernetes.io/aws-load-balancer-type: nlb # <== tell AWS to use NLB load balancer
-  name: k8gb-coredns-lb
+    service.beta.kubernetes.io/aws-load-balancer-type: nlb
```
ytsarev, line 66: What was wrong with NLB clarification comment?

deploy/ingress/nginx-ingress-values.yaml
```diff
@@ -2,18 +2,6 @@
...
-
-  ## Use host ports 80 and 443
-  daemonset:
-    useHostPort: false
-
-    hostPorts:
-      http: 80
-      https: 443
-
-  kind: DaemonSet
```
ytsarev, line 15: I'm skeptical of totally ditching the local setup which is closely replicating our on-prem scenario

ytsarev (commented):
Agreed on configuration point optimization.

Not so sure of the need of switching local setup to loadbalancer - what are the actual benefits?

In any case the following things should be end-to-end tested before the merge:

* Local setup+local demo
* AWS reference setup

somaritane:
> Not so sure of the need of switching local setup to loadbalancer - what are the actual benefits?

@ytsarev I see benefits in replicating scenario with external load balancer, which is equally valid in our case (cloud/AWS).
However, I'd welcome configurable possibility for contributors to decide which scenario to chose for local setup. Not sure why we can't support both.

> In any case the following things should be end-to-end tested before the merge:
> 
> * Local setup+local demo
> * AWS reference setup

👍

chart/k8gb/values.yaml
k0da: this instructs core dns chart to listen on UDP only

docs/exposing_dns.md
k0da: Yes

docs/exposing_dns.md
k0da: This is an example of resulted service. Setting and comment moved above to chart setting

deploy/ingress/nginx-ingress-values.yaml
k0da: The whole point is to simplify quick debug test setup and keep it lightweight. But we do test collecting CoreDNS IPs from both service and ingress. Besides that exposure of a CoreDNS service should be irrelevant to controller configuration.

ytsarev: Should be, but practically it's not for on-prem. We are not going to cover the most critical to us code path at the moment in terratests. LB version of coredns exposure is definitely more valuable for the community. 
We might look into an attempt to automatically test both scenarious as @somaritane proposed.

k0da:
- Switchet to k3d config, so setup is not duplicated across Makefile and GH action. 
- Split test-apps deploy out of create-local-cluster.
- Run two scenarios in parralel in GH action
- Default setup is LoadBalancer

kuritka (dismissed):

ytsarev (changes requested):
```
$ make deploy-full-local-setup
FATA[0000] unknown shorthand flag: 'c' in -c
make: *** [deploy-full-local-setup] Error 1
```

Please document the requirement of specific `k3d` version

ytsarev:
@k0da `make deploy-full-local-setup` was broken, I pushed fix https://github.com/AbsaOSS/k8gb/pull/362/commits/f3a72078001cd9227b21e47224f498e0f75c776f to this branch

ytsarev (changes requested):
```
Error: UPGRADE FAILED: cannot patch "k8gb-coredns" with kind Service: Service "k8gb-coredns" is invalid: spec.ports: Required value
```
Permutation test failed. Upgrade will be broken.

ytsarev (changes requested):
```
k -n k8gb get svc
NAME           TYPE           CLUSTER-IP      EXTERNAL-IP   PORT(S)        AGE
k8gb-coredns   LoadBalancer   172.20.54.112   <pending>     53:30089/UDP   6h59m
```
Another thing - community route53 samples are broken because nlb annotation is not propagated there.

.github/workflows/terratest.yaml
```diff
@@ -19,30 +19,45 @@ jobs:
...
           cluster-name: "test-gslb1"
           args: >-
-            -p "80:80@agent[0]"
-            -p "443:443@agent[0]"
-            -p "5053:53/udp@agent[0]"
-            --agents 1
-            --no-lb
-            --network k3d-action-bridge-network
-            --k3s-server-arg "--no-deploy=traefik,servicelb,metrics-server"
+            --config=k3d/lb/test-gslb1.yaml
```
kuritka, line 22:
```suggestion
  args: --config=k3d/lb/test-gslb1.yaml
```

docs/examples/ns1/k8gb-cluster-ns1-eu-west-1.yaml
```diff
@@ -4,7 +4,11 @@ k8gb:
   edgeDNSServer: "169.254.169.253" # to handle splitbrain situation with TXT timestamp
   clusterGeoTag: "eu-west-1" # used for places where we need to distinguish between differnet Gslb instances
   extGslbClustersGeoTags: "us-east-1" # comma-separated list of external gslb geo tags to pair with
-  exposeCoreDNS: true # Create Service type LoadBalancer to expose CoreDNS
+
+coredns:
+  service:
+    annotations:
+      service.beta.kubernetes.io/aws-load-balancer-type: nlb
```
ytsarev, line 11: Can we move the coredns service configuration to separate PR maybe?

ytsarev (changes requested):

docs/examples/ns1/k8gb-cluster-ns1-eu-west-1.yaml
k0da: @ytsarev you mean introduce scenario based on embedded service?

ytsarev: I mean configuration change option change

Makefile
```diff
@@ -407,11 +419,9 @@ define hit-testapp-host
 endef
 
 define init-test-strategy
- 	kubectl config use-context k3d-test-gslb2
- 	kubectl apply -f $1
- 	kubectl config use-context k3d-test-gslb1
- 	kubectl apply -f $1
- 	$(call testapp-set-replicas,2)
+	kubectl --context k3d-test-gslb2 apply -f $1
+	kubectl --context k3d-test-gslb1 apply -f $1
```
kuritka, line 423: reusing constant
```suggestion
kubectl --context k3d-$(CLUSTER_GSLB2) apply -f $1
kubectl --context k3d-$(CLUSTER_GSLB1) apply -f $1
```

kuritka (changes requested):

PR #367, opened 2021-03-15
Integration zerolog

kuritka:
Related to #331
In order to integrate a zerolog, I had to create log package and take a few steps.

### Log package
log package contains three components: **factory, singleton and adapter**.

#### loger factory, singleton
Instead of injecting `*zerolog.Logger`  into each object (providers, assistant, etc). I created logger [singleton](https://en.wikipedia.org/wiki/Singleton_pattern).
This component is accessible from anywhere and contains one and only one instance of logger, which is accessible through reader function `Logger()`
Singleton must be initialised after resolving configuration and internally calls factory, which builds logger instance.
That's why logger factory is completely private now.

To inject logger into package do following:

```go
import "github.com/AbsaOSS/k8gb/controllers/log"

// injection
var logger = log.Logger()

func MyFunc(){
 logger.Info().Msg("hello from logger")
}
```

#### Adapter
ControllerManager still calls logr internally. It can't use anything else. Unfortunately, the problem is that the logger inside ControllerManager completely ignores the settings of our logger, so plain text would appear between json records, etc.
It also ignores our formatting. The solution provides [**Adapter** pattern](https://en.wikipedia.org/wiki/Adapter_pattern)
which basically allows us to encapsulate zerolog instance into logr interface. So on the outside it looks like logr but
internally it produces records by our zerolog instance. That's why logger within ControllerManager will print the same, expected way as we do in the rest of the application.  Here is how we inject adapter:
```go
// main.go
ctrl.SetLogger(log.NewAdapter(logger))
// controller manager has logger instance injected, and that's all. We don't need to do anything else.
```

### logging messages
- I had to remove rlog and zapr from all components
- passing logger between components by arguments or structure fields was removed, now injection via singleton
- The way how we use zerolog:
```go
logger.Info().Msg("hello from zerolog")
logger.Err(err).Msg("error from zerolog")  // prints "error from zerolog" and error
logger.Info().Msgf("hello from zerolog %s", stringvariable)
logger.Err(err).Msgf("error from zerolog %s", stringvariable)
```
- I made many replacements like this:
```shell
log.Info(fmt.Sprintf("%s annotation is missing, skipping Gslb creation...", primaryGeoTagAnnotation))
# I replaced with
logger.Info().Msgf("%s annotation is missing, skipping Gslb creation...", primaryGeoTagAnnotation)
```
- We had places where we printed Info messages within errors. I replaced them by Warnings.
```go
IPs, err := utils.Dig(r.edgeDNSServer, lbHostname)
	if err != nil {
		logger.Warn().Msgf("Can't dig k8gb-coredns-lb service loadbalancer fqdn %s (%s)", lbHostname, err)
		return nil, err
	}
```

### Output

Screenshot 2021-03-12 at 11 38 35
*Controller manager producing log with zapr*

Screenshot 2021-03-12 at 11 39 09
*Controller manager producing logs with Adapter in json*
Screenshot 2021-03-12 at 11 40 03
*Controller manager producing logs with Adapter in simple*

### TODO

- identify places for debug messages
- We don't need to print everything on info level, some info messages would go to debug
- We don't want to generate errors like this:
```shell
github.com/go-logr/zapr.(*zapLogger).Error
        /Users/ab011th/go/pkg/mod/github.com/go-logr/zapr@v0.4.0/zapr.go:132
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
        /Users/ab011th/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.4/pkg/internal/controller/controller.go:246
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
        /Users/ab011th/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.4/pkg/internal/controller/controller.go:218
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker
        /Users/ab011th/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.4/pkg/internal/controller/controller.go:197
k8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1
        /Users/ab011th/go/pkg/mod/k8s.io/apimachinery@v0.20.4/pkg/util/wait/wait.go:155
k8s.io/apimachinery/pkg/util/wait.BackoffUntil
        /Users/ab011th/go/pkg/mod/k8s.io/apimachinery@v0.20.4/pkg/util/wait/wait.go:156
k8s.io/apimachinery/pkg/util/wait.JitterUntil
        /Users/ab011th/go/pkg/mod/k8s.io/apimachinery@v0.20.4/pkg/util/wait/wait.go:133
k8s.io/apimachinery/pkg/util/wait.Until
        /Users/ab011th/go/pkg/mod/k8s.io/apimachinery@v0.20.4/pkg/util/wait/wait.go:90
```
Instead, we want to have simplified view in JSON. I'll need to rewrite all error producing into
`github.com/pkg/errors`.


main.go
```diff
@@ -59,21 +58,23 @@ func main() {
 
 	resolver := depresolver.NewDependencyResolver()
 	config, err := resolver.ResolveOperatorConfig()
-	// LoggerFactory creates logger ALWAYS - no matter what isn't resolved
-	logger := controllers.NewLogger(config).Get()
+	// Init creates logger ALWAYS - no matter what isn't resolved
```
ytsarev, line 61: Could you rephrase this comment? not really clear

main.go
```diff
@@ -59,21 +58,23 @@ func main() {
...
-	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
+	ctrl.SetLogger(log.NewAdapter(logger))
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:             runtimescheme,
 		MetricsBindAddress: metricsAddr,
 		Port:               9443,
 		LeaderElection:     enableLeaderElection,
 		LeaderElectionID:   "8020e9ff.absa.oss",
+		//Logger: log.NewAdapter(logger),
```
ytsarev, line 77: Is it leftover?

controllers/providers/assistant/gslb.go
```diff
@@ -231,7 +229,7 @@ func (r *GslbLoggerAssistant) GetExternalTargets(host string, fakeDNSEnabled boo
 
 		a, err := dns.Exchange(g, ns)
 		if err != nil {
-			r.Info("Error contacting external Gslb cluster(%s) : (%v)", cluster, err)
+			logger.Warn().Msgf("Contacting external Gslb cluster(%s) : (%v)", cluster, err)
```
ytsarev, line 232:
```suggestion
			logger.Warn().Msgf("Trying to contact external Gslb cluster(%s) : (%v)", cluster, err)
```

ytsarev (changes requested):
Some minor changes/questions.

Otherwise, it is a really great change 👍 

@somaritane @k0da PR is huge, another pair of reviewing eyes will not hurt here :)

main.go
kuritka: Hi @ytsarev,  amended. 

Regarding this concrete comment: 
We are creating logger from configuration, but if resolving configuration fails, from what we create logger to print error? The solution is to initialize default logger in case of configuration failed. 

I change comment to:
_Initialize desired logger or default logger in case of configuration failed_

controllers/log/logr.go
```diff
@@ -0,0 +1,103 @@
...
+	"encoding/json"
+	"fmt"
+
+	"github.com/go-logr/logr"
+	"github.com/rs/zerolog"
+)
+
+// Adapter implements logr.Logger interface.
+// The adapter allows us to encapsulate zerolog into go-logr/logr interface.
+type Adapter struct {
```
somaritane, line 30: Conceptual question: as we're implementing go-logr/logr interface here anyway (like https://github.com/go-logr/zapr does), would it make sense to continue using logr log semantics in the rest of the code instead of replacing it with zerolog-specific chained syntax (log.Info() -> logger.Info().Msg())

somaritane (changes requested):

controllers/providers/assistant/gslb.go
```diff
@@ -24,36 +24,34 @@ import (
...
+// It uses apimachinery client to call kubernetes API
 type GslbLoggerAssistant struct {
-	log           logr.Logger
 	client        client.Client
 	k8gbNamespace string
 	edgeDNSServer string
 }
 
-func NewGslbAssistant(client client.Client, log logr.Logger, k8gbNamespace, edgeDNSServer string) *GslbLoggerAssistant {
+var logger = log.Logger()
```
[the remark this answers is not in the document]
kuritka: amended for all loggers

controllers/log/logr.go
kuritka: @somaritane , good question, and appreciate review. 

First, controllerManager dosn't produce much logs. More or less it is during application start. The primary reason for adapter is to let ControllerManager produce logs same or similar style as the rest of the application. i.e. We don't want to controllerManager print simple message when `LOG_FORMAT=json`

 I'd like to use chained semantics, but zerolog and logr have not compatible syntax. 
logr has `WithValues(slice []interface{})` which adds some key-value pairs of context to a logger (it's slice where, odd keys, even values). Also `Info(msg string, keysAndValues ...interface{})`,  has key/value pairs which can then be used to add additional message information. So conversion is not 1:1. To Zerolog requires some additional logic to print those keyValues slices.

Technically, we can split into standalone `zerologr` package, but would be nice to do it when we will reuse package, do some init() magic, internalize etc.. I suggest to keep it as it is now and split when necessary.

somaritane: @kuritka , thank you for explanation.

somaritane (approved):
lgtm!

ytsarev (approved):
Great stuff 👍

PR #369, opened 2021-03-16
Use simple log format as default for make run

somaritane:
This PR sets log format to simple for local make run target.


Makefile
```diff
@@ -1,17 +1,17 @@
 # Copyright 2021 Absa Group Limited
-# 
+#
```
kuritka, line 2: The PR looks LGTM for me, but we must do any license changes in golic/config.yaml. 

The problem is, when license string differs from template, it injects new license. Here, we will have two licenses in makefile as a result. 

Could you please, in the scope of this PR, skip license change?

kuritka (changes requested):

Makefile
somaritane: @kuritka hmm, might be my editor acting and auto-removing extra spaces, and diff ignores this change. will surely amend

somaritane: @kuritka amended

kuritka (approved):
👍

PR #371, opened 2021-03-17
Drop linter from terratest action

k0da:
It is not really used during that step


kuritka (approved):
Linter was forgotten here and  primarily we use redundant linter action in different workflow. 

Nice finding @k0da !

PR #372, opened 2021-03-17
Golic CI

kuritka:
- Makefile: remove license from test target
- CI, install and run golic v0.4.4
- license target can modify files, while build step is `--dry -x`
 
*picture below shows what happens if some file doesn't have expected license*
Screenshot 2021-03-17 at 18 16 20

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -215,7 +215,7 @@ ifeq (, $(shell which golic))
 	GOLIC_TMP_DIR=$$(mktemp -d) ;\
 	cd $$GOLIC_TMP_DIR ;\
 	go mod init tmp ;\
-	go get github.com/AbsaOSS/golic@v0.3.1 ;\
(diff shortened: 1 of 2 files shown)


somaritane (approved):

k0da (approved):

Makefile
```diff
@@ -292,7 +292,7 @@ start-test-app:
 
 # run tests
 .PHONY: test
-test: license lint
```
ytsarev, line 295: @kuritka why did we remove local invocation?

kuritka: I supposed you didn't want to check license during a test. Should I amend back ?

ytsarev: @kuritka yep, I meant to amend it with CI, not replace. So CI would catch the missing license in case the PR was created without running local tests

kuritka: @ytsarev exactly

PR #378, opened 2021-03-18
Fix changelog in Release

k0da:
Switch to proved to work solution

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -14,14 +14,16 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - name: Get previous tag
-        uses: WyriHaximus/github-action-get-previous-tag@v1.0.1
-        id: previoustag
+      - name: Get tag
+        id: get_tag
+        run: |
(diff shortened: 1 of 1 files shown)


kuritka (approved):

.github/workflows/release.yaml
```diff
@@ -14,14 +14,18 @@ jobs:
...
-      - name: Get previous tag
-        uses: WyriHaximus/github-action-get-previous-tag@v1.0.1
-        id: previoustag
+      - name: Get tag
+        id: get_tag
+        run: |
+          previous_tag=$(git tag --sort=v:refname | tail -2 | head -1)
+          current_tag=$(git tag --sort=v:refname | tail -1)
+          echo "::set-output name=previous_tag::${previous_tag}"
+          echo "::set-output name=current_tag::${current_tag}"
```
ytsarev, line 23: I do not see the usage of `current_tag` here. So we can remove it and leave only `previous_tag` around

ytsarev (changes requested):

.github/workflows/release.yaml
k0da: Dropped

ytsarev (approved):

PR #384, opened 2021-03-19
Workaround goreleaser not pushing docker images.

k0da:
goreleaser on release draft doesn't push docker images to the registry,
hence docker manifest build fails. Build and push images in advance
until https://github.com/goreleaser/goreleaser/pull/2119 is merged

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -34,6 +34,13 @@ jobs:
         with:
           username: ${{ secrets.DOCKER_USER }}
           password: ${{ secrets.DOCKER_PASSWORD }}
+      # Hack around until https://github.com/goreleaser/goreleaser/pull/2119 is fixed
+      - name: Operator image build and push
+        run: |
+          export PATH=~/go/bin:$PATH
+          make docker-build-multi
+          make docker-push-multi
+          make docker-manifest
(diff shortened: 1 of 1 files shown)


ytsarev (dismissed):

kuritka (dismissed):

Makefile
```diff
@@ -170,7 +170,7 @@ dns-smoke-test:
 # build docker images for multiple architectures
 # useful for CI
 .PHONY: docker-build-multi
-docker-build-multi: test
+docker-build-multi:
```
ytsarev, line 173: locally it is properly guarding the docker build by running the tests.

Why the tree is getting dirty in the first place?

ytsarev (changes requested):

Makefile
k0da: generated crd.

ytsarev: so this generation-based deviation is a problem on multiple levels.

Still not a reason to disable the tests

k0da: docker-build-multi is used in CI only as it makes no sense to build arm64 locally as we discussed while implementing amr64 support. CI runs tests in other jobs, if not a generator it would simply just extra time. Eitherway, will implement docker build by using docker-buildx action. Hold on ;)

PR #385, opened 2021-03-19
bump golic version

kuritka:
The problem is , golic uses online master configuration, which is shared between all golic versions. Change in v0.4.7 affects v0.4.4

The problem will be resolved when master config will be attached to golic binary as an embedded resource .

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,7 +19,7 @@ jobs:
           version: v1.32
       - name: golic
         run: |
-          go install github.com/AbsaOSS/golic@v0.4.4
+          go install github.com/AbsaOSS/golic@v0.4.7
           golic inject -c "2021 Absa Group Limited" --dry -x
       - name: go test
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -19,7 +19,7 @@ jobs:
           version: v1.32
       - name: golic
         run: |
-          go install github.com/AbsaOSS/golic@v0.4.4
+          go install github.com/AbsaOSS/golic@v0.4.7
```
somaritane, line 22: @kuritka it's already covered by https://github.com/AbsaOSS/k8gb/pull/375

somaritane:
@kuritka it's already covered by https://github.com/AbsaOSS/k8gb/pull/375

ytsarev (approved):

ytsarev:
We need to unblock rest of the pipelines, please merge it in, #375 will have to be rebased

PR #387, opened 2021-03-19
use gopkg.strings.Format() instead of local utils.ToString()

kuritka:
There are two goals for upgrade
 - replace utils by  reusable utils 
 - I'd like to use utils.Format in main.go in context of log.Debug().Msgf(...)
However utils lives in `internal` which is not enough today 
(all exists in gopkg now, we can slowly get rid of it).


k0da (approved):

controllers/gslb_controller_test.go
```diff
@@ -339,8 +340,8 @@ func TestGslbCreatesDNSEndpointCRForHealthyIngressHosts(t *testing.T) {
 	err = settings.client.Get(context.TODO(), settings.request.NamespacedName, dnsEndpoint)
 	require.NoError(t, err, "Failed to load DNS endpoint")
 	got := dnsEndpoint.Spec.Endpoints
-	prettyGot := utils.ToString(got)
-	prettyWant := utils.ToString(want)
+	prettyGot := str.Format(got)
```
somaritane, line 343: Please consider renaming `Format()` function back to `ToString()` in order to resolve ambiguity.
In many languages and std libraries, format functions are universal, and provide string formatting according to provided format expression.
In this current case we are dealing with simplified conversion function without parameters, plus having its internal logic.
In this case old `ToString()` name perfectly serves the purpose.

somaritane (changes requested):
@kuritka left suggestion in review comments

controllers/gslb_controller_test.go
kuritka: @somaritane , please see: [gopkg #7 ](https://github.com/AbsaOSS/gopkg/pull/7)

kuritka:
@somaritane  amended

somaritane (approved):
@kuritka, lgtm!

PR #391, opened 2021-03-20
Update nginx ingress chart

k0da:
* https://charts.helm.sh/stable/ is outdated
* Switch to https://github.com/kubernetes/ingress-nginx provided chart

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -356,10 +356,10 @@ define deploy-local-cluster
 	$(call deploy-k8gb-with-helm,$1,$2,$3,$4)
 
 	@echo "\n$(YELLOW)Deploy Ingress $(NC)"
-	helm repo add --force-update stable https://charts.helm.sh/stable
+	helm repo add --force-update nginx-stable https://kubernetes.github.io/ingress-nginx
 	helm repo update
-	helm -n k8gb upgrade -i nginx-ingress stable/nginx-ingress \
-		--version 1.41.1 -f deploy/ingress/nginx-ingress-values.yaml
+	helm -n k8gb upgrade -i nginx-ingress nginx-stable/ingress-nginx \
+		--version 3.24.0 -f deploy/ingress/nginx-ingress-values.yaml
 
 	@echo "\n$(YELLOW)Deploy GSLB cr $(NC)"
 	kubectl apply -f deploy/crds/test-namespace.yaml
@@ -433,7 +433,7 @@ endef
 
 # waits for NGINX, GSLB are ready
 define wait-for-ingress
-	kubectl -n k8gb wait --for=condition=Ready pod -l app=nginx-ingress --timeout=600s
+	kubectl -n k8gb wait --for=condition=Ready pod -l app.kubernetes.io/name=ingress-nginx --timeout=600s
 endef
 
 define generate
--- a/deploy/ingress/nginx-ingress-values.yaml
+++ b/deploy/ingress/nginx-ingress-values.yaml
@@ -1,7 +1,11 @@
 ## nginx configuration
-## Ref: https://github.com/kubernetes/ingress/blob/master/controllers/nginx/configuration.md
+## Ref: https://github.com/kubernetes/ingress-nginx/tree/master/charts/ingress-nginx
 ##
 controller:
+  admissionWebhooks:
+    enabled: false
+    patch:
+     enabled: false
   hostNetwork: true
 
   ## Use host ports 80 and 443


k0da:
Fixes https://github.com/AbsaOSS/k8gb/issues/388, note this is not a chart maintained by Nginx, but rather official for ingress controller on k8s side. Advantage over official is, it supports setting udp-services-configmap which is crucial in our setup.

kuritka (approved):

kuritka (changes requested):
fix terratest pls..

k0da:
@kuritka for terratest, it fails randomly

kuritka:
@k0da, must be fixed otherwise. When we merge, master branch will fail randomly 😬 .

k0da:
@kuritka sure, I'm working on it.

k0da:
mergable

Makefile
```diff
@@ -326,10 +326,10 @@ define deploy-local-cluster
 		--set k8gb.log.level=$(LOG_LEVEL)
 
 	@echo "\n$(YELLOW)Deploy Ingress $(NC)"
-	helm repo add --force-update stable https://charts.helm.sh/stable
+	helm repo add nginx-stable https://kubernetes.github.io/ingress-nginx
```
ytsarev, line 329: `--force-update` was added to not fail locally when the repo is already present. Why was it removed?

deploy/ingress/nginx-ingress-values.yaml
```diff
@@ -2,6 +2,10 @@
 ## Ref: https://github.com/kubernetes/ingress/blob/master/controllers/nginx/configuration.md
```
ytsarev, line 2: Looks like ref is outdated, can we update it?

deploy/ingress/nginx-ingress-values.yaml
```diff
@@ -2,6 +2,10 @@
...
 ##
 controller:
+  admissionWebhooks:
+    enabled: false
+    patch:
+     enabled: false
```
ytsarev, line 8: Would be great to understand what this options do. Update of the ref about would help it I guess. Please fix the reference with the link to up-to-date documentation you used for this change

ytsarev (changes requested):

Makefile
k0da:
```shell
 > helm repo add nginx-stable https://kubernetes.github.io/ingress-nginx
"nginx-stable" already exists with the same configuration, skipping```

Comment (k0da) [reply Makefile:329]: Option is deprecated and no longer needed

Comment (ytsarev) [reply Makefile:329]: @k0da since which helm version?

File: deploy/ingress/nginx-ingress-values.yaml (hunk 3 above)
Comment (k0da) [reply deploy/ingress/nginx-ingress-values.yaml:8]: @ytsarev this is something new chart adds, to prevent new features piled up, I just disable them to keep things as they were. In a nutshell, this prevents validation webhook being installed: https://github.com/kubernetes/ingress-nginx/blob/master/charts/ingress-nginx/templates/admission-webhooks/validating-webhook.yaml so far it supports networking.k8s.io only

File: Makefile (hunk 1 above)
Comment (ytsarev) [reply Makefile]:
```
$ helm repo add nginx-stable https://kubernetes.github.io/ingress-nginx
Error: repository name (nginx-stable) already exists, please specify a different name
$ echo $?
1
```

Comment (k0da) [reply Makefile]: @ytsarev I just reverted it, since option is still there.

File: deploy/ingress/nginx-ingress-values.yaml [hunk 4]
```diff
@@ -1,7 +1,11 @@
 ## nginx configuration
-## Ref: https://github.com/kubernetes/ingress/blob/master/controllers/nginx/configuration.md
+## Ref: https://kubernetes.github.io/ingress-nginx/deploy
```
ytsarev, line 2: https://github.com/kubernetes/ingress-nginx/tree/master/charts/ingress-nginx looks more relevant

ytsarev, line 2:
```suggestion
## Ref: https://github.com/kubernetes/ingress-nginx/tree/master/charts/ingress-nginx
```

ytsarev (changes requested):

deploy/ingress/nginx-ingress-values.yaml
k0da: 👍

deploy/ingress/nginx-ingress-values.yaml
kuritka: > so far it supports networking.k8s.io only

@k0da, does it mean we should merge it after #417 ?

k0da: I won't pull it into this PR. We can enable it later. So IMO this is not a blocker

k0da:
Any other comments?

ytsarev:
@k0da please rebase with master

k0da:
@ytsarev done

ytsarev (approved):
@k0da thanks

k0da:
@kuritka pls rereview

kuritka (approved):

Issue #401, opened 2021-03-22
automate ingress validation annotation


k0da:
Currently ingress object taken as is and validation is inherited from upstream. We rely on `spec.rules.http` not to be nil.

We need to find a way to instruct controller-gen to add extra markers so `spec.rules.http` is validated

PR #402, opened 2021-03-22
HTTP ingress rule value is Mandatory

kuritka:
related to #401 

- the custom structures with mandatory field overrides `IngressSpec` and
and implements necessary routines like `DeepCopyInto` and transformations between
`k8gbv1beta1` and `v1beta1`

- controller-gen created `DeepCopy()` functions in `zz_generated.deepcopy.go`


terratest/test/k8gb_ingress_mandatory_http_test.go
```diff
@@ -0,0 +1,65 @@
+/*
```
k0da, line 1: We already have it covered in terratest/test/k8gb_basic_app_test.go

kuritka: @k0da , I see, amending....

kuritka:
@k0da amended

ytsarev (changes requested):
From https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/

```
The v1.22 release will stop serving the following deprecated API versions in favor of newer and more stable API versions:

Ingress in the extensions/v1beta1 API version will no longer be served
Migrate to use the networking.k8s.io/v1beta1 API version, available since v1.14. Existing persisted data can be retrieved/updated via the new version.
```

So we should stick with `networking.k8s.io/v1beta1 `

kuritka:
@ytsarev, @k0da  - so should we upgrade to `networking.k8s.io/v1beta1 ` first and then migrate this PR to `networking.k8s.io/v1beta1 `??

k0da:
@kuritka IMO we can do both

kuritka:
@ytsarev @k0da @somaritane , I succesfully migrated to `networking.k8s.io/v1beta1` (see: [this](https://github.com/AbsaOSS/k8gb/compare/ingress-mandatory-http-networking-v1beta1)), but to make PR slim I rather split large one into into two smaller PR's. 

This one is making `HTTPIngressRuleValue` mandatory. The migration would go to the second PR.

just in case: 
To migrate into `networking.k8s.io/v1beta1`, I had to [change CRD](https://github.com/AbsaOSS/k8gb/compare/ingress-mandatory-http-networking-v1beta1#diff-b89a907c86922705f2e446b83056186485868f3152f770c9b36e5ed675fecd83R221) and `roles.yaml` (would be second PR).

k0da (dismissed):
👍

api/v1beta1/types.go
```diff
@@ -0,0 +1,194 @@
...
+	if in.Rules != nil {
+		in, out := &in.Rules, &out.Rules
+		*out = make([]IngressRule, len(*in))
+		for i := range *in {
+			(*in)[i].DeepCopyInto(&(*out)[i])
+		}
+	}
+}
+
+// DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
```
somaritane, line 146: do we autogenerate this function here too, like in zz_generated.deepcopy.go?

kuritka: These are not auto-generated, amending comments.

kuritka: @somaritane comments amended

api/v1beta1/types.go
```diff
@@ -0,0 +1,196 @@
+/*
```
ytsarev, line 1: What is the filename semantics? `types` vs `gslb_types`?

api/v1beta1/types.go
```diff
@@ -0,0 +1,196 @@
...
+
+Generated by GoLic, for more details see: https://github.com/AbsaOSS/golic
+*/
+package v1beta1
+
+import (
+	"k8s.io/api/extensions/v1beta1"
+)
+
+type IngressSpec struct {
```
ytsarev, line 24: Can we reference the upstream source of `IngressSpec` for future?

ytsarev (changes requested):
some cosmetic comments with the focus on future maintenance

api/v1beta1/types.go
kuritka: @ytsarev  good question, 
It lives originally in `k8s.io/api/extensions/v1beta1/types.go`.

api/v1beta1/types.go
kuritka: @ytsarev , good idea, amended

api/v1beta1/types.go
ytsarev: `upstream_types.go` maybe?

api/v1beta1/types.go
```diff
@@ -0,0 +1,198 @@
...
+Generated by GoLic, for more details see: https://github.com/AbsaOSS/golic
+*/
+package v1beta1
+
+import (
+	"k8s.io/api/extensions/v1beta1"
+)
+
+// IngressSpec overrides https://github.com/kubernetes/api/blob/master/extensions/v1beta1/types.go
+// IngressSpec describes the Ingress the user wishes to exist.
```
somaritane, line 25: This comment probably needs rephrasing.

ytsarev, line 25:
```suggestion
// IngressSpec is upstream Ingress to be included in Gslb specification and extended with Gslb specific requirements.
```

ytsarev (changes requested):
IngressSpec Wording suggestion.

api/v1beta1/types.go
kuritka: Renamed, was just following original convention `upstream_types.go` is well descriptive.

ytsarev (approved):

somaritane (approved):
lgtm

PR #403, opened 2021-03-22
Split changelog PR off helm publish workflow

k0da:
Split publish jobs into two independed worflows

### DIFF
--- a/.github/workflows/changelog_pr.yaml
+++ b/.github/workflows/changelog_pr.yaml
@@ -1,0 +1,39 @@
+name: Offline Changelog
+
+on:
+  release:
+    types:
+      - published
+
+jobs:
+  chanagelog-pr:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/changelog_pr.yaml
```diff
@@ -0,0 +1,39 @@
+name: Open PR with offline changelog
```
somaritane, line 1: Small suggestion:
```suggestion
name: Offline Changelog
```

somaritane (changes requested):
@k0da lgtm, just a small title suggestion

.github/workflows/changelog_pr.yaml
k0da: done

somaritane (approved):
lgtm!

ytsarev:
@k0da @somaritane there is still a chance of race condition/deviation for offline CHANGELOG.md. https://docs.github.com/en/actions/guides/storing-workflow-data-as-artifacts might be a way to enhace it

PR #405, opened 2021-03-22
Improve initial logging experience

somaritane:
- Wire logging settings via helm chart values
- Set `simple` log format as default preference
- Use `NO_COLOR` env variable for color settings as per https://no-color.org/
- Disable `NO_COLOR` by default, but enforce for helm deployments
- Use `RFC3339` as datetime format for all loggers
- Use original common field names

This PR contributes to work made in the scope of  #331

### DIFF
--- a/controllers/depresolver/depresolver_test.go
+++ b/controllers/depresolver/depresolver_test.go
@@ -62,7 +62,7 @@ var predefinedConfig = Config{
 		false,
 	},
 	Log: Log{
-		Format: JSONFormat,
+		Format: SimpleFormat,
 	},
 }
 
@@ -171,8 +171,8 @@ func TestResolveConfigWithoutEnvVarsSet(t *testing.T) {
 	defaultConfig.EdgeDNSType = DNSTypeNoEdgeDNS
 	defaultConfig.ExtClustersGeoTags = []string{}
 	defaultConfig.Log.Level = zerolog.InfoLevel
-	defaultConfig.Log.Format = JSONFormat
-	defaultConfig.Log.NoColor = true
+	defaultConfig.Log.Format = SimpleFormat
+	defaultConfig.Log.NoColor = false
 	resolver := NewDependencyResolver()
 	// act
 	config, err := resolver.ResolveOperatorConfig()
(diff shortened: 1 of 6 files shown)


controllers/logging/log.go
```diff
@@ -42,41 +42,30 @@ func newLogger(config *depresolver.Config) *loggerFactory {
...
+		l.log.Format = depresolver.SimpleFormat
 	}
 	if l.log.Level == zerolog.NoLevel {
 		l.log.Level = zerolog.InfoLevel
 	}
 	// We can retrieve stack in case of pkg/errors
 	zerolog.ErrorStackMarshaler = pkgerrors.MarshalStack
 	zerolog.SetGlobalLevel(l.log.Level)
-	if l.log.Level <= zerolog.DebugLevel {
-		dt = "15:04:05"
```
kuritka, line 56: Do we want `RFC822Z` for Debug?   
```
09:06:19 DBG controllers/logging/log.go:80 > Logger settings: [simple, debug]
or
02 Jan 06 15:04 -0700 DBG controllers/logging/log.go:80 > Logger settings: [simple, debug]
```
Debugging is mostly manual, for better readability I wanted to put shorter time format. @ytsarev , @k0da  Thoughts ?

kuritka (changes requested):
Good Job @somaritane 
I have only one uncertainty

controllers/logging/log.go
somaritane: actually it will be different
```
09:06:19 DBG controllers/logging/log.go:80 > Logger settings: [simple, debug]
vs
2021-03-23T15:04:05Z07:00 DBG controllers/logging/log.go:80 > Logger settings: [simple, debug]
```
as I changed datetime settings to `time. RFC3339` for all log formats.

kuritka (approved):

controllers/depresolver/depresolver_test.go
```diff
@@ -1198,7 +1198,7 @@ func TestResolveLoggerNoColorInvalidValue(t *testing.T) {
 	config, err := resolver.ResolveOperatorConfig()
 	// assert
 	assert.NoError(t, err)
-	assert.Equal(t, true, config.Log.NoColor)
+	assert.Equal(t, false, config.Log.NoColor)
```
ytsarev, line 1201: Could you please explain the semantics of this test?  It is invalid, still intentionally NoError?

ytsarev (commented):
Looks great, just would be great to get some clarification on test sematics. 

We need to be listed on https://no-color.org/ after the merge! :)

controllers/depresolver/depresolver_test.go
somaritane: yes, it is expected that NoColor resolution falls back to `false` without error, when invalid value is used for NO_COLOR env var.

ytsarev (approved):

PR #406, opened 2021-03-23
log debug, optimization

kuritka:
linked to #331
- resolved config and satrtegy is logged by DEBUG
- merge Split brain TXT time stamp into one log.Info() message, refactor code

... feel free to come with places where log.Debug() is desirable

### DIFF
--- a/controllers/gslb_controller.go
+++ b/controllers/gslb_controller.go
@@ -27,6 +27,7 @@ import (
 	"github.com/AbsaOSS/k8gb/controllers/providers/dns"
 	"github.com/AbsaOSS/k8gb/controllers/providers/metrics"
 
+	str "github.com/AbsaOSS/gopkg/strings"
 	corev1 "k8s.io/api/core/v1"
 	v1beta1 "k8s.io/api/extensions/v1beta1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -85,6 +86,9 @@ func (r *GslbReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	if err != nil {
 		return result.RequeueError(fmt.Errorf("resolving spec (%s)", err))
 	}
+	log.Debug().
+		Str("Strategy", str.ToString(gslb.Spec.Strategy)).
+		Msg("Resolved strategy")
 	// == Finalizer business ==
 
 	// Check if the Gslb instance is marked to be deleted, which is
--- a/controllers/providers/assistant/gslb.go
+++ b/controllers/providers/assistant/gslb.go
@@ -189,32 +189,32 @@ func (r *GslbLoggerAssistant) InspectTXTThreshold(fqdn string, fakeDNSEnabled bo
 		log.Info().Msgf("Error contacting EdgeDNS server (%s) for TXT split brain record: (%s)", ns, err)
 		return err
 	}
-	var timestamp string
 	if len(txt.Answer) > 0 {
 		if t, ok := txt.Answer[0].(*dns.TXT); ok {
-			log.Info().Msgf("Split brain TXT raw record: %s", t.String())
-			timestamp = strings.Split(t.String(), "\t")[4]
+			timestamp := strings.Split(t.String(), "\t")[4]
 			timestamp = strings.Trim(timestamp, "\"") // Otherwise time.Parse() will miserably fail
+			timeFromTXT, err := time.Parse("2006-01-02T15:04:05", timestamp)
+			if err != nil {
+				log.Err(err).
+					Str("raw record", t.String()).
+					Str("raw timestamp", timestamp).
+					Msg("Split brain TXT: can't parse timestamp")
+				return err
+			}
+			now := time.Now().UTC()
+			diff := now.Sub(timeFromTXT)
+			log.Debug().
+				Str("raw record", t.String()).
+				Str("raw timestamp", timestamp).
+				Str("parsed", timeFromTXT.String()).
+				Str("diff", diff.String()).
+				Msg("Split brain TXT")
+
+			if diff > splitBrainThreshold {
+				return errors.NewResourceExpired(fmt.Sprintf("Split brain TXT record expired the time threshold: (%s)", splitBrainThreshold))
+			}
+			return nil
 		}
-	}
-
-	if len(timestamp) > 0 {
-		log.Info().Msgf("Split brain TXT raw time stamp: %s", timestamp)
-		timeFromTXT, err := time.Parse("2006-01-02T15:04:05", timestamp)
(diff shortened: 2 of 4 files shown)


controllers/providers/assistant/gslb.go
```diff
@@ -189,32 +189,24 @@ func (r *GslbLoggerAssistant) InspectTXTThreshold(fqdn string, fakeDNSEnabled bo
...
-		}
-
-		log.Info().Msgf("Split brain TXT parsed time stamp: %s", timeFromTXT)
-		now := time.Now().UTC()
-
-		diff := now.Sub(timeFromTXT)
-		log.Info().Msgf("Split brain TXT time diff: %s", diff)
+			timeFromTXT, err := time.Parse("2006-01-02T15:04:05", timestamp)
+			if err != nil {
+				log.Info().Msgf("Split brain TXT [raw record: %s, raw timestamp: %s]", t.String(), timestamp)
```
ytsarev, line 198: Maybe I am missing intention here but the message does not indicate that `time.Parse` has failed.

controllers/providers/assistant/gslb.go
```diff
@@ -189,32 +189,24 @@ func (r *GslbLoggerAssistant) InspectTXTThreshold(fqdn string, fakeDNSEnabled bo
...
+				return err
+			}
+			now := time.Now().UTC()
+			diff := now.Sub(timeFromTXT)
+			log.Info().Msgf("Split brain TXT [raw record: %s, raw timestamp: %s, parsed: %s, diff: %s]", t.String(), timestamp, timeFromTXT, diff)
```
ytsarev, line 203: We can move this one to Debug i think. No need to log it on every heartbeat on Info level

ytsarev (changes requested):
Couple of suggestions

controllers/gslb_controller.go
```diff
@@ -85,6 +86,7 @@ func (r *GslbReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	if err != nil {
 		return result.RequeueError(fmt.Errorf("resolving spec (%s)", err))
 	}
+	log.Debug().Msgf("resolved strategy:\n%s", str.ToString(gslb.Spec.Strategy))
```
somaritane, line 89: I'd avoid using new line (`\n`). It might beautify output in console, but when it comes to logging infra, additional formatting might complicate log parsing and further analysis.

controllers/providers/assistant/gslb.go
somaritane: Are we going to utilize zerolog structure logging capabilities and bring more context into logging events?
In this case the code would look like:
```suggestion
			log.Info().
				Str("raw record", t.String()).
				Str("raw timestamp", timestamp).
				Str("parsed", timeFromTXT.String()).
				Str("diff", diff.String()).
				Msg("Split brain TXT")
```
or
```suggestion
			log.Info().Fields(map[string]interface{}{
				"raw record":    t.String(),
				"raw timestamp": timestamp,
				"parsed":        timeFromTXT.String(),
				"diff":          diff.String()}).
				Msg("Split brain TXT")
```

main.go
```diff
@@ -65,6 +67,7 @@ func main() {
 		log.Err(err).Msg("can't resolve environment variables")
 		os.Exit(1)
 	}
+	log.Debug().Msgf("resolved config:\n%s", str.ToString(config))
```
somaritane, line 70: Consider removing newline (`\n`) from log message:
```suggestion
	log.Debug().Msgf("resolved config: %s", str.ToString(config))
```

somaritane (changes requested):
@kuritka, looks good, just few suggestions and general questions about log improvements.

kuritka:
Amended.

 I also added to factory
```go
	logger.Debug().
		Str("Format", l.log.Format.String()).
		Str("Level", l.log.Level.String()).
		Msg("Logger settings")
```
The reason I use `Level` instead of `level` is because `level` keyword is already occupied  within Logger context and is not printed.

controllers/gslb_controller.go
```diff
@@ -85,6 +86,7 @@ func (r *GslbReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	if err != nil {
 		return result.RequeueError(fmt.Errorf("resolving spec (%s)", err))
 	}
+	log.Debug().Msgf("Resolved strategy: %s", str.ToString(gslb.Spec.Strategy))
```
somaritane, line 89: Probably should be structured as well:
```suggestion
	log.Debug().
		Str("Strategy", str.ToString(gslb.Spec.Strategy)).
		Msg("Resolved strategy")
```

main.go
```diff
@@ -65,6 +67,7 @@ func main() {
 		log.Err(err).Msg("can't resolve environment variables")
 		os.Exit(1)
 	}
+	log.Debug().Msgf("Resolved config: %s", str.ToString(config))
```
somaritane, line 70: Consider structuring too:
```suggestion
	log.Debug().
		Str("config", str.ToString(config)).
		Msg("Resolved config")
```

somaritane (changes requested):
@kuritka , left some extra comments for affected log commands

kuritka:
@somaritane, thx for suggestions,  amended

controllers/gslb_controller.go
```diff
@@ -85,6 +86,9 @@ func (r *GslbReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	if err != nil {
 		return result.RequeueError(fmt.Errorf("resolving spec (%s)", err))
 	}
+	log.Debug().
```
somaritane, line 89: @kuritka, I noticed that gopkg ToString() preserves JSON pretty-print formatting, do you think it's possible to add possibility to strip resulting serialized JSON out of `\n` and `\t`?
Now it looks like this:
```
2021-03-25T15:31:16+01:00 DBG controllers/gslb_controller.go:92 > Resolved strategy Strategy="{\n\t\"type\": \"failover\",\n\t\"primaryGeoTag\": \"eu\",\n\t\"dnsTtlSeconds\": 30,\n\t\"splitBrainThresholdSeconds\": 300\n}"
```

somaritane (approved):
@kuritka, thanks for updates, lgtm!
P.S. There's one comment I left about gopkg ToString() JSON formatting, but I don't think it's in the scope of this PR.

controllers/gslb_controller.go
kuritka: @somaritane,  see: [gopkg issue](https://github.com/AbsaOSS/gopkg/issues/8)

ytsarev (approved):

PR #407, opened 2021-03-23
Added golangci-lint as pre-requisite to local setup doc

somaritane:


Makefile
```diff
@@ -219,7 +219,7 @@ ns1-secret:
 		--from-literal=apiKey=$${NS1_APIKEY}
 
 
-# run all linters from .golangci.yaml; see: https://golangci-lint.run/usage/install/#local-installation
+# runs golangci-lint aggregated linter
```
ytsarev, line 222: Imho `.golangci.yaml` still looks useful in the comment

```suggestion
# runs golangci-lint aggregated linter; see .golangci.yaml for linter list
```

ytsarev (changes requested):
Just a small suggestion

Makefile
somaritane: Thanks, @ytsarev, makes sense, amended with suggested change.

ytsarev (approved):

PR #413, opened 2021-03-26
Improve logging for missing environment variables

somaritane:
- Use real env variable names instead of internal variables in errors
- Fix `make run` target for missing environment variables


kuritka (approved):
good job, thx 👍

ytsarev (approved):

Issue #349, opened 2021-03-05
Automate permutation testing


ytsarev:
To avoid regressions connected with the upgrade procedure we need to automatically test upgrade from current to future release before stabilizing the new version.

ytsarev:
@kuritka @somaritane @k0da I guess we can close this one or do you envision more work in this area in foreseeable future?

somaritane:
> @kuritka @somaritane @k0da I guess we can close this one or do you envision more work in this area in foreseeable future?

@ytsarev yep, I think we can close this one, the scope is pretty much fulfilled by #429 and #435. Just one thing, as I mentioned earlier, "Permutation" and "permutation testing" are quite a broad terms, let's rename the issue with "Automate upgrade testing" in order to avoid confusions.

ytsarev:
@somaritane I was just trying to sound smart :D Closing this one

PR #417, opened 2021-03-30
Migration to networking.k8s.io/v1beta1

### LINKED ISSUE #401: automate ingress validation annotation

kuritka:
- migration gslb_types and upstream_types to `networking/v1beta1`
 - migration yamls in terratest, generates warnings see: [terratest #840](https://github.com/gruntwork-io/terratest/issues/840#issue-844462699)
 - migration apiGroup in roles
 - ~~**CRD change**~~ (is already done by #425) 

closes #401


k0da (approved):
LGTM

ytsarev (commented):
How the k8gb upgrade will look like with the CRD change? Can we please test permutation?

kuritka:
@ytsarev, necessary CRD change was implemented by #425. Would be happy to implement permutation testing in scope of #349 .

k0da (dismissed):

ytsarev:
@kuritka it's an api group change, if the automation is not yet there we need to at least test the upgrade from previous version manually

k0da (approved):

kuritka:
> @kuritka it's an api group change, if the automation is not yet there we need to at least test the upgrade from previous version manually

@ytsarev , upgrade test +  [local playground](https://github.com/AbsaOSS/k8gb/blob/master/docs/local.md) test passed

ytsarev (approved):

PR #419, opened 2021-03-31
Bump operator SDK to v1.5.0

### LINKED ISSUE #376: Upgrade to operator-sdk v1.5.0
https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.5.0/

kuritka:
close #376

 - v1.4.0 Change operator's finalizer names (`finalizer.cache.example.com` should be changed to `cache.example.com/finalizer`)
 - v1.4.0  directory argument in `docker-build` was moved to the last position to align with `podman's` expectations, making substitution cleaner.
 - v1.5.0 `PROJECT` config version `3-alpha`  must be upgraded to `"3"`
 - v1.5.0  Upgrade controller-runtime to `v0.7.2` controller runtime.
The package changed, and controller must implement Reconciler interface.
To be able to build project, I had to implement Reconciler interface (by adding context to Reconcile function)
and use `client.Object` in handler functions

### links
- https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.4.0/
- https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.5.0/


PROJECT
```diff
@@ -6,7 +6,7 @@ resources:
 - group: k8gb
   kind: Gslb
   version: v1beta1
-version: 3-alpha
+version: "3"
```
ytsarev, line 9: Is it required to be a string?

kuritka: Not sure if it is required, they use it like a string:
 - https://github.com/operator-framework/operator-sdk/pull/4613/files#diff-81aae5fab7d1e80a3b56a8736536bb188e82fe5141d55b22d089c41cd4ec9bfcR71-R82
 - https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.5.0/#project-config-version-3-alpha-must-be-upgraded-to-3

somaritane: I guess using `3` might cause issues with version parsing in osdk toolset, as it's treated as integer type

controllers/gslb_controller.go
```diff
@@ -53,7 +53,7 @@ type GslbReconciler struct {
 }
 
 const (
-	gslbFinalizer           = "finalizer.k8gb.absa.oss"
+	gslbFinalizer           = "k8gb.absa.oss/finalizer"
```
somaritane, line 56: This might not work well with existing gslbs having "finalizer.k8gb.absa.oss" finalizer set. Imo we need to use an array of valid finalizers here (`["finalizer.k8gb.absa.oss" , "k8gb.absa.oss/finalizer"]`) in order to support smooth upgrade for set/get finalizer logic.

PROJECT
ytsarev: cool, thanks for double checking

controllers/gslb_controller.go
ytsarev: @somaritane Conceptually agreed. But how the removal logic will work in that case? Just in case, in order to delete resource we need to remove *all* finalizers from the object

ytsarev: Sounds like a plan

kuritka: amended, test coverage for finalizers

kuritka:
> A little bit concerned with controller-runtime version bump. Was the stuff e2e tested (outside of terratest) ?

@ytsarev, already checked with Upgrade testing

> A little bit concerned with controller-runtime version bump. Was the stuff e2e tested (outside of terratest) ?

Hi, Upgrade test is successful + I also made [local playground](https://github.com/AbsaOSS/k8gb/blob/master/docs/local.md) test.

k0da (approved):
👍

ytsarev (approved):

ytsarev:
Cool stuff, let's merge it

PR #420, opened 2021-04-01
Serve specific zone by CoreDNS

k0da:
Serving . could lead into long resolves. Especially while service A
record only and alpine based workloads do both AAAA and A queries.

### DIFF
--- a/chart/k8gb/values.yaml
+++ b/chart/k8gb/values.yaml
@@ -10,7 +10,7 @@ k8gb:
   # image tag defaults to Chart.AppVersion, see Chart.yaml
   # but can be overrided with imageTag key
   # imageTag:
-  dnsZone: "cloud.example.com" # dnsZone controlled by gslb
+  dnsZone: &dnsZone "cloud.example.com" # dnsZone controlled by gslb
   edgeDNSZone: "example.com" # main zone which would contain gslb zone to delegate
   edgeDNSServer: "1.1.1.1" # use this DNS server as a main resolver to enable cross k8gb DNS based communication
   clusterGeoTag: "eu" # used for places where we need to distinguish between differnet Gslb instances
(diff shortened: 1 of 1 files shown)


ytsarev (commented):
Before we merge it - could you please check the yaml anchor propagation in ArgoCD context? I recall some issues with it before we switched to `.` in coredns

k0da:
@ytsarev Argo can serve it through metachart

kuritka (approved):

ytsarev:
@k0da okay, but will it work here https://github.com/bison-cloud-platform/bks-registry/tree/master/argocd/k8gb ?

somaritane:
> @k0da okay, but will it work here https://github.com/bison-cloud-platform/bks-registry/tree/master/argocd/k8gb ?

@ytsarev, we're talking about default chart values here, project using k8gb helm chart can always fallback to dnsZone duplication in their custom values, if there are any issues with anchor annotation.

somaritane (approved):

ytsarev:
@somaritane yes, but in existing argo configuration with the removal of `.` default and not working yaml anchoring the coredns configuration will config empty value for the zone breaking existing configuration completely.

So I still suggest testing the actual behavior with the existing argo-based deployments as the coredns zone parameter is not overlayed there at the moment.

k0da:
@ytsarev it have its own values and fixed chart version. This merge won't break it.

ytsarev (changes requested):
@k0da it doesn't have its own value for coredns zone. It is not overlayed either. If we merge it right now - we forget about it and the future upgrade will potentially break our argo based artifactory deployments.

It will also make the case of undocumented requirement of explicit override of coredns zone in argocd context which will deviate from standard helm installation.

So please check it.

somaritane:
> @somaritane yes, but in existing argo configuration with the removal of `.` default and not working yaml anchoring the coredns configuration will config empty value for the zone breaking existing configuration completely.
> 
> So I still suggest testing the actual behavior with the existing argo-based deployments as the coredns zone parameter is not overlayed there at the moment.

@ytsarev makes sense.

PR #425, opened 2021-04-02
Extend Gslb CRD with additionalPrinterColumns

ytsarev:
Use https://book.kubebuilder.io/reference/generating-crd.html#additional-printer-columns
to add `additionalPrinterColumns` exposing strategy and geoTag of Gslb CR

From the user perspective it looks like
```sh
kubectl -n test-gslb get gslb
NAME                 STRATEGY     GEOTAG
test-gslb            roundRobin   eu-west-1
test-gslb-failover   failover     eu-west-1
```

### DIFF
--- a/api/v1beta1/gslb_types.go
+++ b/api/v1beta1/gslb_types.go
@@ -59,6 +59,8 @@ type GslbStatus struct {
 // +kubebuilder:subresource:status
 
 // Gslb is the Schema for the gslbs API
+// +kubebuilder:printcolumn:name="strategy",type=string,JSONPath=`.spec.strategy.type`
+// +kubebuilder:printcolumn:name="geoTag",type=string,JSONPath=`.status.geoTag`
 type Gslb struct {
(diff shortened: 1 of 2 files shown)


k0da (changes requested):
While there, can we please have host as well?

ytsarev:
> While there, can we please have host as well?

There can be many hosts for the same Gslb also there is a representation type limitation of
```
 must be one of boolean,date,integer,number,string
```
So 
```
k -n test-gslb get gslb test-gslb-failover
NAME                 GEOTAG      HEATHYRECORDS                                                            SERVICEHEALTH
test-gslb-failover   eu-west-1   map[failover.test.k8gb.io:[34.254.84.32 52.212.133.90 54.195.241.226]]   map[failover.test.k8gb.io:Healthy]
```
doesn't look cool

So I don't know how to properly represent the `object` in the print column.

k0da (approved):

PR #426, opened 2021-04-06
Enable Ingress to Gslb Owner Reference

### LINKED ISSUE #361: Issue when deleting ingress rule or annotations removal doesn't remove the gslb records
Hi Team

I have deployed the k8gb to a cluster with Kube 1.16.15

I have used the k8gb annotations for the ingress and deleted it

what I have found is that the gslb record does not get deleted but tries to get recreated although the ingress has been deleted.

is there a way by means of annotation removal or when ingress is deleted that the ingress >>>> gslb relationship works smoothly and seamlessly to delete the actual gslb record when the ingress is unannotated or deleted

I had to manually use

k delete gslb -n namespace namegslbingress

ytsarev:
* Enables proper Ingress -> Gslb CR ownership in case of
annotation based management approach
with pre-existing Ingress as in

https://www.k8gb.io/docs/ingress_annotations.html

* Resolves #361

* Covered by Terratest


Makefile
```diff
@@ -434,8 +434,8 @@ endef
 
 define docker-test-build-push
 	docker build . -t k8gb:$(COMMIT_HASH)
-	docker tag k8gb:$(COMMIT_HASH) $(REPO):v$(COMMIT_HASH)
-	docker push $(REPO):v$(COMMIT_HASH)
+	docker tag k8gb:$(COMMIT_HASH) $(REPO):$(COMMIT_HASH)
+	docker push $(REPO):$(COMMIT_HASH)
```
kuritka, line 438: Although without `v` it is valid semver, wouldn't we keep it for consistency? (we use prefix `v` everywhere)

ytsarev: in this specific case `v` is getting mixed with short hash and looks like `vc105f41` instead of `c105f41` which is quite misleading.
In any case it
* was broken for local setup
* exists for local testing only

kuritka (changes requested):
`go.sum` changed so much although `go.mod` remains the same.

Makefile
kuritka: true 👍

ytsarev:
> `go.sum` changed so much although `go.mod` remains the same.

@kuritka I clearly share the same observation. What is the suggestion though?:)

ytsarev:
@kuritka I've also run `go mod tidy` to reduce number of changes, still there are some left

kuritka:
@ytsarev , the point is, the `go.sum` is not refreshed in master as well.

`go mod download` downloads necessary packages

kuritka (approved):

chart/k8gb/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: k8gb
 description: A Helm chart for Kubernetes Global Balancer
 type: application
-version: v0.7.7
-appVersion: v0.7.7
+version: v0.7.8
```
k0da, line 5: Are we doing release?

ytsarev: @k0da not at the moment. 

Still whenever we make core controller modifications we tend to bump the version even without an official helm release.

It will make git-based helm installations consistent and straightforward, avoiding confusing situations when the tip of the branch contains the fix meanwhile helm chart configuration points to the old controller build

ytsarev:
@kuritka 
I've used `make terratest` local which also incorporates `go mod download`. 
What do I miss? Or is it all good now and we just missed the bumps before?

kuritka:
@ytsarev  I missed bump together with bumping `go.mod`.  All `go mod download` in cross the code caused it unnoticed.

kuritka (dismissed):

chart/k8gb/Chart.yaml
k0da: Then we should really split chart and appVersions

somaritane: @ytsarev IMO since this version bumping rule is not obvious (I can hardly remember it myself :) ), it should be reflected in contributors documentation. 
I remember there were we pretty serious changes and refactoring made in controller code, and these were not reflected in chart version bump, so we should be consistent here.
Also, what happens if I try to checkout the project and deploy full cluster setup locally, when chart version is bumped, but release is not made, hence docker image for corresponding version is not publicly available?

controllers/gslb_controller.go
```diff
@@ -248,12 +249,18 @@ func (r *GslbReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			}
 		}
 
+		err = controllerutil.SetControllerReference(ingressToReuse, gslb, r.Scheme)
+		if err != nil {
+			log.Err(err).Msg("Can't set the ownership reference(Ingress owns Gslb)")
```
somaritane, line 254: Consider adding information about involved Gslb and Ingress objects using zerolog structured logging:
```suggestion
			log.Err(err).
			Str("Ingress", ingressToReuse.Name).
			Str("Gslb", gslb.Name).
			Msg("Cannot set the Ingress as the owner of the Gslb")
```

somaritane (changes requested):
@ytsarev added few comments

chart/k8gb/Chart.yaml
ytsarev: @k0da could you please elaborate why? I really fail to see why we need to add additional complexity

@somaritane it was almost the very beginning of the project, we agreed to keep it simple and keep appVersion and Version in sync until it is really necessary to deviate. Currently, it is more of tribe knowledge. Good idea to document it in general, not in scope of this PR though.
`v0.7.8` is available in the public docker hub repo but guarded against public consumption by unreleased helm chart.

ytsarev:
@somaritane suggested change incorporated, thanks

chart/k8gb/Chart.yaml
somaritane: @ytsarev, was `v0.7.8` pushed to docker hub manually?

k0da: It just leads to confusion. And git based installation would imply setting override tag eitherway. v0.7.8 doesn't reflect any git status.

ytsarev: @somaritane yep, with 
```
make docker-build-multi
make docker-push-multi
```
Again, it is just a tag and does not indicate any release

@k0da deviation between appVersion and Version leads to even more confusion.

Anyway, what we are discussing here? I can exlcude chart version bump from PR making it effectively not-installable at the moment. Is that what we want?

k0da: I'd preffer to not include chart version bump.

What do you mean by not-installable? You can always set image version.

ytsarev: Meaning that the tip of the master branch will install an old version of the controller by default.

Ok, I will exclude it, no worries

ytsarev: @k0da @somaritane no version bump anymore

somaritane: @ytsarev , I think we should avoid manual publishing of docker images with version tag outside of release process. There's no guaranty that the image really reflects the claimed version and git state and also it is prone to errors.

For example, I can see the following image Docker Hub https://hub.docker.com/layers/absaoss/k8gb/v0.7.8-arm64/images/sha256-4c13f5931cbdfda226f915affbecf8ce9ffe856da0cf1575401eba4a7bbcb1cd?context=explore,
which has `arm64` arch in the version, but at the same time has `linux/amd64` arch in metadata.

ytsarev: then we have a bug in make targets.

Again, it was not a release tag and not available for end users. But I understand the confusion, let's keep the stable-looking versions to the release pipelines.

somaritane (approved):
@ytsarev lgtm 👍

kuritka (approved):

PR #432, opened 2021-04-08
FIX: race condition detected

kuritka:
terratests failed sometimes

 - https://github.com/AbsaOSS/k8gb/runs/2298208578?check_suite_focus=true#step:6:196
 - https://github.com/AbsaOSS/k8gb/runs/2296105748?check_suite_focus=true#step:6:2248

The reason was that GSLB was deleted, although it was already removed by removing ingress. Depending on system performance, sometimes it passed, sometimes not.


k0da (approved):
👍

ytsarev (approved):

PR #439, opened 2021-04-13
Service CoreDNS Corefile by k8gb chart

k0da:
In order to propogate dnsZone into CoreDNS Corefile we need to service
configmap ourself and instruct CoreDNS chart to skip config creation.
This is done by skipConfig: true option.

### DIFF
--- a/chart/k8gb/templates/coredns-cm.yaml
+++ b/chart/k8gb/templates/coredns-cm.yaml
@@ -1,0 +1,20 @@
+kind: ConfigMap
+metadata:
+  labels:
+{{ include "chart.labels" . | indent 4  }}
+  name: {{ .Release.Name }}-coredns
+apiVersion: v1
+data:
+  Corefile: |-
+    {{ .Values.k8gb.dnsZone }}:53 {
+        errors
+        health
+        ready
+        loadbalance round_robin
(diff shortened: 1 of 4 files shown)


chart/k8gb/templates/coredns-cm.yaml
```diff
@@ -0,0 +1,23 @@
+kind: ConfigMap
+metadata:
+  labels:
+    app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+    app.kubernetes.io/instance: {{ .Release.Name | quote }}
+    helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"
+    app.kubernetes.io/name: {{ template "coredns.name" . }}
```
ytsarev, line 7:
```suggestion
   {{ include "chart.labels" . | indent 4  }}
```

ytsarev (changes requested):

chart/k8gb/templates/coredns-cm.yaml
k0da: @ytsarev thanks amended

ytsarev (approved):

PR #443, opened 2021-04-14
FIX: Annotate and Label existing DNSEndpoints

k0da:
When upgrading from 0.7.5 to 0.7.6 new generated objectmeta wasn't
concidered during DNSEndpoint reconcilation. Fix this by respecting
computed Labels and Annotations while reconciling DNSEndpoint.

fixes https://github.com/AbsaOSS/k8gb/issues/324

### DIFF
--- a/controllers/gslb_controller_test.go
+++ b/controllers/gslb_controller_test.go
@@ -395,14 +395,18 @@ func TestLocalDNSRecordsHasSpecialAnnotation(t *testing.T) {
 
 	// act
 	createHealthyService(t, &settings, serviceName)
+	// delete DNSEndpoint so we can pretend it wasn't annotated before
+	deleteDNSEndpoint(t, &settings)
+	createDNSEndpoint(t, &settings)
 	defer deleteHealthyService(t, &settings, serviceName)
+	defer deleteDNSEndpoint(t, &settings)
 	reconcileAndUpdateGslb(t, settings)
 	err = settings.client.Get(context.TODO(), settings.request.NamespacedName, dnsEndpoint)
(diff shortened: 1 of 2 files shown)


controllers/gslb_controller_test.go
```diff
@@ -395,14 +395,18 @@ func TestLocalDNSRecordsHasSpecialAnnotation(t *testing.T) {
...
 	defer deleteHealthyService(t, &settings, serviceName)
+	defer deleteDNSEndpoint(t, &settings)
 	reconcileAndUpdateGslb(t, settings)
 	err = settings.client.Get(context.TODO(), settings.request.NamespacedName, dnsEndpoint)
 	require.NoError(t, err, "Failed to load DNS endpoint")
 	got := dnsEndpoint.Annotations["k8gb.absa.oss/dnstype"]
 
 	// assert
-	assert.Equal(t, got, want, "got:\n %q annotation value,\n\n want:\n %q", got, want, got, want)
+	assert.Equal(t, want, got, "got:\n %q annotation value,\n\n want:\n %q", got, want, got, want)
```
ytsarev, line 409:
```suggestion
	assert.Equal(t, want, got, "got:\n %q annotation value,\n\n want:\n %q", got, want)
```

ytsarev (changes requested):
Looks great, just a small suggestion re assert failure output

controllers/gslb_controller_test.go
k0da: @ytsarev thanks, amended

ytsarev (approved):

PR #444, opened 2021-04-14
Shrink k8gb role to what is really required

k0da:
fixes https://github.com/AbsaOSS/k8gb/issues/330

### DIFF
--- a/chart/k8gb/templates/role.yaml
+++ b/chart/k8gb/templates/role.yaml
@@ -7,54 +7,11 @@ rules:
 - apiGroups:
   - ""
   resources:
-  - pods
-  - nodes
-  - services
-  - services/finalizers
   - endpoints
-  - persistentvolumeclaims
-  - events
-  - configmaps
-  - secrets
+  - services
   verbs:
-  - '*'
-- apiGroups:
-  - apps
-  resources:
-  - deployments
-  - daemonsets
-  - replicasets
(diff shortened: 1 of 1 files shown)


chart/k8gb/templates/role.yaml
```diff
@@ -7,54 +7,11 @@ rules:
 - apiGroups:
   - ""
   resources:
-  - pods
-  - nodes
   - services
-  - services/finalizers
   - endpoints
```
ytsarev, line 11: For Endpoints we probably need only `get` verb

chart/k8gb/templates/role.yaml
```diff
@@ -7,54 +7,11 @@ rules:
 - apiGroups:
   - ""
   resources:
-  - pods
-  - nodes
   - services
```
ytsarev, line 10: Only `get` for Services

ytsarev (changes requested):

chart/k8gb/templates/role.yaml
k0da: makes sense

chart/k8gb/templates/role.yaml
k0da: Also dropped events, so amended

ytsarev (approved):

PR #447, opened 2021-04-15
Stabilize local setup

### LINKED ISSUE #446: Split quickstart focused and developer focused make targets
Currently, we use `make deploy-full-local-setup` for both local contributor setup and the user quickstart.

From time to time it leads to an unclear state of deployment when e.g. new k8gb controller version is not yet released meanwhile there are already relevant (sometimes non-compatible) chart modifications at the top of the master branch.

To address that we might:

* Switch `deploy-full-local-setup` target to deploy stable versions of controller+helm chart
* Dedicate `deploy-candidate-with-helm` target to developer scenario deploying testing version of controller and latest unreleased version of chart
* Document it in CONTRIBUTING.md

ytsarev:
* Deploy stable helm chart together with stable k8gb controller version
* Deploy test helm chart together with temporary test build of k8gb
  controller from current tree
* Update contributor documentation
* Resolves #446

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,6 +29,25 @@ This is a rough outline of what a contributor's workflow looks like:
 
 Thanks for contributing!
 
+### Local setup
+
+```sh
+make deploy-full-local-setup
+```
+Deploys k8gb from scracth including:
+
+* 2 local clusters
+* Stable k8gb helm chart
+* Gslb Custom Resources examples
+* Test applications
+
+```sh
+make upgrade-candidate
+```
+Performs upgrade of k8gb helm chart and controller
+to the testing version that is built from your current
+development tree.
+
 ### Testing
 
 * Unit tests should be updated for any functional code change at [test suite location](https://github.com/AbsaOSS/k8gb/tree/master/controllers/gslb_controller_test.go).
@@ -85,7 +104,8 @@ The first line is the subject and should be no longer than 70 characters, the se
 
 ## Documentation
 
-If the contribution changes the existing APIs or user interface it must include sufficient documentation to explain the use of the new or updated feature. Likewise the [CHANGELOG][changelog] should be updated with a summary of the change and link to the pull request.
+If the contribution changes the existing APIs or user interface it must include sufficient documentation to explain the use of the new or updated feature.
+[CHANGELOG](CHANGELOG.md) is automatically generated from Github PRs and Issues. Please use special [keywords](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) to link PR to and Issue for a clean changelog generation.
 
 
 [golang-style-doc]: https://github.com/golang/go/wiki/CodeReviewComments
--- a/.github/workflows/upgrade-testing.yaml
+++ b/.github/workflows/upgrade-testing.yaml
@@ -29,8 +29,8 @@ jobs:
       - name: K8GB deploy stable
         run: make deploy-stable
 
-      - name: K8GB deploy-candidate-with-helm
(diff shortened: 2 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -29,6 +29,25 @@ This is a rough outline of what a contributor's workflow looks like:
 
 Thanks for contributing!
 
+### Local setup
+
+```sh
+make make deploy-full-local-setup
```
k0da, line 35: make duplicated

ytsarev: thanks for the catch, amended

kuritka: basically lgtm, but do we need `deploy-full-local-setup` anymore ?

kuritka (approved):

CONTRIBUTING.md
```diff
@@ -29,6 +29,25 @@ This is a rough outline of what a contributor's workflow looks like:
 
 Thanks for contributing!
 
+### Local setup
```
somaritane, line 32: There's local.md document that describes local setup. Is this section somehow different?

ytsarev: it's the same but the reflection from the contributor standpoint. Actually to address the same, maybe in a more concise and focused way. 

Basically this an attempt to differentiate generic playground vs contributor testing setup as in #446 problem statement

CONTRIBUTING.md
```diff
@@ -29,6 +29,25 @@ This is a rough outline of what a contributor's workflow looks like:
...
+
+```sh
+make deploy-full-local-setup
```
somaritane, line 35: Maybe we can rename this target to simply become `deploy` for the sake of simplicity of first-user experience?
iirc we don't have one-cluster setup targets anyway, and this one acts as default deployment in most cases.

ytsarev: deploy here means composite of 2 clusters+ k8gb helm chart.+ CR + test apps. Just `deploy` is semantically more about k8gb helm chart only. That's why we have so many different `deploy-*` targets. To handle different levels of granularity

CONTRIBUTING.md
somaritane: From what I see in the changes, `deploy-full-local-setup` target is still being used in both cases, playground and contributor.
It is mentioned in README.md's  QuickStart section. Is this PR considered as just an initial phase of #446 implementation?

ytsarev: It's kind of layered. First, you deploy stable with `deploy-full-local-setup` then apply test candidate with `upgrade-candidate`.  In QuickStart it is `deploy-full-local-setup`  only

somaritane: OK I see, makes sense

k0da (approved):
LGTM

CONTRIBUTING.md
somaritane: It's hard to see mentioned target semantics unless you thoroughly investigate Makefile. 
Deploy is just deploy to me and can be done in different ways (helm, kustomize...), but I don't see this in target names. Also most of `deploy-*` targets we use are deploying to 2 clusters, but I don't see mentioning of that as well, also they all deploy to local setup:
- deploy-full-local-setup
- deploy-stable
- deploy-candidate
- upgrade-candidate
Hence I suggest renaming of `deploy-full-local-setup` to simply `deploy`, because there's no sense of using `full` and `local-setup` there if we want to be consistent and simple.

ytsarev: @somaritane I suggest we create another Issue for renaming-based refactoring if needed. This PR mainly targets the current degraded state of `deploy-full-local-setup` and unreliable state of `upgrade-testing` pipeline.

somaritane: @ytsarev Sure, let's split

somaritane (approved):
lgtm

PR #452, opened 2021-04-20
Enhance terratest suite with ability to be executed against real clusters 

### LINKED ISSUE #350: Reuse/enhance terratest test suite for real cluster validation
To avoid human error-prone post-upgrade cluster validation we can reuse and potentially extend the existing [terratest suite](https://github.com/AbsaOSS/k8gb/tree/master/terratest) to be executable against real cluster k8gb deployments.

Automated verification of correctly functioning load-balancing strategies will be a very good post-upgrade smoke test

ytsarev:
* Ability to point the terratest suite over the real pair of clusters
* Make tests reliable in parallelization scenario
* Multiple inline enhancements
* Execution against AWS+route53 reference setup takes ~3 minutes
* Resolves #350


terratest/test/helpers.go
```diff
@@ -47,8 +57,8 @@ func GetIngressIPs(t *testing.T, options *k8s.KubectlOptions, ingressName string
 }
 
 // Dig gets sorted slice of records related to dnsName
-func Dig(t *testing.T, dnsServer string, dnsPort int, dnsName string) ([]string, error) {
-	port := fmt.Sprintf("-p%v", dnsPort)
+func Dig(t *testing.T, dnsServer string, dnsPort string, dnsName string) ([]string, error) {
+	port := fmt.Sprintf("-p%s", dnsPort)
 	dnsServer = fmt.Sprintf("@%s", dnsServer)
 
 	digApp := shell.Command{
```
somaritane, line 64: Just a thought, should we rather use `github.com/miekg/dns` instead of shell command? 
Like here https://github.com/AbsaOSS/coredns-crd-plugin/blob/main/terratest/test/helper.go#L35

terratest/test/helpers.go
```diff
@@ -36,6 +39,13 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+var dnsZone = getEnv("GSLB_DOMAIN", "cloud.example.com")
+var dnsServer1 = getEnv("DNS_SERVER1", "localhost")
+var dnsServer1Port = getEnv("DNS_SERVER1_PORT", "5053")
+var dnsServer2 = getEnv("DNS_SERVER2", "localhost")
+var dnsServer2Port = getEnv("DNS_SERVER2_PORT", "5054")
+var primaryGeoTag = getEnv("PRIMARY_GEO_TAG", "eu")
+
```
kuritka, line 48: consider create file `init.go` and use this instead:

```go
package test

import "github.com/AbsaOSS/gopkg/env"

type clusterSettings struct {
	DNSZone       string
	PrimaryGeoTag string
	DNSServer1    string
	Port1         string
	DNSServer2    string
	Port2         string
}

var (
	settings clusterSettings
)

func init(){
	settings = clusterSettings{
		DNSZone:       env.GetEnvAsStringOrFallback("GSLB_DOMAIN", "cloud.example.com"),
		PrimaryGeoTag: env.GetEnvAsStringOrFallback("PRIMARY_GEO_TAG", "eu"),
		DNSServer1:    env.GetEnvAsStringOrFallback("DNS_SERVER1","localhost"),
		Port1:         env.GetEnvAsStringOrFallback("DNS_SERVER1_PORT","5053"),
		DNSServer2:    env.GetEnvAsStringOrFallback("DNS_SERVER2","localhost"),
		Port2:         env.GetEnvAsStringOrFallback("DNS_SERVER2_PORT","5054"),
	}
	// add different things needs to be initialised here....
}

```
`init()` is used for initializing packages, function should start before tests (btw: after variable initialison)

terratest/test/helpers.go
```diff
@@ -47,8 +57,8 @@ func GetIngressIPs(t *testing.T, options *k8s.KubectlOptions, ingressName string
 }
 
 // Dig gets sorted slice of records related to dnsName
-func Dig(t *testing.T, dnsServer string, dnsPort int, dnsName string) ([]string, error) {
-	port := fmt.Sprintf("-p%v", dnsPort)
+func Dig(t *testing.T, dnsServer string, dnsPort string, dnsName string) ([]string, error) {
```
kuritka, line 60: I know it is out of the scope of this task, just reacting to @somaritane note about `mikeg dns`, maybe we can reuse gopkg, so implementation would look like this:
```go
func Dig(t *testing.T, dnsServer string, dnsPort string, dnsName string) ([]string, error) {
	return dns.Dig(fmt.Sprintf("%s:%v", dnsServer,dnsPort),dnsName)
}
```

or we can completelly remove `Dig` function in helper

terratest/test/k8gb_ingress_annotation_failover_test.go
```diff
@@ -58,7 +58,7 @@ func TestK8gbIngressAnnotationFailover(t *testing.T) {
 
 	ingress := k8s.GetIngress(t, options, "test-gslb-annotation-failover")
 	require.Equal(t, ingress.Name, "test-gslb-annotation-failover")
-	assertGslbStatus(t, options, "test-gslb-annotation-failover", "notfound.cloud.example.com:NotFound roundrobin.cloud.example.com:NotFound unhealthy.cloud.example.com:NotFound")
+	assertGslbStatus(t, options, "test-gslb-annotation-failover", "notfound."+dnsZone+":NotFound roundrobin."+dnsZone+":NotFound unhealthy."+dnsZone+":NotFound")
 	assertGslbSpec(t, options, "test-gslb-annotation-failover", ".spec.strategy.type", "failover")
 	assertGslbSpec(t, options, "test-gslb-annotation-failover", ".spec.strategy.primaryGeoTag", "eu")
```
kuritka, line 63: shouldn't be `primaryGeoTag` or `settings.PrimaryGeoTag` ?

kuritka (changes requested):
please consider my comments

terratest/test/helpers.go
ytsarev: @somaritane @kuritka what is the motivation to use miekg dns in the e2e tests? Not strongly against it, just trying to understand. For me `dig` invocation is directly  reproducible in the CLI and I see a value in it.

kuritka: @ytsarev , from my side: Main reason, it is reusing gopkg which reduces complexity of `Dig` function in `helpers.go`. 

gopkg abstracts implementation details, but internally it uses mikeg.

somaritane: @ytsarev, for me motivation is to be able to use programmatic API, instead of external shell command and parse its results, which might be cumbersome when we test some complex cases. 
Also in this case we don't have to ensure that test run environment has `dig` installed.

ytsarev: @somaritane I might be crushing fancy golang party here but the whole `terratest` as a tool is a wrapper around shell commands :D And it is for a reason. We are writing e2e tests mimicking customer behavior, not extending the core controller.

somaritane: @ytsarev I tend to disagree here. 
Speaking about `terratest`: while it offers wrappers for commands like `helm` and `kubectl`, its functions for k8s primitives are using [k8s apimachinery](https://github.com/gruntwork-io/terratest/blob/master/modules/k8s/ingress.go#L19), it's cloud providers are using respective go apis and sdks. 
I guess, this is also for a reason. It would be not feasible to parse CLI response every time, even in json format.

As I previously pointed out [before](https://github.com/AbsaOSS/k8gb/pull/452#discussion_r616756472), similar helper in https://github.com/AbsaOSS/coredns-crd-plugin/blob/main/terratest/test/helper.go#L35 has been rewritten in order to use `github.com/miekg/dns` instead of `dig` wrapper, and we were discussing about consolidating the code in `gopkg`. Do you consider this bad idea? Should it be rewritten back?
`miek/dns` allowed to efficiently test for proper RCODE in response for `AAAA` query, as well as resolved IPs, instead of parsing `dig` output for both.

https://github.com/miekg/dns is pretty stable and mature project, supporting many DNS-related RFCs, having wide community of [users](https://github.com/miekg/dns#users), so I don't know how it maybe inferior to pure `dig`.

And if we talk about real e2e tests mimicking customer behavior, then we should be using something close to the browser or http client and check if we really hit test app endpoint with particular IP address.

ytsarev: @somaritane No need to rewrite anything and I agree almost on all points.

There was no real need to sell me https://github.com/miekg/dns as it is the main library of coredns which is an integral part of k8gb from day 0. 

Also, we already use it for k8gb testing in places where it makes direct sense, e.g. https://github.com/AbsaOSS/k8gb/blob/master/controllers/fakedns.go#L27

`coredns-crd-plugin` also had a proper use case for it.

But in this specific test suite, I do not observe the need except for pure aesthetics reasons at the moment.

When I execute the test suite I see output like
```sh
 Running command dig with args [-p5053 @localhost terratest-failover.k8gb-test-nonprod.cloud.example.com +short]
```
If something goes wrong I can directly copy-paste it to my shell and run `dig` without any additional cognitive load of params substitution.

And I'm pretty happy with this flow.

somaritane: @ytsarev In your initial comment https://github.com/AbsaOSS/k8gb/pull/452#discussion_r617635483, you've asked for reasoning about using programmatic api vs command-line wrapper. @kuritka and I provided them.
It feels like at the end you have strong argument to use dig wrapper.
But imho it is not close to mimicking e2e customer behavior, customer use case is not to run `dig` CLI with @localhost as resolver, but rather access service endpoint via browser or http client.

ytsarev: @somaritane sure, but there are different k8gb customers I guess. Here we are focusing more on the ones who operate k8gb, not consuming the app. Dig-based tests are checking DNS response correctness the same way as the person validating the lb strategy in action just in an automated way... Curling would not expose the same level of detailed check.

terratest/test/k8gb_basic_failover_test.go
```diff
@@ -43,14 +43,14 @@ func TestK8gbBasicFailoverExample(t *testing.T) {
...
+	namespaceName := fmt.Sprintf("k8gb-test-failover-%s", strings.ToLower(random.UniqueId()))
 
 	// Here we choose to use the defaults, which is:
 	// - HOME/.kube/config for the kubectl config file
 	// - Current context of the kubectl config file
 	// - Random namespace
-	optionsContext1 := k8s.NewKubectlOptions("k3d-test-gslb1", "", namespaceName)
-	optionsContext2 := k8s.NewKubectlOptions("k3d-test-gslb2", "", namespaceName)
+	optionsContext1 := k8s.NewKubectlOptions(getEnv("K8GB_CLUSTER1", "k3d-test-gslb1"), "", namespaceName)
+	optionsContext2 := k8s.NewKubectlOptions(getEnv("K8GB_CLUSTER2", "k3d-test-gslb2"), "", namespaceName)
```
kuritka, line 53: isn't this `settings.Cluster1`, `settings.Cluster2` ?

kuritka (changes requested):

terratest/test/k8gb_basic_failover_test.go
ytsarev: @kuritka thanks for the catch, updated

kuritka (approved):

PR #459, opened 2021-04-27
Shorten NS names for zone delegation

### LINKED ISSUE #456: Shorten NS names for zone delegation
As of now, k8gb constructs NS name for delegated zone from the following chunks:

`gslb-ns-<FlatDNSZone>-<ClusterGeoTag>.<EdgeDNSZone>`, where
`FlatDNSZone` is `DNSZone` with `'.'`s replaced by `'-'`s, e.g. `cloud.example.com -> cloud-example-com`

so for 
```
DNSZone = cloud.example.com 
ClusterGeoTag = eu
EdgeDNSZone = example.com
```
resulting NS server name looks like:
```
gslb-ns-cloud-example-com-eu.example.com
```
max length of DNS label is 63 characters
which leaves max allowed length for `<ConvertedDNSZone>-<ClusterGeoTag>` at 55 characters.

While it might be enough for relatively short DNSZones, like `cloud.example.com`,
for longer DNSZones it might quickly exceed the 55 characters limit and lead to delegated zone request error from EdgeDNS:

```
k8gb-77bf5b98f4-lwxm8 k8gb 2021/04/22 10:06:49 WAPI request error: 400('400 Bad Request')
k8gb-77bf5b98f4-lwxm8 k8gb Contents:
k8gb-77bf5b98f4-lwxm8 k8gb { "Error": "AdmConDataError: A domain label is longer than 63 characters",
k8gb-77bf5b98f4-lwxm8 k8gb   "code": "Client.Ibap.Data",
k8gb-77bf5b98f4-lwxm8 k8gb   "text": "A domain label is longer than 63 characters"
k8gb-77bf5b98f4-lwxm8 k8gb }
```

Logic to construct NS should be improved to give max possible room for DNS zone, while keep NS names unique
k8gb documentation should reflect the limits for DNSZone+ClusterGeoTag combination

kuritka:
closes #456

`nsServerName` and `nsExtServerNames` - shortly nsnames are calculated from PrimaryGeoTag and ExtClusterGeoTags.

- DNS_ZONE k8gb-test-nonprod.gslb.bcp.absa.co.za
- EDGE_DNS_ZONE: bcp.absa.co.za
- CLUSTER_GEOTAG: 270-nonprod
 will generate **gslb-ns-270-nonprod-k8gb-test-nonprod-gslb.bcp.absa.co.za**

Nsnames are provided by depresolver config via functions `GetPrimaryClusterNsName() string` and `GetExtClusterNsNames() []string` and
are validated during k8gb initialization. Regarding [RFC1035](https://www.ietf.org/rfc/rfc1035.txt), particular
DNS labels are validated on max length 63 characters and whole domain is validated on max length 253 characters
[1](https://stackoverflow.com/questions/32290167/what-is-the-maximum-length-of-a-dns-name). Such behavior is covered by unit tests.

Because nsnames are provided via config, `nsServerName(...)` and `nsServerNameExt(...)` were removed including tests.
Because nsname structure had changed I had to fix `controller_tests`, `values.yaml` and `Makefile`.


controllers/depresolver/depresolver_test.go
```diff
@@ -435,7 +435,7 @@ func TestResolveConfigWithHostnameEdgeDnsServer(t *testing.T) {
 	// arrange
 	defer cleanup()
 	expected := predefinedConfig
-	expected.EdgeDNSServer = "server-nonprod.on.domain.l3.2l.com"
+	expected.EdgeDNSServer = "server-preprod.on.domain.l3.2l.com"
```
ytsarev, line 438: What are the semantics behind the name change?

controllers/depresolver/depresolver_test.go
```diff
@@ -1283,6 +1283,145 @@ func TestResolveLoggerEmptyValues(t *testing.T) {
 	assert.Equal(t, SimpleFormat, config.Log.Format)
 }
 
+func TestNsServerNamesWithMultipleExtClusterGeoTag(t *testing.T) {
+	// arrange
+	defer cleanup()
+	customConfig := predefinedConfig
+	customConfig.DNSZone = "k8gb-test-preprod.gslb.cloud.example.com"
+	customConfig.EdgeDNSZone = "cloud.example.com"
+	customConfig.ClusterGeoTag = "westus-1"
```
ytsarev, line 1292: is it intentionally formed as non-aws looking like tag?

docs/deploy_route53.md
```diff
@@ -71,7 +71,7 @@ kubectl -n test-gslb get gslb test-gslb-failover -o yaml
 aws route53 list-resource-record-sets --hosted-zone-id $YOUR_HOSTED_ZONE_ID
 ```
 
-You should see that `gslb-ns-$dnsZone-$geotag` NS and glue A records were created to
+You should see that `gslb-ns-$geotag-flatten($dnsZone)` NS and glue A records were created to
```
ytsarev, line 74: it's not obvious what flatten means here without the context or example

controllers/depresolver/depresolver_test.go
kuritka: I replaced with standard AWS tags:  https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.RegionsAndAvailabilityZones.html

docs/deploy_route53.md
kuritka: Extending with example:

_You should see new NS and glue A records were created to
automatically configure DNS zone delegation._

_The NS record is calculated from $dnsZone, $edgeDnsZone and $geotag.
For example: `gslb-ns-us-test-gslb.cloud.example.com` will be calculated from 
following values:_
```
$geotag = cloud.example.com
$dnsZone = test.gslb.cloud.example.com
$edgeDnsZone = cloud.example.com
```

ytsarev: Sounds good!

controllers/depresolver/depresolver_test.go
somaritane: In this case, it would look like `us-west-1`?

kuritka: @somaritane  [sure](https://github.com/AbsaOSS/k8gb/pull/459/files#diff-384d90ba8267cb88a0aab2845e0778c734551e380ce8040b033b006584153fdeR1292)

somaritane: IMO It's better if we use vendor-neutral tags in the examples and tests, for example, `location-1`, `location-2`

ytsarev (commented):
@kuritka Is it possible/makes sense we add single terratest around this shortage?

kuritka:
@ytsarev, makes sense, will amend

kuritka:
will reopen when it's ready

PR #460, opened 2021-04-29
reconciliation wait, infoblox logout()

kuritka:
To prevent Infoblox connection will be repeatedly open I updated reconciliation queue.
When error in `CreateZoneDelegationForExternalDNS` occurs, the error is logged, but controller-runtime is
informed to start next loop after `RECONCILE_REQUEUE_SECONDS`. If I don't do it, and Infoblox connection fail, the controller runtime will schedule next loop ASAP.

Second change concerns implementation of standalone `logout()` which is now called after ObjectManager
makes all interaction with Infoblox. For that reason I also encapsulated infoblox client into standalone class.
Now logout is called explicitly in CreateZoneDelegationForExternalDNS and Finalize, see:

```go
objMgr, err := p.infobloxClient.login()
if err != nil {
  return err
}
defer p.infobloxClient.logout()
...  // interaction with Infoblox
```
instead of 
```go
p. infobloxConnection() // defered logout() occured within infobloxConnection()
... // interaction with Infoblox
```

### DIFF
--- a/controllers/gslb_controller.go
+++ b/controllers/gslb_controller.go
@@ -154,13 +154,15 @@ func (r *GslbReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 	// == handle delegated zone in Edge DNS
 	err = r.DNSProvider.CreateZoneDelegationForExternalDNS(gslb)
 	if err != nil {
-		return result.RequeueError(err)
+		log.Err(err).Msg("Create zone delegation")
+		return result.Requeue()
 	}
 
 	// == Status =
 	err = r.updateGslbStatus(gslb)
 	if err != nil {
-		return result.RequeueError(err)
+		log.Err(err).Msg("Update Gslb status")
+		return result.Requeue()
 	}
 
 	// == Finish ==========
--- a/controllers/providers/dns/infoblox-ibclient.go
+++ b/controllers/providers/dns/infoblox-ibclient.go
@@ -21,24 +21,36 @@ import (
 	"fmt"
 	"strconv"
 
+	"github.com/AbsaOSS/k8gb/controllers/depresolver"
+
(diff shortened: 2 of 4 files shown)


ytsarev (commented):
Is it tested end-to-end?

controllers/gslb_controller.go
```diff
@@ -154,13 +154,15 @@ func (r *GslbReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 	// == handle delegated zone in Edge DNS
 	err = r.DNSProvider.CreateZoneDelegationForExternalDNS(gslb)
 	if err != nil {
-		return result.RequeueError(err)
+		log.Err(err).Msgf("Create zone delegation")
```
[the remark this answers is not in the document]
kuritka: something like this:
```
2021-04-29T17:41:31+02:00 ERR main.go:66 > Create zone delegation error="error message aaaaa"
```

Hovewer, the main purpose for error is the fact that based on error != nil,  controller-runtime decides to schedule new loop ASAP or wait for next period. In case of CreateZoneDelegationForExternalDNS we want to wait for the next period. 

Technically, we can skip logging here , because an error is already logged within `CreateZoneDelegationForExternalDNS `.

somaritane, line 157: - `Msg()` can be used instead of `Msgf()` here
- It will look like `Create zone delegation error="error message aaaaa"` only in case of simplified logging format. With json logging format, error context will be provided as separate field, so the message will not look obvious for user in the log or log aggregation solution.
 
```suggestion
		log.Err(err).Msg("Create zone delegation error")
```

kuritka: good point, thx 👍

kuritka:
#462

PR #462, opened 2021-05-03
Don't reconcile immediately when creating zone delegation fails

### LINKED ISSUE #463: Thin down available Infoblox connections
Related to #462 

When zone delegation creation via EdgeDNS provider API fails (error), the reconciliation loop resets immediately.

k8gb is trying to connect but the provider's API can be under a load of waiting connections, so the loop for a single GSLB looks like this:
10s (connecting..., timeout error) + 2s (reconciliation reset) + 10s (connecting..., timeout error) + 2s (reconciliation reset) + 10s + 2s,…. 

Connection is recreated because the reconciliation loop returns an error. If we return nil and wait until the reconciliation period passes, the loop for a single GSLB will look like: 10s (connecting..., timeout error) + RECONCILE_REQUEUE_SECONDS (recover) (30s by default) + 10s (connecting..., timeout error) + RECONCILE_REQUEUE_SECONDS (recover)…CONNECTED

With waiting for the next reconciliation loop instead of the reset loop, the Infoblox will not be under connection load.
There will be more time to share the available connections between all gslbs. In addition, it doesn't produce such amount of timeout errors.

kuritka:
closes #463 

This PR changes the behavior of zone delegation creation check so that when an error happens, the controller is not forced to re-queue reconciliation immediately.

### DIFF
--- a/controllers/gslb_controller.go
+++ b/controllers/gslb_controller.go
@@ -154,7 +154,8 @@ func (r *GslbReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 	// == handle delegated zone in Edge DNS
 	err = r.DNSProvider.CreateZoneDelegationForExternalDNS(gslb)
 	if err != nil {
-		return result.RequeueError(err)
+		log.Err(err).Msg("Unable to create zone delegation")
+		return result.Requeue()
 	}
 
 	// == Status =


controllers/gslb_controller.go
```diff
@@ -154,6 +154,10 @@ func (r *GslbReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 	// == handle delegated zone in Edge DNS
 	err = r.DNSProvider.CreateZoneDelegationForExternalDNS(gslb)
 	if err != nil {
+		if r.Config.EdgeDNSType == depresolver.DNSTypeInfoblox {
```
somaritane, line 157: Do we want provider-specific logic to be in that package?

somaritane:
@kuritka I feel like we're trying to solve the situation with a particular provider and local environment limitations in the upstream project. This might potentially impact users in the future. Probably we should make this change in the reconciliation loop optional for that provider, with the respective option flag. WDYT?

kuritka:
@somaritane , updated. The change is independent on provider.

somaritane (commented):
> @somaritane , updated. The change is independent on provider.

@kuritka cool, let's amend the commit message and PR to make them consistent with changes.

controllers/gslb_controller.go
```diff
@@ -154,7 +154,8 @@ func (r *GslbReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 	// == handle delegated zone in Edge DNS
 	err = r.DNSProvider.CreateZoneDelegationForExternalDNS(gslb)
 	if err != nil {
-		return result.RequeueError(err)
+		log.Err(err).Msg("Create zone delegation")
```
somaritane, line 157:
```suggestion
		log.Err(err).Msg("Unable to create zone delegation")
```

somaritane (changes requested):

controllers/gslb_controller.go
kuritka: @somaritane , message updated

somaritane (approved):

somaritane:
@kuritka LGTM 👍

kuritka:
@ytsarev, does this go to v0.8.0 ?

ytsarev (approved):

ytsarev:
@kuritka sure :)

PR #465, opened 2021-05-03
Flag enabling SplitBrain

kuritka:
**SplitBrainCheck flag**; If false (default) splitBrain is disabled, no metter what EdgeDNSType you use.
To enable SplitBrain, set `SPLIT_BRAIN_CHECK` to true

- I added new field to depresolver Config (SplitBrainCheck) and introduced new env var `SPLIT_BRAIN_CHECK`.
- I updated infoblox provider to skip TXT interractions when !SplitBrainCheck
- Me & k0da made e2e tests against test clusters
- Fix heartbeat issue

### DIFF
--- a/chart/k8gb/templates/operator.yaml
+++ b/chart/k8gb/templates/operator.yaml
@@ -121,3 +121,5 @@ spec:
               value: {{ quote .Values.k8gb.log.level }}
             - name: NO_COLOR
               value: "true"
+            - name: SPLIT_BRAIN_CHECK
+              value: {{ quote .Values.k8gb.splitBrainCheck }}
--- a/controllers/depresolver/depresolver_test.go
+++ b/controllers/depresolver/depresolver_test.go
@@ -48,6 +48,7 @@ var predefinedConfig = Config{
 	EdgeDNSZone:             "8.8.8.8",
 	DNSZone:                 "example.com",
 	K8gbNamespace:           "k8gb",
+	SplitBrainCheck:         true,
 	Infoblox: Infoblox{
 		"Infoblox.host.com",
 		"0.0.3",
@@ -1283,6 +1284,39 @@ func TestResolveLoggerEmptyValues(t *testing.T) {
 	assert.Equal(t, SimpleFormat, config.Log.Format)
 }
 
+func TestResolveConfigSplitBrainCheckEnabled(t *testing.T) {
+	// arrange
+	defer cleanup()
+	expected := predefinedConfig
+	expected.SplitBrainCheck = true
+	// act,assert
+	arrangeVariablesAndAssert(t, expected, assert.NoError)
+}
+
+func TestResolveConfigSplitBrainCheckNotSet(t *testing.T) {
+	// arrange
+	defer cleanup()
+	expected := predefinedConfig
+	expected.SplitBrainCheck = false
+	// act
+	// assert
+	// SplitBrainCheck is not set and expecting to resolve SplitBrainCheck = true and NoError
+	arrangeVariablesAndAssert(t, expected, assert.NoError, SplitBrainCheckKey)
+}
+
+func TestResolveConfigSplitBrainCheckDisabledInvalid(t *testing.T) {
+	// arrange
+	defer cleanup()
+	configureEnvVar(predefinedConfig)
+	_ = os.Setenv(SplitBrainCheckKey, "i.am.wrong??.")
+	resolver := NewDependencyResolver()
+	// act
+	config, err := resolver.ResolveOperatorConfig()
+	// assert
+	assert.NoError(t, err)
+	assert.Equal(t, false, config.SplitBrainCheck)
+}
+
 // arrangeVariablesAndAssert sets string environment variables and asserts `expected` argument with
 // ResolveOperatorConfig() output. The last parameter unsets the values
 func arrangeVariablesAndAssert(t *testing.T, expected Config,
@@ -1306,7 +1340,7 @@ func cleanup() {
 	for _, s := range []string{ReconcileRequeueSecondsKey, ClusterGeoTagKey, ExtClustersGeoTagsKey, EdgeDNSZoneKey, DNSZoneKey, EdgeDNSServerKey,
 		Route53EnabledKey, NS1EnabledKey, InfobloxGridHostKey, InfobloxVersionKey, InfobloxPortKey, InfobloxUsernameKey, InfobloxPasswordKey,
 		OverrideWithFakeDNSKey, OverrideFakeInfobloxKey, K8gbNamespaceKey, CoreDNSExposedKey, InfobloxHTTPRequestTimeoutKey,
-		InfobloxHTTPPoolConnectionsKey, LogLevelKey, LogFormatKey, LogNoColorKey} {
+		InfobloxHTTPPoolConnectionsKey, LogLevelKey, LogFormatKey, LogNoColorKey, SplitBrainCheckKey} {
 		if os.Unsetenv(s) != nil {
 			panic(fmt.Errorf("cleanup %s", s))
 		}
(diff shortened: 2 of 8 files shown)


k0da (dismissed):
LGTM

chart/k8gb/templates/operator.yaml
```diff
@@ -121,3 +121,5 @@ spec:
               value: {{ quote .Values.k8gb.log.level }}
             - name: NO_COLOR
               value: "true"
+            - name: NO_SPLIT_BRAIN
```
somaritane, line 124: Split-brain is the situation that we might want to avoid.  In that case, it makes sense to name the feature flag as `SPLIT_BRAIN_PREVENTION`. It also makes sense to disable the `SPLIT_BRAIN_PREVENTION` flag by default.

controllers/providers/dns/infoblox.go
```diff
@@ -68,7 +68,8 @@ func (p *InfobloxProvider) CreateZoneDelegationForExternalDNS(gslb *k8gbv1beta1.
 		if err != nil {
 			return err
 		}
-		if len(findZone.Ref) > 0 {
+		log.Debug().Msgf("NoSplitBrain: %v", p.config.NoSplitBrain)
```
somaritane, line 71: Imo the message is important enough to be on info log-level
```suggestion
		if p.config.NoSplitBrain {
			log.Info().Msg("Split-brain prevention handling is enabled")
		}
```

controllers/providers/dns/infoblox.go
```diff
@@ -181,3 +165,27 @@ func (p *InfobloxProvider) SaveDNSEndpoint(gslb *k8gbv1beta1.Gslb, i *externaldn
...
+func (p *InfobloxProvider) saveHeartbeatTXTRecord(objMgr *ibclient.ObjectManager, gslb *k8gbv1beta1.Gslb) (err error) {
+	var heartbeatTXTRecord *ibclient.RecordTXT
+	edgeTimestamp := fmt.Sprint(time.Now().UTC().Format("2006-01-02T15:04:05"))
+	heartbeatTXTName := fmt.Sprintf("%s-heartbeat-%s.%s", gslb.Name, p.config.ClusterGeoTag, p.config.EdgeDNSZone)
+	heartbeatTXTRecord, err = objMgr.GetTXTRecord(heartbeatTXTName)
+	if err != nil {
+		return
+	}
+	if heartbeatTXTRecord == nil {
+		log.Info().Msgf("Creating split brain TXT record(%s)...", heartbeatTXTName)
```
somaritane, line 178: Let's add `heartbeatTXTName` variable to the log message context
```suggestion
		log.Info().Str("HeartbeatTXTName", heartbeatTXTName).Msg("Creating split brain TXT record")
```

controllers/providers/dns/infoblox.go
```diff
@@ -181,3 +165,27 @@ func (p *InfobloxProvider) SaveDNSEndpoint(gslb *k8gbv1beta1.Gslb, i *externaldn
...
+		_, err = objMgr.CreateTXTRecord(heartbeatTXTName, edgeTimestamp, gslb.Spec.Strategy.DNSTtlSeconds, "default")
+		if err != nil {
+			return
+		}
+	} else {
+		log.Info().Msgf("Updating split brain TXT record(%s)...", heartbeatTXTName)
```
somaritane, line 184: Let's add `heartbeatTXTName` variable to the log message context
```suggestion
		log.Info().Str("HeartbeatTXTName", heartbeatTXTName).Msg("Updating split brain TXT record")
```

somaritane (changes requested):
@kuritka , looks good, just a few suggestions on logging and feature flag naming

kuritka:
@somaritane @ytsarev @k0da 
  pushed, 
 - e2e tests with @k0da 
 - with k0da we also fixed the heartbeat issue [[1]](https://github.com/AbsaOSS/k8gb/pull/465/files#diff-0ca0bfa319083cd9c6382a7128eaf86aef3dd5241a87938c574b7b343db151e8R82-R93) [[2]](https://github.com/AbsaOSS/k8gb/pull/465/files#diff-0ca0bfa319083cd9c6382a7128eaf86aef3dd5241a87938c574b7b343db151e8R82-R93)

chart/k8gb/templates/operator.yaml
```diff
@@ -121,3 +121,5 @@ spec:
               value: {{ quote .Values.k8gb.log.level }}
             - name: NO_COLOR
               value: "true"
+            - name: SPLIT_BRAIN_PREVENTION
+              value: {{ quote .Values.k8gb.noSplitBrain }}
```
k0da, line 125: Chart value name is contraversial with env.

I suggest: `noSplitBrainDetection`

k0da (changes requested):

chart/k8gb/templates/operator.yaml
k0da: or diableSplitBrainDetection

somaritane: @k0da +1, no sure though why we need to use `no-` or `disable-` prefix for variables, we can simply set the default value to `false`, like already done here: https://github.com/AbsaOSS/k8gb/pull/465/files#diff-b94e4b6c67fca17dce52708f765679ec46a8553531b432d30b8da1b11b1bc89cR84.
I'm also fine with either `splitBrainDetection` or `splitBrainPrevention`, it's just should be named consistently over the codebase.

ytsarev: Can we have just `splitBrain` true/false?

ytsarev (changes requested):
Consider making splitbrain 'enabling' semantics instead of prevention. It will be much more obvious for end user

chart/k8gb/templates/operator.yaml
somaritane: @ytsarev, as I mentioned earlier, split-brain is the name of the issue happening in the distributed system. So if we tell that we want to enable or disable the split-brain - this looks exactly confusing from user point of view

somaritane: It would be more correct to tell that we want to enable or disable the split-brain detection or prevention feature. We don't want to enable the problem, we want to enable the solution.

kuritka:
> Consider making splitbrain 'enabling' semantics instead of prevention. It will be much more obvious for end user

yes I see -  with default value true, but what if value is invalid, also true? disabling semantics is much easier in this.

chart/k8gb/templates/operator.yaml
k0da: splitBrainCheck true/false

kuritka: @somaritane, @k0da  
`SPLIT_BRAIN_CHECK` 
  - default value (empty or not set) false
  - invalid value - false
  - true, t, 1,T ,True, ...  - enables split-brain otherwise it's disabled 

If we do it this way, then enabling split-brain will require to set `SPLIT_BRAIN_CHECK` to true on each k8gb you want to have split-brain, otherwise it will be disabled.

k0da: 👍

ytsarev:
> > Consider making splitbrain 'enabling' semantics instead of prevention. It will be much more obvious for end user
> 
> yes I see - with default value true, but what if value is invalid, also true? disabling semantics is much easier in this.

Ok, https://github.com/AbsaOSS/k8gb/pull/465#discussion_r626320501 proposal looks fine

k0da (approved):
LGTM

controllers/depresolver/depresolver_test.go
```diff
@@ -1283,6 +1284,39 @@ func TestResolveLoggerEmptyValues(t *testing.T) {
...
+	defer cleanup()
+	expected := predefinedConfig
+	expected.SplitBrainCheck = false
+	// act
+	// assert
+	// SplitBrainCheck is not set and expecting to resolve SplitBrainCheck = true and NoError
+	arrangeVariablesAndAssert(t, expected, assert.NoError, SplitBrainCheckKey)
+}
+
+func TestResolveConfigSplitBrainDisabledInvalid(t *testing.T) {
```
somaritane, line 1307: Small suggestion here
```suggestion
func TestResolveConfigSplitBrainCheckDisabledInvalid(t *testing.T) {
```

chart/k8gb/templates/operator.yaml
somaritane: @kuritka looks good 👍  , let's also update the PR title and body for consistency with amendments.

somaritane (approved):
@kuritka LGTM 👍

ytsarev (approved):

PR #470, opened 2021-05-06
Pass endpoint params for ns1 external-dns provider

k0da:

### DIFF
--- a/chart/k8gb/values.yaml
+++ b/chart/k8gb/values.yaml
@@ -62,3 +62,6 @@ route53:
 
 ns1:
   enabled: false
+  # optional custom NS1 API endpoint for on-prem setups
+  # endpoint: https://api.nsone.net/v1/
+  ignoreSSL: false
--- a/docs/examples/ns1/k8gb-cluster-ns1-eu-west-1.yaml
+++ b/docs/examples/ns1/k8gb-cluster-ns1-eu-west-1.yaml
@@ -8,3 +8,4 @@ k8gb:
 
 ns1:
   enabled: true
+  ignoreSSL: false
--- a/docs/examples/ns1/k8gb-cluster-ns1-us-east-1.yaml
+++ b/docs/examples/ns1/k8gb-cluster-ns1-us-east-1.yaml
@@ -8,3 +8,4 @@ k8gb:
 
 ns1:
   enabled: true
+  ignoreSSL: false
--- a/chart/k8gb/templates/external-dns/external-dns-ns1.yaml
+++ b/chart/k8gb/templates/external-dns/external-dns-ns1.yaml
@@ -27,6 +27,12 @@ spec:
         - --domain-filter={{ .Values.k8gb.edgeDNSZone }} # will make ExternalDNS see only the hosted zones matching provided domain, omit to process all available hosted zones
         - --annotation-filter=k8gb.absa.oss/dnstype=ns1 # filter out only relevant DNSEntrypoints
         - --provider=ns1
+{{- if .Values.ns1.endpoint }}
+        - --ns1-endpoint={{ .Values.ns1.endpoint }}
+{{- end }}
+{{- if .Values.ns1.ignoreSSL }}
+        - --ns1-ignoressl={{ .Values.ns1.ignoreSSL }}
+{{- end }}
         - --txt-owner-id=k8gb-{{ .Values.k8gb.dnsZone }}-{{ .Values.k8gb.clusterGeoTag }}
         - --policy=sync # enable full synchronization including record removal
         - --log-level=debug # debug only


docs/examples/ns1/k8gb-cluster-ns1-us-east-1.yaml
```diff
@@ -8,3 +8,5 @@ k8gb:
 
 ns1:
   enabled: true
+  # endpoint: https://ns1.example.com
+  # ignoreSSL: false
```
ytsarev, line 12: `ignoreSSL` can stay `false` uncommented

chart/k8gb/values.yaml
```diff
@@ -62,3 +62,5 @@ route53:
 
 ns1:
   enabled: false
+  # endpoint: http://ns1.example.com
```
ytsarev, line 65: please mentioned in the comment that the value is optional and it defaults to managed NS1

ytsarev (changes requested):

docs/examples/ns1/k8gb-cluster-ns1-us-east-1.yaml
k0da: This way it will be set into deployment, which makes args list longer and hard to read. I'd leave defaults into provider's default settings

chart/k8gb/values.yaml
k0da: @ytsarev done

docs/examples/ns1/k8gb-cluster-ns1-us-east-1.yaml
ytsarev: No, in case of `false` the `if` in the template will not add the parameter at all

k0da: @ytsarev amended

somaritane:
@k0da , @ytsarev, do we want to also propagate MinTTLSeconds ?

docs/examples/ns1/k8gb-cluster-ns1-us-east-1.yaml
ytsarev: @k0da but apparently not pushed?

ytsarev (changes requested):

docs/examples/ns1/k8gb-cluster-ns1-us-east-1.yaml
k0da: Sorry, missed that you referenced example values, while I've been amending default chart values. Fixed now

k0da:
> @k0da , @ytsarev, do we want to also propagate MinTTLSeconds? Or are we OK with default TTL?

Do we have need for that atm?

ytsarev:
> > @k0da , @ytsarev, do we want to also propagate MinTTLSeconds? Or are we OK with default TTL?
> 
> Do we have need for that atm?

depends on the default. If it is higher than our 30sec we might need to override it. Overall, the collision with DNSEndpoint CR does not look convenient. I am not sure about behavior here at all btw - will minTTL override specific CR TTL value? Does it have precedence? We need to figure out. Anyway, it does not block this specific PR

docs/examples/ns1/k8gb-cluster-ns1-eu-west-1.yaml
```diff
@@ -8,3 +8,5 @@ k8gb:
 
 ns1:
   enabled: true
+  # endpoint: https://ns1.example.com
```
ytsarev, line 11: why do we need it in this sample?

docs/examples/ns1/k8gb-cluster-ns1-us-east-1.yaml
```diff
@@ -8,3 +8,5 @@ k8gb:
 
 ns1:
   enabled: true
+  # endpoint: https://ns1.example.com
```
ytsarev, line 11: same as above

chart/k8gb/values.yaml
```diff
@@ -62,3 +62,6 @@ route53:
 
 ns1:
   enabled: false
+  # optional custom NS1 API endpoint for on-prem setups
+  # endpoint: http://ns1.example.com
```
somaritane, line 66: @k0da IIRC, NS1 endpoint setting should not contain URI scheme
```suggestion
  # optional custom NS1 API endpoint, defaults to managed NS1 (api.nsone.net)
  # endpoint: ns1.example.com
```

somaritane (changes requested):

chart/k8gb/values.yaml
k0da: app.Flag("ns1-endpoint", "When using the NS1 provider, specify the URL of the API endpoint to target (default: https://api.nsone.net/v1/)").Default(defaultConfig.NS1Endpoint).StringVar(&cfg.NS1Endpoint)

docs/examples/ns1/k8gb-cluster-ns1-us-east-1.yaml
k0da: dropped

docs/examples/ns1/k8gb-cluster-ns1-eu-west-1.yaml
k0da: same as below

chart/k8gb/values.yaml
somaritane: Yep, you're right, just checked in NS1 Golang SDK https://github.com/ns1/ns1-go/blob/v2/rest/client.go#L40, should be full URL.
Anyway, still makes sense to point to the default value

k0da: yes, thought the same, amended

somaritane: cool 👍

somaritane (approved):
LGTM 👍

ytsarev (approved):

PR #475, opened 2021-05-10
Add Jekyll local authoring setup for GitHub pages

somaritane:
Add Jekyll local authoring setup for GitHub pages

Setup instructions for Jekyll and prerequisites will be a part of CONTRIBUTING.md documentation on main branch (https://github.com/AbsaOSS/k8gb/pull/477).

Signed-off-by: Timofey Ilinykh <timofey.ilinykh@absa.africa>


kuritka (approved):

k0da:
How is this intended to be run? Through local jekyll? Can docker image be used instead?

somaritane:
> How is this intended to be run? Through local jekyll? Can docker image be used instead?

Good point, k0da!
For now, [this CONTRIBUTE.md update](https://github.com/AbsaOSS/k8gb/pull/477) refers to local Jekyll installation and execution.
It definitely would be better to investigate running via docker image to avoid installing unnecessary local dependencies.
This setup will however work in both cases.

PR #481, opened 2021-05-12
Unify external-dns deployment

k0da:
Use single template for Route53 and NS1 external-dns deployments


ytsarev (approved):
LGTM, also was e2e tested by @k0da

PR #484, opened 2021-05-12
Fix chart repo url after org move

k0da:


k0da:
@ytsarev coredns-helm wasn't moved

ytsarev:
@k0da should it be moved then?

k0da:
Not sure we should do this during release :D

ytsarev:
fair enough :)

ytsarev (approved):

PR #492, opened 2021-05-19
Refactor test target

kuritka:
`test` target runs unit tests only while `check` target runs golic + lint + test

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -77,6 +77,12 @@ performs upgrade of k8gb helm chart and controller to the testing version built 
   These tests are updated only if the change is substantial enough to affect the main end-to-end flow.
 - See the [local playground guide](https://github.com/k8gb-io/k8gb/blob/master/docs/local.md) for local testing environment setup and integration test execution.
 
+### Unit testing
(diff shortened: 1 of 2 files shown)


ytsarev (commented):
In which scenarios `make check`  is going to be executed? Doesn't look obvious to me.

kuritka:
> In which scenarios `make check` is going to be executed? Doesn't look obvious to me.

It's implementation of yesterday's discussion with @somaritane, where `make test` would test doing no additional steps, while `check` is doing necessary checks. 

`make check` is an optional step but would executed before pushing to GitHub by developer. If `make check` is not executed and anything wrong, build pipeline will reveal the issue.

somaritane (dismissed):
lgtm

ytsarev (changes requested):
Ok, then please make it visible in `make help` and probably mention it in `CONTRIBUTING.md` ?

kuritka:
> Ok, then please make it visible in `make help` and probably mention it in `CONTRIBUTING.md` ?

@ytsarev, thx amended

k0da (approved):
LGTM

ytsarev (approved):

PR #495, opened 2021-05-20
Add override for dev env variables with "dotenv" file

somaritane:
This PR adds the possibility for persistent override of local dev environment variables with ".env" file.

A contributor can now store the environment variables with respective values in the `.env` file at the local repository root:
```sh
cat .env

# .env:
LOG_LEVEL=info
LOG_FORMAT=json
```
Overrides done this way can persist between terminal sessions and can be used as a single point of configuration for development in terminal and IDE of choice, so they don't need to set environment variables separately for terminal and IDE, the experience remains consistent.

Signed-off-by: Timofey Ilinykh <timofey.ilinykh@absa.africa>


Makefile
```diff
@@ -418,12 +426,6 @@ define testapp-set-replicas
 	kubectl scale deployment frontend-podinfo -n test-gslb --replicas=$1
 endef
 
-define hit-testapp-host
-	kubectl run -it --rm busybox --restart=Never --image=busybox -- sh -c \
-	"echo 'nameserver `$(K8GB_COREDNS_IP)`' > /etc/resolv.conf && \
-	wget -qO - $1"
-endef
```
ytsarev, line 425: Could you please clarify this removal?

somaritane: @ytsarev Sure, this is the removal of the duplicate definition to https://github.com/k8gb-io/k8gb/pull/495/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L403 I've noticed while updating the Makefile.

somaritane: I can make this fix as separate PR to split the scope.

somaritane: @ytsarev will be done separately, amended the changes.

kuritka (approved):

PR #491, opened 2021-05-18
refactor: interface rename

kuritka:
In order to https://golang.org/doc/effective_go#interface-names I'm renaming interfaces
IDnsProvider -> Provider
IAssistant -> Assistant

Signed-off-by: kuritka <kuritka@gmail.com>

PR #503, opened 2021-05-24
refactor (1/3): simplify controller tests

kuritka:
- I'm removing config settings through environment variables in controller_tests.
Depresolver tests already cover such functionality.  Thanks to that I'm simplifying `provideSettings` function. 
Also, introducing new property will not require setting environment variables
in two places.

- In order to rename interfaces (#491) I'm regenerating mocks to get new names.


ytsarev (commented):
Does it mean we don't set config through environment variables? How do we test the expected env var name?

kuritka:
@ytsarev yes and no. We resolve env vars in [main() func ](https://github.com/k8gb-io/k8gb/blob/master/main.go#L62) while controller (or better reconciler) [accepts ready-made configuration](https://github.com/k8gb-io/k8gb/blob/master/main.go#L100). It still happens in main() func.   

The same we did in controller_tests. Set envvars, read them, pass to reconciler, factory, etc. At the end, the test started. From my POV we can pass configuration directly to controller with no envvars manipulation.  

Envvars manipulation itself is well tested in depresolver tests.

ytsarev (approved):
`Got it, I like the separation

PR #505, opened 2021-05-27
refactor (2/3): Remove responsibility for target DNS from GSLB assistant

kuritka:
GSLB assistant with help of `OVERRIDE_WITH_FAKE_EXT_DNS` decides whether to use mock DNS (localhost:7753) or calculated NSServerName (gslb-ns-uk-cloud.example.com:53).

This PR removes such functionality from gslb assistant. Depresolver `getNSName` returns NSName based on EdgeDNSServer.
If edgeDNSServer is localhost than nsname is localhost for any region, otherwise proper nsname is returned.
I also removed `OVERRIDE_WITH_FAKE_EXT_DNS` and introduced `EDGE_DNS_SERVER_PORT`.
Now we can request edge DNS server on any port. All current settings are backward compatible as default 53 port is used.

Because OverrideWithFakeDNS was removed I had to change Assistant interface contract and regenerate mocks, covered by tests.
 

### DIFF
--- a/controllers/depresolver/depresolver_test.go
+++ b/controllers/depresolver/depresolver_test.go
@@ -43,9 +43,10 @@ import (
 var predefinedConfig = Config{
 	ReconcileRequeueSeconds: 30,
 	ClusterGeoTag:           "us",
-	ExtClustersGeoTags:      []string{"uk", "eu"},
+	ExtClustersGeoTags:      []string{"za", "eu"},
 	EdgeDNSType:             DNSTypeInfoblox,
-	EdgeDNSServer:           "8.8.8.8",
+	EdgeDNSServer:           "dns.cloud.example.com",
+	EdgeDNSServerPort:       53,
 	EdgeDNSZone:             "example.com",
 	DNSZone:                 "cloud.example.com",
 	K8gbNamespace:           "k8gb",
@@ -60,7 +61,6 @@ var predefinedConfig = Config{
 		11,
 	},
 	Override: Override{
-		false,
 		false,
 	},
 	Log: Log{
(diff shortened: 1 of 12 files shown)


controllers/depresolver/depresolver_test.go
```diff
@@ -1390,6 +1415,23 @@ func TestNsServerNamesWithMultipleExtClusterGeoTag(t *testing.T) {
...
+		customConfig := predefinedConfig
+		customConfig.EdgeDNSServer = edgeDNSServer
+		configureEnvVar(customConfig)
+		resolver := NewDependencyResolver()
+		// act
+		config, err := resolver.ResolveOperatorConfig()
+		// assert
+		assert.NoError(t, err)
+		assert.Equal(t, config.GetClusterNSName(), edgeDNSServer)
+		assert.True(t, reflect.DeepEqual(config.GetExternalClusterNSNames(), map[string]string{"uk": edgeDNSServer, "eu": edgeDNSServer}))
```
ytsarev, line 1431: `"uk"` looks unexpected. Typo?

ytsarev: https://github.com/k8gb-io/k8gb/pull/505/files#diff-384d90ba8267cb88a0aab2845e0778c734551e380ce8040b033b006584153fdeR46 - hm. I didn't realize we check to Brexit in our tests :D

controllers/gslb_controller_test.go
```diff
@@ -69,6 +69,7 @@ var predefinedConfig = depresolver.Config{
 	ClusterGeoTag:           "us-west-1",
 	ExtClustersGeoTags:      []string{"us-east-1"},
 	EdgeDNSServer:           "8.8.8.8",
```
ytsarev, line 71: Shouldn't we make `localhost` here? Or use `53` and override `EdgeDNSServerPort` in specific tests?
Otherwise `8.8.8.8` with `7753` is actually never working combination.

ytsarev (commented):
Great improvement, just added couple of comments

controllers/gslb_controller_test.go
kuritka: @ytsarev , thx, amended!

regions:
uk -> za

EdgeDNServer I put there more reasonable domain name.  Tests for localhost EdgeDNSServer are already implemented:  https://github.com/k8gb-io/k8gb/pull/505/files#diff-384d90ba8267cb88a0aab2845e0778c734551e380ce8040b033b006584153fdeR1418

ytsarev (approved):
cool stuff, LGTM

PR #507, opened 2021-05-31
refactor (3/3): Introducing local FakeDNS

kuritka:
I removed global fake DNS from controller and replaced it by local DNS server from utils.
I fixed controller unit tests to be interacting with local FakeDNS instances.


controllers/internal/utils/fakedns.go
```diff
@@ -0,0 +1,172 @@
...
+func (m *DNSMock) handleReflect(w dns.ResponseWriter, r *dns.Msg) {
+	msg := new(dns.Msg)
+	msg.SetReply(r)
+	msg.Compress = false
+	if m.records[r.Question[0].Qtype] != nil {
+		for _, rr := range m.records[r.Question[0].Qtype] {
+			fqdn := strings.Split(rr.String(), "\t")[0]
+			if fqdn == r.Question[0].Name {
+				msg.Answer = append(msg.Answer, rr)
+				// msg.Extra = append(msg.Extra, rr)
```
ytsarev, line 167: Reserved for the future?

ytsarev (commented):
Just one comment above.

Overall looks really clean and manageable. Is it a candidate for https://github.com/AbsaOSS/gopkg/ in the future? Can be reused outside of k8gb

kuritka:
@ytsarev , amended.

💯  it will go to gopkg. The idea is to make it run for a while in k8gb. Maybe some changes will need to be implemented and when it settles down a bit, we'll move it into gopkg.

ytsarev (approved):

PR #516, opened 2021-06-08
Use edgeDNSServer for NS name resolution

ytsarev:
* Use edgeDNSServer for NS name resolution with the fallback to the local nameserver
* Supporting dns related functions
* Logs to stdout during the tests
* Use localhost as default edgeDNSServer to avoid network timeout

### DIFF
--- a/controllers/providers/assistant/gslb.go
+++ b/controllers/providers/assistant/gslb.go
@@ -221,16 +221,16 @@ func (r *GslbLoggerAssistant) InspectTXTThreshold(fqdn string, edgeDNSServerPort
 	return errors.NewResourceExpired(fmt.Sprintf("Can't find split brain TXT record at EdgeDNS server(%s) and record %s ", ns, fqdn))
 }
 
-func GetARecords(msg *dns.Msg) []string {
+func getARecords(msg *dns.Msg) []string {
 	var ARecords []string
 	for _, nsA := range msg.Answer {
-		IP := strings.Split(nsA.String(), "\t")[4]
-		ARecords = append(ARecords, IP)
+		ip := nsA.(*dns.A).A.String()
+		ARecords = append(ARecords, ip)
 	}
 	return ARecords
 }
 
-func DNSQuery(fqdn string, nameserver string, nameserverport int) (*dns.Msg, error) {
+func dnsQuery(fqdn string, nameserver string, nameserverport int) (*dns.Msg, error) {
 	dnsMsg := new(dns.Msg)
 	edgeDNSServer := fmt.Sprintf("%s:%v", nameserver, nameserverport)
 	fqdn = fmt.Sprintf("%s.", fqdn) // Convert to true FQDN with dot at the end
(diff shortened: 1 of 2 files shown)


controllers/providers/assistant/gslb.go
```diff
@@ -221,25 +221,50 @@ func (r *GslbLoggerAssistant) InspectTXTThreshold(fqdn string, edgeDNSServerPort
 	return errors.NewResourceExpired(fmt.Sprintf("Can't find split brain TXT record at EdgeDNS server(%s) and record %s ", ns, fqdn))
 }
 
+func GetARecords(msg *dns.Msg) []string {
```
kuritka, line 224: keep it private or consider to expose via interface `Assistant`

controllers/providers/assistant/gslb.go
```diff
@@ -221,25 +221,50 @@ func (r *GslbLoggerAssistant) InspectTXTThreshold(fqdn string, edgeDNSServerPort
...
+	var ARecords []string
+	for _, nsA := range msg.Answer {
+		IP := strings.Split(nsA.String(), "\t")[4]
+		ARecords = append(ARecords, IP)
+	}
+	return ARecords
+}
+
+func DNSQuery(fqdn string, nameserver string, nameserverport int) (*dns.Msg, error) {
```
kuritka, line 233: keep it private or consider to expose via interface `Assistant`

controllers/providers/assistant/gslb.go
```diff
@@ -221,25 +221,50 @@ func (r *GslbLoggerAssistant) InspectTXTThreshold(fqdn string, edgeDNSServerPort
 	return errors.NewResourceExpired(fmt.Sprintf("Can't find split brain TXT record at EdgeDNS server(%s) and record %s ", ns, fqdn))
 }
 
+func GetARecords(msg *dns.Msg) []string {
+	var ARecords []string
+	for _, nsA := range msg.Answer {
+		IP := strings.Split(nsA.String(), "\t")[4]
+		ARecords = append(ARecords, IP)
```
kuritka, line 228: consider this:
```suggestion
		ip := nsA.(*dns.A).A.String()
		ARecords = append(ARecords, ip)
```

controllers/providers/assistant/gslb.go
```diff
@@ -221,25 +221,50 @@ func (r *GslbLoggerAssistant) InspectTXTThreshold(fqdn string, edgeDNSServerPort
...
+	}
+	return ARecords
+}
+
+func DNSQuery(fqdn string, nameserver string, nameserverport int) (*dns.Msg, error) {
+	dnsMsg := new(dns.Msg)
+	edgeDNSServer := fmt.Sprintf("%s:%v", nameserver, nameserverport)
+	fqdn = fmt.Sprintf("%s.", fqdn) // Convert to true FQDN with dot at the end
```
kuritka, line 236: should be the argument name fqdn, when it bacomes fqdn here ^^?

kuritka (changes requested):
thx for fix tests, I have some points to consider

controllers/providers/assistant/gslb.go
ytsarev: Not sure if I got the questions right but we need to add dot(.) to the end of the argument to satisfy miekgdns expectations

controllers/providers/assistant/gslb.go
ytsarev: wasn't aware it is available this way, will try it out 👍

ytsarev:
@kuritka thanks for the careful review, amended, please double check

controllers/providers/assistant/gslb.go
kuritka: I mean, the argument accepting by DNSQuery is not fqdn. We are making it fqdn.  So the point is, to rename input argument to e.g. `host` ?

ytsarev: ah, ok, makes sense

ytsarev: @kuritka amended

kuritka (approved):

PR #519, opened 2021-06-09
Scan k8gb image by Artifacthub.io

k0da:

### DIFF
--- a/chart/k8gb/Chart.yaml
+++ b/chart/k8gb/Chart.yaml
@@ -24,6 +24,9 @@ keywords:
 annotations:
   artifacthub.io/operator: "true"
   artifacthub.io/operatorCapabilities: Seamless Upgrades
+  artifacthub.io/images: |
+    - name: k8gb
+      image: absaoss/k8gb
   artifacthub.io/crds: |
     - kind: Gslb
       version: v1beta1


chart/k8gb/Chart.yaml
```diff
@@ -24,6 +24,9 @@ keywords:
 annotations:
   artifacthub.io/operator: "true"
   artifacthub.io/operatorCapabilities: Seamless Upgrades
+  artifacthub.io/images: |
+    - name: k8gb
+      image: absaoss/k8gb
```
ytsarev, line 29: meaning it will scan all the tags?

k0da: I guess. There is no documentation about this, and `latest` is not scanned

ytsarev (approved):
ok, let's try it out

chart/k8gb/Chart.yaml
tegioz: Hi 👋 

This is Sergio, from Artifact Hub.

Artifact Hub is now able to identify containers images automatically in many cases. This feature was released late May, so any new version pushed after that date may get its images automatically detected. You can still define them manually if you prefer though. Only the provided tag will be scanned (you can use `latest` if you wish).

k0da: @tegioz Yeah, I'm trying to understand/workaround issue we have atm with https://artifacthub.io/packages/helm/k8gb/k8gb . It wasn't scanned automatically, nor with applied annotation.

tegioz: It looks like no new versions have been released since May 13. The automatic detection feature was released a couple of weeks after that. Regarding the annotation, I can't see it in any version released in the `index.yaml` file (https://www.k8gb.io/index.yaml), and the most recent version in the index is the same that Artifact Hub displays, `0.8.0`.

Please let me know if I can be of any help 🙂

k0da: @tegioz thanks, for calrification. Will check again once we release a new version.

PR #524, opened 2021-06-15
Revert "Scan k8gb image by Artifacthub.io (#519)"

k0da:
This reverts commit f6766a06b0107c5ae453405d3bd44119f6931427.

### DIFF
--- a/chart/k8gb/Chart.yaml
+++ b/chart/k8gb/Chart.yaml
@@ -24,9 +24,6 @@ keywords:
 annotations:
   artifacthub.io/operator: "true"
   artifacthub.io/operatorCapabilities: Seamless Upgrades
-  artifacthub.io/images: |
-    - name: k8gb
-      image: absaoss/k8gb
   artifacthub.io/crds: |
     - kind: Gslb
       version: v1beta1


ytsarev:
@k0da looks like DCO doesn't like pure git reverts without a Sign-off

k0da:
fxd

ytsarev (approved):

k0da:
@tegioz we've reverted Chart.yaml annotation assuming image detection would work, but now we got spammed with messages like:
```* unable to inspect the image (index.docker.io/absaoss/k8gb:latest: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?

* unable to initialize Podman client: no podman socket found: stat podman/podman.sock: no such file or directory

* GET https://index.docker.io/v2/absaoss/k8gb/manifests/latest MANIFEST_UNKNOWN: manifest unknown; map[Tag:latest]
```

Is there a channel we could follow to track artifacthub issues/announcements? 
Thanks,

tegioz:
Hi @k0da 

It looks like the images annotation is still there, it's visible in your `index.yaml` file (https://www.k8gb.io/index.yaml) for version `0.8.1`. Maybe the reverted change hasn't been published yet? The image you added to the annotation has been added to your package just fine, you can see it here: https://artifacthub.io/packages/helm/k8gb/k8gb.

Screen Shot 2021-06-17 at 08 13 36

Regarding the errors you're getting, the problem seems to be that the `latest` tag for that image does not exist in the registry. You could push that tag, but I'd recommend you to provide the tag corresponding to the package version if possible:

In your `index.yaml`, you have:

```yaml
artifacthub.io/images: |
  - name: k8gb
    image: absaoss/k8gb
```

and maybe you could do something like this (if you want to provide them manually):

```yaml
artifacthub.io/images: |
  - name: k8gb
    image: absaoss/k8gb:v0.8.1
```

If the version tag is used when rendering the manifests from the Helm chart, the image detection should pick it up and you don't need to worry about handling this manually.

To track Artifact Hub announcements you can:

- Subscribe to the new release notifications in Artifact Hub itself (bell icon on the top right). This way we'll send you an email with the corresponding portion of the [changelog](https://artifacthub.io/packages/helm/artifact-hub/artifact-hub?modal=changelog) every time a new version is released. This will give you the most details, but only on every new release.
- Join the #artifact-hub channel in the CNCF Slack.
- Follow us on twitter (@cncfartifacthub).

Hope this helps 👋

ytsarev:
@tegioz thanks a lot for the detailed clarification, I really appreciate it

@k0da looks like current index.yaml update is depending on the release process/helm publish pipeline https://github.com/k8gb-io/k8gb/commits/gh-pages/index.yaml

PR #525, opened 2021-06-15
Metrics configuration #(1/4)

kuritka:
Related to #124

I created a MetricsAddress field.  The default value is “0.0.0.0:8080" and defines where Prometheus metrics endpoint is listening. (see: https://docs.openshift.com/container-platform/4.1/applications/operator_sdk/osdk-monitoring-prometheus.html#osdk-monitoring-prometheus-metrics-helper_osdk-monitoring-prometheus)

I also removed flag.StringVar `"metrics-addr", ":8080”,` from `main.go` as moving inputs into depresolver and also, removed `enableLeaderElection` and overall flags completely.

### DIFF
--- a/controllers/depresolver/depresolver_config.go
+++ b/controllers/depresolver/depresolver_config.go
@@ -19,6 +19,7 @@ package depresolver
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/AbsaOSS/gopkg/env"
@@ -51,6 +52,7 @@ const (
 	LogFormatKey                   = "LOG_FORMAT"
 	LogNoColorKey                  = "NO_COLOR"
 	SplitBrainCheckKey             = "SPLIT_BRAIN_CHECK"
+	MetricsAddressKey              = "METRICS_ADDRESS"
 )
 
 // ResolveOperatorConfig executes once. It reads operator's configuration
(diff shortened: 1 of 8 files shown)


controllers/depresolver/depresolver_config.go
```diff
@@ -193,6 +195,10 @@ func (dr *DependencyResolver) validateConfig(config *Config, recognizedDNSTypes
 			return fmt.Errorf("error for geo tag: %s. %s in ns name %s", geoTag, err, nsName)
 		}
 	}
+	err = field(PrometheusAddressPortKey, config.Prometheus.Port).isHigherThanZero().isLessOrEqualTo(65535).isHigherThan(1024).err
```
somaritane, line 198: `.isHigherThanZero()` is within `.isHigherThan(1024)` rule

kuritka: @somaritane, good point. Will refactor....

k0da:
@kuritka this is a bit confusing. 9090 is a default p8s server port. 8080 and MetricsAddr was pretty self explanatory term.

kuritka:
> @kuritka this is a bit confusing. 9090 is a default p8s server port. 8080 and MetricsAddr was pretty self explanatory term.

Ok, will change to 8080. What do you mean by MetricsAddr ?

kuritka:
> @kuritka you have it now as Prometheus.Port (sounds like server port), MetricsAddr defines metrics endpoint better IMO

@k0da , good point 👍 . I refactored config and tests. Now everything seems to be ok.

controllers/depresolver/depresolver_config.go
```diff
@@ -51,6 +52,7 @@ const (
 	LogFormatKey                   = "LOG_FORMAT"
 	LogNoColorKey                  = "NO_COLOR"
 	SplitBrainCheckKey             = "SPLIT_BRAIN_CHECK"
+	MetricsAddressKey              = "METRICS_ADDRESS"
```
somaritane, line 55: @kuritka this env variable should be wired in `operator.yaml` helm template too. Or is k8gb helm chart out of the scope of this change?

k0da (dismissed):
LGTM

main.go
```diff
@@ -75,9 +66,9 @@ func main() {
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:             runtimescheme,
-		MetricsBindAddress: metricsAddr,
+		MetricsBindAddress: config.MetricsAddress,
 		Port:               9443,
-		LeaderElection:     enableLeaderElection,
+		LeaderElection:     false,
```
ytsarev, line 71: hmm, we never tested it, but what if somebody would like to try to execute multiple k8gb controllers with leader election mechanism in the future? currently it is optional but here it is becoming hardcoded.

kuritka: enableLeaderElection was accessible through an argument of executable. Making such functionality available would mean passing the argument to the controller pod. Till now it was always false anyway.

What we can do is create an issue, expose an environment variable and test.

ytsarev: @kuritka ok let's log the issue then

k0da: even more, I think it requires to have locking to be implemented.

kuritka: good point @k0da. Feel free and add your comments by #526 please.

controllers/depresolver/depresolver_config.go
kuritka: good point  @somaritane 👍 , amended

ytsarev (approved):
lgtm

PR #529, opened 2021-06-18
Install prometheus on local clusters (2/4)

kuritka:
related to #124

I want to check metrics that will be implemented. That's why I'm adding the option to install prometheus on both clusters with `make deploy-prometheus`.

Such deployed prometheus scrapes metrics of k8gb operator and is accessible directly from the browser:

 - http://127.0.0.1:8080 - test-gslb1
 - http://127.0.0.1:8081 - test-gslb2

Co-authored-by: Timofey Ilinykh <timofey.ilinykh@absa.africa>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -185,6 +185,16 @@ destroy-full-local-setup: ## Destroy full local multicluster setup
 	k3d cluster delete $(CLUSTER_GSLB2)
 	docker network rm $(CLUSTER_GSLB_NETWORK)
 
+.PHONY: deploy-prometheus
+deploy-prometheus:
+	$(call deploy-prometheus,$(CLUSTER_GSLB1))
+	$(call deploy-prometheus,$(CLUSTER_GSLB2))
+
+.PHONY: uninstall-prometheus
+uninstall-prometheus:
+	$(call uninstall-prometheus,$(CLUSTER_GSLB1))
+	$(call uninstall-prometheus,$(CLUSTER_GSLB2))
+
 .PHONY: dns-tools
 dns-tools: ## Run temporary dnstools pod for debugging DNS issues
 	@kubectl -n k8gb get svc k8gb-coredns
@@ -486,4 +496,29 @@ endef
 define list-running-pods
 	@echo "\n$(YELLOW)Local cluster $(CYAN)$1 $(NC)"
 	kubectl get pods -A --context=k3d-$1
+endef
+
+define deploy-prometheus
+	@echo "\n$(YELLOW)Local cluster $(CYAN)$1$(NC)"
+
+	@echo "\n$(YELLOW)Set annotations on pods that will be scraped by prometheus$(NC)"
+	kubectl annotate pods -l name=k8gb -n k8gb --overwrite prometheus.io/scrape="true" --context=k3d-$1
+	kubectl annotate pods -l name=k8gb -n k8gb --overwrite prometheus.io/port="8080" --context=k3d-$1
+
+	@echo "\n$(YELLOW)install prometheus $(NC)"
+	helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
+	helm repo update
+	helm -n k8gb upgrade -i prometheus prometheus-community/prometheus -f deploy/prometheus/values.yaml \
+		--version 14.2.0 \
+		--wait --timeout=2m0s \
+		--kube-context=k3d-$1
+endef
+
+define uninstall-prometheus
+	@echo "\n$(YELLOW)Local cluster $(CYAN)$1$(NC)"
+
+	@echo "\n$(YELLOW)uninstall prometheus $(NC)"
+	helm uninstall prometheus -n k8gb --kube-context=k3d-$1
+	kubectl annotate pods -l name=k8gb -n k8gb prometheus.io/scrape- --context=k3d-$1
+	kubectl annotate pods -l name=k8gb -n k8gb prometheus.io/port- --context=k3d-$1
 endef
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,6 +12,7 @@
 - [Testing](#testing)
   - [Testing against real k8s clusters](#testing-against-real-k8s-clusters)
 - [Debugging](#debugging)
+- [Metrics](#metrics)
 - [Code style](#code-style)
 - [Commit and Pull Request message](#commit-and-pull-request-message)
   - [Signature](#signature)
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -487,3 +492,22 @@ define list-running-pods
...
 	kubectl get pods -A --context=k3d-$1
 endef
+
+define deploy-prometheus
+	@echo "\n$(YELLOW)Local cluster $(CYAN)$1$(NC)"
+	kubectl config use-context k3d-$1
+
+	@echo "\n$(YELLOW)Set annotations on pods that will be scraped by prometheus$(NC)"
+	kubectl annotate pods -l name=k8gb -n k8gb --overwrite prometheus.io/scrape="true"
+	kubectl annotate pods -l name=k8gb -n k8gb --overwrite prometheus.io/path=/metrics
```
k0da, line 502: prometheus.io/path: If the metrics path is not /metrics, define it with this annotation.

Makefile
```diff
@@ -487,3 +492,22 @@ define list-running-pods
...
+	kubectl annotate pods -l name=k8gb -n k8gb --overwrite prometheus.io/port="8080"
+
+	@echo "\n$(YELLOW)install prometheus $(NC)"
+	helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
+	helm repo add kube-state-metrics https://kubernetes.github.io/kube-state-metrics
```
k0da, line 507: Why to add repo which is not used?

k0da (changes requested):

Makefile
kuritka: @k0da , I was following recommended way: https://artifacthub.io/packages/helm/prometheus-community/prometheus#scraping-pod-metrics-via-annotations

Makefile
kuritka: amended.

Makefile
k0da: This is the same as default, it not worth to annotate it explicitly

somaritane: @kuritka , there's no need to install full-fledged Prometheus if you want to check or test the output of the metrics endpoint. All you have to do is to set up port-forwarding to the k8gb controller pod with `kubectl`
```
kubectl port-forward -n k8gb pod/k8gb-XXXXXXXX-YYYYY 8080:8080
curl localhost:8080/metrics
```
(You can also temporarily expose the endpoint using Service)

kuritka: @k0da , amended

k0da (approved):
LGTM

Makefile
kuritka: @sommaritane, thanks. `make deploy-prometheus` and then have an accessible web UI without any further steps was the intention. I will definitely test the automatic way. Still, it is better (and not only for the scope implementation of this task) that the k8gb developer has the possibility to quickly check the metrics in prometheus.

ytsarev (commented):
@kuritka is it intentionally without prometheus operator? just long-term thinking of ServiceMonitor CRDs

kuritka:
@ytsarev - no operator was the intention, because we are on two local test-clusters. I didn't want to deliver something too extensive , just to check the metrics on the k8gb controller. But we can extend it at any time.

k0da:
I'm against pulling p8s operator here, we have single app in a single ns, we can simply maintain rule in a config map. We're not scalling testing clusters.

ytsarev:
@kuritka so it is for querying metrics with prom UI, correct?

@k0da more context: https://docs.openshift.com/container-platform/4.1/applications/operator_sdk/osdk-monitoring-prometheus.html. Not a fond of pulling either but discussing what's the most relevant

main.go
```diff
@@ -102,7 +102,7 @@ func main() {
 	}
 	reconciler.DNSProvider = f.Provider()
 	log.Info().Msgf("provider: %s", reconciler.DNSProvider)
-	log.Info().Msg("starting metrics")
+	log.Info().Msgf("starting metrics on %s", reconciler.Config.MetricsAddress)
```
somaritane, line 105: Consider making this change as separate PR, as it's touching the controller code and not being exactly the scope of this PR.

Makefile
```diff
@@ -185,6 +185,11 @@ destroy-full-local-setup: ## Destroy full local multicluster setup
 	k3d cluster delete $(CLUSTER_GSLB2)
 	docker network rm $(CLUSTER_GSLB_NETWORK)
 
+.PHONY: deploy-prometheus
+deploy-prometheus:
```
somaritane, line 189: @kuritka do you plan to add the possibility to uninstall Prometheus from the local clusters (`make uninstall-promethus` target) as well? If we install something, there should be a similar way to uninstall it.

somaritane: Makes sense to update contributor's documentation with that target

kuritka: @somaritane, all your comments implemented.

kuritka:
> @kuritka so it is for querying metrics with prom UI, correct?
> 
> @k0da more context: https://docs.openshift.com/container-platform/4.1/applications/operator_sdk/osdk-monitoring-prometheus.html. Not a fond of pulling either but discussing what's the most relevant

@ytsarev , yes.

k0da:
@ytsarev JFYI  "github.com/operator-framework/operator-sdk/pkg/metrics" is no longer there.

ytsarev:
@k0da yeah, it might be all deprecated with the major 1.0 release of operator-sdk
@somaritane we have ServiceMonitor mentioned in https://github.com/k8gb-io/k8gb/blob/master/docs/metrics.md IIRC it was even a part of the controller startup, is that correct?

somaritane:
> @somaritane we have ServiceMonitor mentioned in https://github.com/k8gb-io/k8gb/blob/master/docs/metrics.md IIRC it was even a part of the controller startup, is that correct?
@ytsarev Lots of things have changed with 1.0 Operator SDK migration, and the mentioned document definitely needs an update.  AFAIK, `ServiceMonitor` can be created automatically during operator deployment together with auth proxy Service, since it is part of the default kustomize scaffolding of operator deployment generated by OSDK.
We've got rid of k8gb kustomize deployment config at some point because it was not used. But I don't think we've taken support for Prometheus Operator to account when created helm chart. 
If we plan to support Prometheus Operator, we need to add the creation of the ServiceMonitor and respective Service object into the helm templates.

Makefile
```diff
@@ -487,3 +497,30 @@ define list-running-pods
 	@echo "\n$(YELLOW)Local cluster $(CYAN)$1 $(NC)"
 	kubectl get pods -A --context=k3d-$1
 endef
+
+define deploy-prometheus
+	@echo "\n$(YELLOW)Local cluster $(CYAN)$1$(NC)"
+	kubectl config use-context k3d-$1
```
somaritane, line 503: Consider specifying the context inline with `helm` and `kubectl` command options instead of setting it permanently with `kubectl config`. This might be good to avoid errors when dealing with multiple clusters. `dns-tools` target is a good example for kubectl, and helm has `--kube-context` option.

somaritane (changes requested):
@kuritka thanks for the documentation and uninstall make target updates. Looking good, just a few improvement suggestions from my side.

Makefile
kuritka: @somaritane , thx for review,  amended

somaritane (approved):
@kuritka LGTM!

Issue #528, opened 2021-06-18
'test-round-robin' often fails


tpiperatgod:
Hi, I deployed the local environment via `make-deploy-full-local-setup` command and then followed [Round Robin](https://github.com/k8gb-io/k8gb/blob/master/docs/local.md#round-robin) to experiment with the round robin demo.

I find that the following error is often reported:
```shell
~# make test-round-robin
If you don't see a command prompt, try pressing enter.
wget: can't connect to remote host (10.43.154.107): Connection timed out
pod "busybox" deleted
pod default/busybox terminated (Error)
Makefile:320: recipe for target 'test-round-robin' failed
make: *** [test-round-robin] Error 1

~# make test-round-robin
If you don't see a command prompt, try pressing enter.
wget: can't connect to remote host (10.43.62.9): Connection timed out
pod "busybox" deleted
pod default/busybox terminated (Error)
Makefile:320: recipe for target 'test-round-robin' failed
make: *** [test-round-robin] Error 1
```

The output of `kubectl describe po busybox` as follows(also in cluster k3d-test-gslb1).
```shell
Name:         busybox
Namespace:    default
Priority:     0
Node:         k3d-test-gslb2-server-0/172.19.0.4
Start Time:   Fri, 18 Jun 2021 11:11:15 +0800
Labels:       run=busybox
Annotations:  <none>
Status:       Running
IP:           10.42.0.13
IPs:
  IP:  10.42.0.13
Containers:
  busybox:
    Container ID:  containerd://641ba8d5f2800a56b7e296a932a31e5b94f3186a88cee8ae33a69ae498e74441
    Image:         busybox
    Image ID:      docker.io/library/busybox@sha256:930490f97e5b921535c153e0e7110d251134cc4b72bbb8133c6a5065cc68580d
    Port:          <none>
    Host Port:     <none>
    Args:
      sh
      -c
      echo 'nameserver 10.43.95.74' > /etc/resolv.conf && wget -qO -  roundrobin.cloud.example.com
    State:          Running
      Started:      Fri, 18 Jun 2021 11:11:17 +0800
    Ready:          True
    Restart Count:  0
    Environment:    <none>
    Mounts:
      /var/run/secrets/kubernetes.io/serviceaccount from default-token-xqrv5 (ro)
Conditions:
  Type              Status
  Initialized       True
  Ready             True
  ContainersReady   True
  PodScheduled      True
Volumes:
  default-token-xqrv5:
[...]

PR #532, opened 2021-06-21
Fix local playground setup

k0da:
Don't create service and don't point publish service to it. Istead use
node IPs as Ingress address

### DIFF
--- a/deploy/ingress/nginx-ingress-values.yaml
+++ b/deploy/ingress/nginx-ingress-values.yaml
@@ -7,7 +7,8 @@ controller:
     patch:
      enabled: false
   hostNetwork: true
-
+  publishService:
+    enabled: false
   ## Use host ports 80 and 443
   daemonset:
     useHostPort: false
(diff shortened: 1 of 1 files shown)


k0da:
fixes https://github.com/k8gb-io/k8gb/issues/528

ytsarev (approved):
Good to merge assuming we create additional Issue to cover the problematic e2e scenario with terratest.

@kuritka do you agree to handle it separately assuming that `make test-round-robin` is green now?

kuritka (approved):

kuritka:
@ytsarev, @k0da  agreed

ytsarev:
@k0da sorry if I missed it, was the issue for terratest extension created before the merge?

k0da:
@ytsarev just opened [one](https://github.com/k8gb-io/k8gb/issues/533)

ytsarev:
@k0da thank you 👍

PR #534, opened 2021-06-21
node IP's equal to Ingress IPs test

kuritka:
related to #532
terratest guarantees that ingress IP's are compatible with node internal IP's

|                    | k3d-test-gslb1              | k3d-test-gslb2              |
|--------------------|-----------------------------|-----------------------------|
| server-0           | `192.168.16.2`              | `192.168.16.4`              |
| agent-0            | `192.168.16.3`              | `192.168.16.5`              |
| test-gslb-failover | `192.168.16.4,192.168.16.5` | `192.168.16.4,192.168.16.5` |
| test-gslb          | `192.168.16.2,192.168.16.3` | `192.168.16.2,192.168.16.3` |


k0da:
@kuritka can this test be optional? In case of running terratest against real workload in AWS or with LoadBalancer in place, this test will fail

kuritka:
@k0da will we run it against aws ?

k0da:
@kuritka there is a possibility to run terratest against real clusters. I think it will be better to check if communication to workflow succeeds. Something like we have in test-round-robin make target

kuritka:
@k0da, What should be different ? In such case we can extend test abstraction to support AWS.

k0da:
@kuritka it won't depend on type of exposure, but rather if IP collected by k8gb is usable

ytsarev:
it would be easier to make http test for the failover scenario, this way it would cover the case of the remote cluster.
The fact of nodeIP==IngressIP is quite special configuration.
And yes, I was running terratest suite against real clusters in AWS so the current implementation will fail there.

terratest/utils/extensions.go
```diff
@@ -271,6 +271,19 @@ func (i *Instance) GetLocalTargets() []string {
 	return dig
 }
 
+func (i *Instance) GetInternalNodeIPs() (result []string) {
+	result = make([]string,0)
+	nodes := k8s.GetNodes(i.w.t,i.w.k8sOptions)
+	for _,n := range nodes {
+		for _, a := range n.Status.Addresses {
+			if a.Type == corev1.NodeInternalIP {
```
kuritka, line 279: @k0da, @ytsarev - sorry, the problem is this line ?

ytsarev: @kuritka for me it is the overall expectation of https://github.com/k8gb-io/k8gb/pull/534/files#diff-02a37d22e625bb75658c30a4e16e1d0408225b83209862a48285b63c4af98f35R37-R38

kuritka:
#538 reimplements

PR #537, opened 2021-07-01
Temporarly downgrade GitHub runner to Ubuntu 18.04

kuritka:
The GitHub runner has been upgraded to the latest [version 20210628.1](https://github.com/actions/virtual-environments/releases/tag/ubuntu20%2F20210628.1) which has changed the kernel.
The K3d action stopped working, so we don't get terratests, releases, etc..

I found that the tests would pass if I upgraded the k3d version, but unfortunately k3d has changed too much and just upgrading the version inside the k3d action is not enough and I will need more time to investigate.

That's why I'm temporarily downgrading the github runner until I fix the k3d-action.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   release:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-18.04
     env:
       DOCKER_CLI_EXPERIMENTAL: "enabled"
     steps:
--- a/.github/workflows/terratest.yaml
+++ b/.github/workflows/terratest.yaml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   terratest:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-18.04
     steps:
       - uses: actions/checkout@v1
 
--- a/.github/workflows/upgrade-testing.yaml
+++ b/.github/workflows/upgrade-testing.yaml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   upgrade-testing:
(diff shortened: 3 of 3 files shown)


k0da (approved):

ytsarev (commented):
Sorry if am missing something but aren't we using  ubuntu-20.4 for a while already?
```
2020-11-27 12:41:14 +0100 10)     runs-on: ubuntu-20.04
``` 
Shouldn't we try to pin to a specific patch release? as in `major.minor.patch` ?

kuritka:
> Sorry if am missing something but aren't we using ubuntu-20.4 for a while already?
> 
> ```
> 2020-11-27 12:41:14 +0100 10)     runs-on: ubuntu-20.04
> ```
> 
> Shouldn't we try to pin to a specific patch release? as in `major.minor.patch` ?

Hi, good question:

run-on defines Environment version, but behind that new ubuntu versions are deployed (last line)
Screenshot 2021-07-01 at 09 45 39


I was comparing raw logs and  this is  difference between a successful and unsuccessful run:
```
2021-06-28T09:48:22.3873627Z Environment: ubuntu-20.04
2021-06-28T09:48:22.3874229Z Version: 20210614.1
```
```
2021-06-30T13:31:45.1527941Z Environment: ubuntu-20.04
2021-06-30T13:31:45.1528600Z Version: 20210628.1
```

somaritane:
@kuritka , I had a similar issue with https://github.com/AbsaOSS/terraform-controller/pull/55, downgrade to `ubuntu-18.04` helped!

somaritane (approved):
@kuritka LGTM

Issue #533, opened 2021-06-21
Extend terratest suite with http end-to-end test.


k0da:
It would be nice to see ``curl`` for example is able to reach workload by using addresses k8gb provides.

PR #538, opened 2021-07-01
Failover HTTP test

kuritka:
related to #533

I enhanced terratest extensions and utils with a function that can run a command in busybox and a function that can read json output from test app.

Thanks to this I can automate tests from local playground.

Besides the test itself, I had to extend `WithTestApp(uiMessage string)` with the message argument and bump the podinfo from `4.0.6` to `5.0.1`.

I also implemented InstanceStatus, which is a functionality that takes all available instance data, such as Dig, endpoint status, CoreDNS IP etc... The functionality is useful for tracing errors and possible assertions.


kuritka:
this PR was raising, but I already fixed.

k0da (changes requested):
Typo in commit message: "subinfo"

k0da (approved):

PR #543, opened 2021-07-12
Cleanup old NS name format

k0da:
We changed nameserver name format, but for migration puposes we need to
take care and drop entries with an old format. Each cluster will take
care of its own entries.


kuritka (commented):
Hey @k0da , thanks for the PR. The implemented mechanism filters the original NS names from the EdgeDNS returned records per each reconciliation. 

The problem I see is that the records will remain in EdgeDNS forever. Similarly, the filtering mechanism remaining in the code forever, which without knowing the historical context will confuse contributors. 

Maybe it would be better to consider the migration as a separate k8gb class / package, which would contain the current and future migrations (separation of concerns).  
The migrations would be run once when the application is started (or redeployed) and would make permanent changes.  

Now we have chain:
read config -> logger -> reconciler + scheme builder -> provider -> start operator

So we can extend it like this:

read config -> logger -> reconciler + scheme builder -> provider -> **migration** (at this point we have everything important) -> start operator

I understand that the change I'm describing requires a bit more coding, although the result will be exactly the same, but in the long run we will find it easier to keep the system migration together.

k0da:
I see your concern. But cleaning up old naming should go away in one release. We don't have anything else yet to put into "migration machinery". This affects only infoblox implementation. The rest is handled by DNSEndpoint NS object, which shouldn't be affected by this.

kuritka:
approving based on plan we made during meeting, good job @k0da !

kuritka (approved):

somaritane (changes requested):
@k0da , @kuritka pls confirm, based on discussion, this should not go to master

k0da:
@somaritane yeah

kuritka:
@somaritane , sure, thx. Let's keep it blocked pls to not accidently push into master..

kuritka (approved):

ytsarev (approved):
After discussion, we agreed to merge it in with the long term strategy of removal of clean up code as part of the 0.9 release/milestone

PR #544, opened 2021-07-13
logging delegated records

kuritka:

Co-authored-by: Timofey Ilinykh <timofey.ilinykh@absa.africa>

### DIFF
--- a/controllers/providers/dns/infoblox.go
+++ b/controllers/providers/dns/infoblox.go
@@ -106,6 +106,7 @@ func (p *InfobloxProvider) CreateZoneDelegationForExternalDNS(gslb *k8gbv1beta1.
 			}
 
 			if !reflect.DeepEqual(findZone.DelegateTo, currentList) {
+				log.Info().Msgf("Found delegated zone records (%v)", findZone.DelegateTo)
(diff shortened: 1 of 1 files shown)


k0da (approved):

controllers/providers/dns/infoblox.go
```diff
@@ -106,6 +106,7 @@ func (p *InfobloxProvider) CreateZoneDelegationForExternalDNS(gslb *k8gbv1beta1.
 			}
 
 			if !reflect.DeepEqual(findZone.DelegateTo, currentList) {
+				log.Debug().Msgf("FindZone delegated records (%v)", findZone.DelegateTo)
```
somaritane, line 109: Since this message appears only in case of discrepancies, IMO it is safe to make it on the Info level instead of Debug.
Also, `findZone` is an internal implementation, I suggest replacing it with a more user-friendly explanation.  
```suggestion
				log.Info().Msgf("Found delegated zone records (%v)", findZone.DelegateTo)
```

somaritane (changes requested):

k0da (approved):

somaritane (approved):
@kuritka, lgtm!

PR #545, opened 2021-07-13
Handling error code for defer functions

kuritka:
related to #124

because of the metrics we have to run the deferred function even if an error occurred during main processing. Until now, we used to call `os.Exit(1)` in main if any error occurred. Unfortunately, `os.exit` skips the deferred function call.

The solution is implementation. In case of an error we call `return` instead of `os.Exit`, and `os.Exit` itself is a deferred function.

### DIFF
--- a/main.go
+++ b/main.go
@@ -48,6 +48,10 @@ func init() {
 }
 
 func main() {
+	var exitCode = 1
+	defer func() {
+		os.Exit(exitCode)
+	}()
 	var f *dns.ProviderFactory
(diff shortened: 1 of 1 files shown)


ytsarev (approved):
@kuritka looks good! Just thinking, is it something we can send upstream to operator-sdk? If I am not mistaking the main structure is inherited from the generated code

kuritka:
@ytsarev , thx. Yes it's inherited from generated code which doesn't contain any deferred functions. But yes, we can try to push upstream as template.

PR #551, opened 2021-07-23
Use `k8gb` prefix for external dns rbac

### LINKED ISSUE #542: Can't deploy K8GB in a Cluster that has an Existing ExternalDNS Deployment
Hi -

First, great project - big fan of the idea and implementation.   I have been talking to Yury on a few things and as I was adding K8GB to our second edition book, I came across a potential bug.   

If a cluster has ExternalDNS already installed, you cannot deploy K8GB due to a K8GB looking for labels and annotation's on the ExternalDNS ClusterRole and Binding:

Error: rendered manifests contain a resource that already exists. Unable to continue with install: ClusterRole "external-dns" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "k8gb"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "k8gb"

Error: rendered manifests contain a resource that already exists. Unable to continue with install: ClusterRoleBinding "external-dns-viewer" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "k8gb"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "k8gb"

I could remediate this by adding the missing values, but I was hoping for a "cleaner" solution.  I have run into this with a few other projects, where one project assumes it is the owner of a resource and in a cluster that might have a solution like ExternalDNS and K8GB, this will cause issues.

Anything else I can add, let me know.  Thanks!

ytsarev:
* Use k8gb prefix for rbac resources to avoid the conflict
  with the standard upstream external-dns names
* Fixes #542

### DIFF
--- a/chart/k8gb/templates/external-dns/external-dns.yaml
+++ b/chart/k8gb/templates/external-dns/external-dns.yaml
@@ -14,7 +14,7 @@ spec:
       labels:
         app: external-dns
     spec:
-      serviceAccountName: external-dns
+      serviceAccountName: k8gb-external-dns
       securityContext:
         fsGroup: 65534 # For ExternalDNS to be able to read Kubernetes and AWS token files
         runAsUser: 1000
--- a/chart/k8gb/templates/external-dns/rbac.yaml
+++ b/chart/k8gb/templates/external-dns/rbac.yaml
@@ -2,7 +2,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: external-dns
+  name: k8gb-external-dns
 rules:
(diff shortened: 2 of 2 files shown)


k0da (changes requested):
I think you need to change POD's sa then.

k0da (approved):
LGTM

PR #553, opened 2021-08-06
Update to external-dns v0.9.0

k0da:
* update to external-dns v0.9.0
* refactor provider options into helm functions

Signed-off-by: Dinar Valeev <dinar.valeev@absa.africa>

PR #555, opened 2021-08-09
[Fix] busybox HitTestApp

kuritka:
Unfortunately I ran the crashed upgrade tests and lost the logs but #553 (upgrade tests) crashed on hitting busybox returning wget error.

What happened?
In our abstraction we use StartTestApp which changes the replicaset to 1 instance of the podinfo. The application is Healthy , localtargets are in the correct state but the application itself is still booting. If I make an HTTP request while the application is booting, I get 503 instead of the expected json.

This fix polls the podinfo for 3 seconds while the application is booting. It also prints `response.Body` if the response is not the expected json.

Here is an example where polling was successfully used https://github.com/k8gb-io/k8gb/runs/3281617073?check_suite_focus=true#step:6:2492


k0da (approved):
LGTM

ytsarev (approved):

PR #561, opened 2021-08-17
Propogate DNS Zone Negative TTL down to CoreDNS

k0da:

### DIFF
--- a/chart/k8gb/values.yaml
+++ b/chart/k8gb/values.yaml
@@ -11,6 +11,7 @@ k8gb:
   # but can be overrided with imageTag key
   # imageTag:
   dnsZone: "cloud.example.com" # dnsZone controlled by gslb
+  dnsZoneNegTTL: 300 # Negative TTL for SOA record
   edgeDNSZone: "example.com" # main zone which would contain gslb zone to delegate
   edgeDNSServer: "1.1.1.1" # use this DNS server as a main resolver to enable cross k8gb DNS based communication
   clusterGeoTag: "eu" # used for places where we need to distinguish between differnet Gslb instances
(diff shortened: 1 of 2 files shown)


k0da:
Requires https://github.com/k8gb-io/coredns-crd-plugin/pull/16

kuritka (dismissed):

chart/k8gb/values.yaml
```diff
@@ -11,6 +11,7 @@ k8gb:
   # but can be overrided with imageTag key
   # imageTag:
   dnsZone: "cloud.example.com" # dnsZone controlled by gslb
+  dnsZoneNegTTL: 3600 # Negative TTL for SOA record
```
ytsarev, line 14: why default negative is 3600 when it is clearly problematic got GSLB scenario? 
```suggestion
  dnsZoneNegTTL: 60 # Negative TTL for SOA record
```

ytsarev (changes requested):

chart/k8gb/values.yaml
k0da: What about we don't set any default value and keep plugins one, if set it will be propogated

ytsarev: @k0da i'm concerned that another customers will hit the same issue without having a clue that the root cause is actually negative TTL.  It is safer to keep it low imho

k0da: 10 minutes?

ytsarev: why?

k0da: IMO 1 minute is too agressive

k0da: agreed on 5 minutes

ytsarev (approved):
Looks great, thanks

somaritane (approved):

PR #564, opened 2021-08-19
Health status enum

kuritka:
for type safety I need HealthStatus to be an enumerator.
The most important change is that I had to change the GslbTypes from string to HealthStatus.


ytsarev (approved):
Looks great 👍

PR #586, opened 2021-08-28
Add GitHub buttons to k8gb.io header

ytsarev:
* Add https://ghbtns.com/ powered buttons
* Remove dead tar.gz downloading code


ytsarev:
Looks like
image

somaritane:
Can we try to move these to the footer? Or maybe we need to restructure and slim down the template a bit.
I'm just afraid that by adding these we already eating out half of the screen estate, leaving little place for the actual content.

The same goes about the mobile view, (probably it needs some burger button for navigation instead of vertical buttons):

ytsarev:
@somaritane moving to footer defeats the obvious goal of the star gathering ;)

kuritka (approved):
👍

somaritane (approved):

PR #587, opened 2021-08-30
Make K8gbGslbStatusPerIngressHosts metric readable

kuritka:
For readibility I splited K8gbGslbStatusPerIngressHosts into three different metrics:
 - k8gb_gslb_healthy_records
 - k8gb_gslb_unhealthy_records
 - k8gb_gslb_not_found_records

The metric observes staus of GSLB serviceHealth
```json
status:
  geoTag: us
  healthyRecords:
    roundrobin.cloud.example.com:
    - 192.168.240.2
    - 192.168.240.3
  serviceHealth:
    notfound.cloud.example.com: NotFound
    roundrobin.cloud.example.com: Unhealthy
    unhealthy.cloud.example.com: Unhealthy
```

I renamed `k8gb_gslb_healthy_records` to `k8gb_endpoint_healthy_records`.
The metric observes number of endpoint targets:
```json
spec:
  endpoints:
  - dnsName: roundrobin.cloud.example.com
    labels:
      strategy: roundRobin
    recordTTL: 30
    recordType: A
    targets:
    - 192.168.240.2
    - 192.168.240.3
```

### DIFF
--- a/controllers/providers/metrics/prometheus.go
+++ b/controllers/providers/metrics/prometheus.go
@@ -38,9 +38,8 @@ const (
 
 const (
 	K8gbGslbErrorsTotal               = "k8gb_gslb_errors_total"
-	K8gbGslbHealthyRecords            = "k8gb_gslb_healthy_records"
+	K8gbEndpointHealthyRecords        = "k8gb_endpoint_healthy_records"
 	K8gbGslbReconciliationLoopsTotal  = "k8gb_gslb_reconciliation_loops_total"
-	K8gbGslbStatusPerIngressHosts     = "k8gb_gslb_status_per_ingress_hosts"
 	K8gbGslbStatusCountForFailover    = "k8gb_gslb_status_count_for_failover"
 	K8gbGslbStatusCountForRoundrobin  = "k8gb_gslb_status_count_for_roundrobin"
 	K8gbGslbStatusCountForGeoIP       = "k8gb_gslb_status_count_for_geoip"
@@ -48,12 +47,14 @@ const (
 	K8gbInfobloxHeartbeatErrorsTotal  = "k8gb_infoblox_heartbeat_errors_total"
 	K8gbInfobloxZoneUpdatesTotal      = "k8gb_infoblox_zone_updates_total"
 	K8gbInfobloxZoneUpdateErrorsTotal = "k8gb_infoblox_zone_update_errors_total"
+	K8gbGslbHealthyRecords            = "k8gb_gslb_healthy_records"
+	K8gbGslbUnhealthyRecords          = "k8gb_gslb_unhealthy_records"
+	K8gbGslbNotFoundRecords           = "k8gb_gslb_not_found_records"
 )
 
 // collectors contains list of metrics.
 type collectors struct {
-	K8gbGslbHealthyRecords            *prometheus.GaugeVec
-	K8gbGslbStatusPerIngressHosts     *prometheus.GaugeVec
+	K8gbEndpointHealthyRecords        *prometheus.GaugeVec
 	K8gbGslbStatusCountForFailover    *prometheus.GaugeVec
 	K8gbGslbStatusCountForRoundrobin  *prometheus.GaugeVec
 	K8gbGslbStatusCountForGeoip       *prometheus.GaugeVec
@@ -63,6 +64,9 @@ type collectors struct {
 	K8gbInfobloxZoneUpdateErrorsTotal *prometheus.CounterVec
 	K8gbInfobloxHeartbeatsTotal       *prometheus.CounterVec
 	K8gbInfobloxHeartbeatErrorsTotal  *prometheus.CounterVec
+	K8gbGslbHealthyRecords            *prometheus.GaugeVec
+	K8gbGslbUnhealthyRecords          *prometheus.GaugeVec
+	K8gbGslbNotFoundRecords           *prometheus.GaugeVec
 }
 
 type PrometheusMetrics struct {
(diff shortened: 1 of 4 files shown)


ytsarev (changes requested):
Some suggestions.

```
I renamed k8gb_gslb_healthy_records to k8gb_endpoint_healthy_records.

```
Also, it is not clear if above mentioned metrics should coexist and why

kuritka:
> Some suggestions.
> 
> ```
> I renamed k8gb_gslb_healthy_records to k8gb_endpoint_healthy_records.
> ```
> 
> Also, it is not clear if above mentioned metrics should coexist and why

The question is whether someone needs to monitor Service health, which doesn't say anything about guests. It only shows the number of Healthy, Unhealthy, and NotFound in time.

Similar is the DNSEndpoint tracking, do we want to monitor the number of A records ? 

@k0da , @somaritane , @jkremser, WDYT?

jkremser:
imho, the more metrics the better at the beginning. It's pretty common to deploy also k8s node exporters and a lot of info can be obtained there so probably we should not be redundant but instead report only k8gb specific metrics. Not sure if the `k8gb_gslb_healthy_records` metric can be somehow directly calculated by some smart PromQL from pods statuses, but still if I get it right it's kind of more abstract and app specific, so why not to report it.

Ideally we should come up with some list of metrics that we find useful and the others will come from people who actually use it and needs to alert on them or trying to detect some drift. I can imagine that coredns have it's own set of exported metrics as well that might be actually useful in conjunction with ours.

For my previous operators, I've been also maintaining metrics like counts of created CRs or JVM internal metrics (cos it was Java based), version of the app, gitSha, versions of tools it was compiled with, etc. The string based metrics are kind of tricky w/ Prometheus, but there is a workaround to this, using them as a label and having the value of the metrics set to something dummy like 1.

controllers/providers/metrics/prometheus.go
```diff
@@ -174,20 +178,36 @@ func (m *PrometheusMetrics) Unregister() {
...
 func (m *PrometheusMetrics) init() {
+	m.metrics.K8gbEndpointHealthyRecords = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: K8gbEndpointHealthyRecords,
+			Help: "Number of healthy records observed by GSLB.",
+		},
+		[]string{"namespace", "name"},
+	)
+
 	m.metrics.K8gbGslbHealthyRecords = prometheus.NewGaugeVec(
```
somaritane, line 189: I guess the metric name should be changed too in order to reflect the updated metric description ("Number of unhealthy Services observed by GSLB.")
```suggestion
	m.metrics.K8gbGslbHealthyServices = prometheus.NewGaugeVec(
```

somaritane:
> The string based metrics are kind of tricky w/ Prometheus, but there is a workaround to this, using them as a label and having the value of the metrics set to something dummy like 1.

IIRC this is exactly how the `K8gbGslbStatusPerIngressHosts` metric was organized, by using the status type ("NotFound", "Healthy", "Unhealthy") as a label, so that status count can be later found out for particular status type in a query. 
Not sure if it makes sense to split `K8gbGslbStatusPerIngressHosts` into 3 metrics.

jkremser:
> IIRC this is exactly how the K8gbGslbStatusPerIngressHosts metric was organized, by using the status type ("NotFound", "Healthy", "Unhealthy") as a label, so that status count can be later found out for particular status type in a query.

correct, but these were and still are normal gauges that stores doubles, I was more talking about string based metrics, where you would like to expose some strings, like what's the version, what's the gitSha, image tag, etc. and there is a hack for this:

you can have a metric

```
k8gb_my_fancy_runtime_info{k8gb_version="1.0.0",goLangVersion="1.17",gitSha="xyz"} 1.0
```
I.e. storing them as a key-value pairs where key is the label name and value is the label value and the real value of the metric is not important at all so it can be `1.0`, little hackish, but it's being used especially if you need this info in Grafana where you can't call `kubectl` to get the info from annotations/labels/env vars.

somaritane:
> ```
> k8gb_my_fancy_runtime_info{k8gb_version="1.0.0",goLangVersion="1.17",gitSha="xyz"} 1.0
> ```
Ah I see, looks like a good way to store project-related metadata 👍

kuritka:
@jkremser , @somaritane Thx for helpful discussion.

I think you've summed up well what is needed
 - update, metric.md
 - K8gbGslbStatusPerIngressHosts leave as it is (don't split)
 - introducing the `k8gb_runtime_info` would be sufficient
 
Basically I agree, I decided to close this PR and create a new one(s) according to the discussion.  I will probably move `k8gb_endpoint_healthy_records` to the new PR. 

@ytsarev, @k0da , @jkremser , @somaritane  shout if disagree.

somaritane:
> Basically I agree, I decided to close this PR and create a new one(s) according to the discussion. I will probably move `k8gb_endpoint_healthy_records` to the new PR.
> 
> @ytsarev, @k0da , @jkremser , @somaritane shout if disagree.

@kuritka, thanks for summing things up, the plan looks good, just one thing from my side:

> * K8gbGslbStatusPerIngressHosts leave as it is (don't split)

I think it still makes sense to improve the naming of this metric and remove the confusion about its purpose and the nature of its values.
Here's the suggestion:
`K8gbGslbStatusPerIngressHosts` -> `K8gbGslbServiceStatusCount (k8gb_gslb_service_status_count)`
What do you think?

kuritka:
Based on the discussion I have decided to close this PR and replace it with several other PRs. See log for references.

PR #589, opened 2021-08-31
Release build optimisations

kuritka:
this PR affects three things
 - (`-X`) setting the k8gb version during the compile time, version is logged. The default version on local environment is `development` while prod version has release tag e.g: `v0.8.2`.
 - (`-s -w`) deletion of linter stamps that are not needed in the production binary (reduction from 47MB to 38MB).
 - (`-trimpath`) remove all file system paths from the resulting executable. Instead of file system paths,   a plain import path is printed. (Instead of  workspace/controllers/providers/assistant/gslb.go:68 -> github.com/AbsaOSS/k8gb/controllers/providers/assistant/gslb.go:68)

The PR doesn't affect local environment

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -12,6 +12,10 @@ builds:
     goarch:
       - amd64
       - arm64
+    flags:
+      - -trimpath
+    ldflags:
+      - -s -w -X main.Version={{ .Tag }}
 dockers:
 - image_templates:
   - "absaoss/k8gb:{{ .Tag }}-amd64"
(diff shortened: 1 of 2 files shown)


ytsarev (approved):
Looks great, I would just not name it PROD but rather Release build as the target deployment environment can be different

kuritka:
> Looks great, I would just not name it PROD but rather Release build as the target deployment environment can be different

thx @ytsarev, I PR name changed.

PR #590, opened 2021-08-31
align go1.16

kuritka:
In the main go.mod we have been using GO1.16 for some time, but the test dockerfile has not been changed, terratests and releaser are still on GO1.15. I'm aligning the version


k0da (approved):

ytsarev (approved):
good spot

PR #605, opened 2021-09-07
s/edgeDNSServer/edgeDNSServers/g

### LINKED ISSUE #154: Oppertunity to enhance edgeDNSServer from single instance to multiple
This would allow in the use case where you may have more than 1 DNS server to query (for redundancy)

```
edgeDNSServer: &edgeDNSServer:
  - 1.1.1.1
  - 2.2.2.2 
```

jkremser:
As described in the Issue #154 this adds the ability to fall-back to another edge dns server from the list. These are passed to the operator as env variable `EDGE_DNS_SERVERS` (was singular before) and the value is a comma-separted list of ips/hostnames followed optionally with port number. example: `1.1.1.1:11, 2.2.2.2,hostname` (spaces between entries are supported).

Then function `Dig` in `dns.go` has been modified to work with this list (variadic function) and also `Exchange` function in the same file was tweaked to support it.

This change doesn't break the backward compatibility, because both `EDGE_DNS_{SERVER,PORT}` still works, but user is notified in the operator logs that these two were deprecated.

Some old tests needed to be modified in order to pass and I've added also these new tests:
(`in controllers/depresolver/depresolver_test.go`)
-  TestResolveConfigWithEmptyEdgeDNSServersKey
-  TestResolveConfigWithMalformedEdgeDNSServersKey
-  TestResolveConfigWithTwoEdgeDnsServers
-  TestResolveConfigWithMultipleEdgeDnsServers1
-  TestResolveConfigWithMultipleEdgeDnsServers2
-  TestResolveConfigWithMultipleEdgeDnsServersMalformed1
-  TestResolveConfigWithMultipleEdgeDnsServersMalformed2
-  TestResolveConfigWithInvalidEdgeDnsServersValue

(in `controllers/gslb_controller_test.go`)
- TestReturnsExternalRecordsUsingFailoverStrategyAndFallbackDNSserver

(in `controllers/internal/utils/dns_test.go`)
-  TestOneValidEdgeDNSInTheList
-  TestNoValidEdgeDNSInTheList
-  TestEmptyEdgeDNSInTheList
-  TestMultipleValidEdgeDNSInTheList
-  TestEmptyDNSList
-  TestValidDigFQDNWithDot


ytsarev:
Thanks for the early exposure, I've converted it to Draft

kuritka (changes requested):
thx for the PR. We shouldn't keep in master untested code or have possibility to start app with invalid inputs.
Please add tests and validations.

kuritka:
@jkremser, I don't see all checks including linter, testrun, terratest etc. Could you make PR from feature branch?

ytsarev:
@kuritka that's kinda of weird - workflows should be executed for all PRs even from the remote forks aren't they?

kuritka:
@ytsarev , I think @jkremser must allow actions in his fork, where workflows are running

jkremser:
> @ytsarev , I think @jkremser must allow actions in his fork, where workflows are running

I've enabled the actions on my fork, let me try to close and open the pr

jkremser:
@kuritka 
Enabling the gh actions on my fork is not related to the fact that the actions are not run against the pull request from a different fork. This should be fixed on our (`k8gb-io/k8gb`) repo. I will create another PR for this.

kuritka:
@jkremser, please  squash commits into one , align PR message with squashed commit.
thx

kuritka (changes requested):
@jkremser, please  squash commits into one , align PR message with squashed commit.
thx

controllers/internal/utils/dns.go
```diff
@@ -20,30 +20,99 @@ package utils
...
+		return msg, err
+	}
+	return nil, fmt.Errorf("exchange error: all dns servers were tried and none of them were able to resolve, err: %s", err)
+}
+
+func OneToString(s DNSServer) string {
+	return fmt.Sprintf("%s:%v", s.Host, s.Port)
+}
+
+func AllToString(s []DNSServer) string {
```
kuritka, line 112: Is used by depresolver only, could be private within depresolver package. 

not necesary, but consider:
```go
type DNSList []DNSServer

func (l DNSList) StringSlice() (s []string) {
	for _, x := range l {
		s = append(s, x.String())
	}
	return
}
// strings.Join(DNSList(xx).StringSlice(),",")

or

func (l DNSList) String() (s string) {
	var a []string
	for _, x := range l {
		a = append(a, x.String())
	}
	return strings.Join(a,",")
}

// fmt.Println(DNSList(xx))
```

controllers/internal/utils/dns.go
```diff
@@ -20,30 +20,99 @@ package utils
...
+func DigWithFallback(edgeDNSServers []DNSServer, fqdn string) ([]string, error) {
+	var dig dnsutil.Dig
+	if len(edgeDNSServers) == 0 {
+		return nil, fmt.Errorf("empty edgeDNSServers, provide at least one")
 	}
-	a, err := dig.A(fqdn)
-	if err != nil {
-		err = fmt.Errorf("dig error: can't dig fqdn(%s) with error(%s)", fqdn, err)
-		return nil, err
+	for i := 0; i < len(edgeDNSServers); i += 2 {
```
kuritka, line 51: why 0,2,4,.. ??

controllers/internal/utils/dns.go
```diff
@@ -20,30 +20,99 @@ package utils
...
+		if edgeDNSServers[i].Host == "" {
```
kuritka, line 52: is that valid ? Shouldn't fail here?

controllers/internal/utils/dns.go
```diff
@@ -20,30 +20,99 @@ package utils
...
-		err = fmt.Errorf("dig error: can't set query dns (%s) with error(%s)", edgeDNSServer, err)
-		return nil, err
+	dnsServer := DNSServer{
+		Host: edgeDNSServer,
+		Port: 53,
+	}
+	return DigWithFallback([]DNSServer{dnsServer}, fqdn)
+}
+
+func DigWithFallback(edgeDNSServers []DNSServer, fqdn string) ([]string, error) {
```
kuritka, line 46: Please, could you describe, what should be result of function and how it works?

Basically it's very hard to read implementation here.  Is it necessary to distinguish between odd and even elements? 

Many states can occur in this function, should be tested for  possible corner cases (currently tested for not nil and result[0]) .

controllers/internal/utils/dns.go
```diff
@@ -20,30 +20,99 @@ package utils
...
+		a, err := dig.A(fqdn)
+		if err != nil {
+			if len(edgeDNSServers) > i+2 { // there are still fallback dns servers listed
+				continue
+			} else {
+				err = fmt.Errorf("dig error: can't dig fqdn(%s) with error(%s)", fqdn, err)
+				return nil, err
+			}
+		}
+		var IPs []string
```
kuritka, line 82: `var ips []string` local variable

kuritka (changes requested):
looks, ok. 
`Dig` needs to be refactored and must be tested for corner cases. PR affects several parts of k8gb and I didn't have time to go through everything. I will continue with the review tomorrow.

controllers/internal/utils/dns.go
```diff
@@ -20,30 +20,99 @@ package utils
...
+	"github.com/miekg/dns"
 )
 
+type DNSServer struct {
+	Host string
+	Port int
+}
+
 // Dig retrieves list of tuple <IP address, A record > from edge DNS server for specific FQDN
 func Dig(edgeDNSServer, fqdn string) ([]string, error) {
```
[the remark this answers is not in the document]
jkremser: the old `Dig` is still being called from couple of tests so I was thinking to left it there, because it's easier to call it as a one-liner rather than preparing that array first (4 lines or so), but the var args function will probably do the trick, so yea, let me refactor it 👍  then it would be called w/ the `...` expansion on that array, right?

controllers/internal/utils/dns.go
```diff
@@ -20,30 +20,99 @@ package utils
...
+		msg, err = dns.Exchange(m, OneToString(ns))
+		if err != nil {
+			continue
+		}
+		return msg, err
+	}
+	return nil, fmt.Errorf("exchange error: all dns servers were tried and none of them were able to resolve, err: %s", err)
+}
+
+func OneToString(s DNSServer) string {
```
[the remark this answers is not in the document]
jkremser: If I do that will the `String()` func work ootb on the array of that type? Or do I have to do it via type alias, like:

```go
type DNSServers = []DNSServer
func (foo DNSServers) String() string{ ... }
```

controllers/internal/utils/dns.go
jkremser: because in one iteration we actually use two dns servers from the list, one is set as a primary one (`SetDNS()`) and other as a backup one (`SetBackupDNS()`)

controllers/internal/utils/dns.go
jkremser: hmm, I think this should never occur when called from the operator, because the empty dns server should fail the validation phase, but it was there before so I left it there.

The `Dig` helper is called also from tests so I assume the check is there because of it. The goal of the function (as I understand it) is to get the `[]DNSServer` by a) using explicitly selected dns server b) using a list of dns servers in a fallback fashion that if first fails, another tries and whoever will get the results first will return them.

controllers/internal/utils/dns.go
jkremser: good catch, I will add more tests and some comment

jkremser: actually it's being tested for 3 elements and 4 elements, in dns_test.go (`TestOneValidEdgeDNSInTheList` and `TestNoValidEdgeDNSInTheList`)

controllers/internal/utils/dns.go
jkremser: couldn't this be catched actually by some of our linters?

controllers/internal/utils/dns.go
```diff
@@ -20,30 +20,99 @@ package utils
...
+}
+
+func AllToString(s []DNSServer) string {
+	var aux []string
+	for _, el := range s {
+		aux = append(aux, OneToString(el))
 	}
-	sort.Strings(IPs)
-	return IPs, nil
+	return strings.Join(aux, ",")
```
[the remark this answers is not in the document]
jkremser: done, these are now gone and I went with the approach you've suggested (`.String()`), which looks much more golang-idiomatic

controllers/internal/utils/dns.go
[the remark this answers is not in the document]
kuritka: Should we keep the function just because it is used by tests ?

controllers/internal/utils/dns.go
kuritka: This is the question. The point is, the utility `Dig` function is `public static`, which means, it can be potentionally called out of operator universe (internality is irrelevant now). That means, the input should be validated (at least for empty args etc..) and tested.

controllers/depresolver/depresolver_test.go
```diff
@@ -459,53 +533,98 @@ func TestResolveConfigWithInvalidIpAddressEdgeDnsServer(t *testing.T) {
 	// arrange
 	defer cleanup()
 	expected := predefinedConfig
-	expected.EdgeDNSServer = "22.147.90.2."
+	expected.EdgeDNSServers = []utils.DNSServer{
+		{
+			Host: "22.147.90.2.",
```
ytsarev, line 538: Let's use more dummy ip here please

ytsarev (changes requested):

controllers/depresolver/depresolver_test.go
jkremser: `1.2.3.4`? or CSI-Miami-lvl dummy :D

jkremser:
>@jkremser, please squash commits into one , align PR message with squashed commit.

why? iirc, github should do that automatically during the merge, including the messages. Isn't it easier to review couple of independent commits in the PR rather than 1 big blob? I can do that, but I am really interested why anybody would want to do it.

controllers/depresolver/depresolver_test.go
jkremser: I will wait for https://github.com/k8gb-io/k8gb/pull/616 to go in and then resolve the conflicts

jkremser: done

chart/k8gb/values.yaml
```diff
@@ -14,7 +14,8 @@ k8gb:
   dnsZone: "cloud.example.com" # dnsZone controlled by gslb
   dnsZoneNegTTL: 300 # Negative TTL for SOA record
   edgeDNSZone: "example.com" # main zone which would contain gslb zone to delegate
-  edgeDNSServer: "1.1.1.1" # use this DNS server as a main resolver to enable cross k8gb DNS based communication
+  edgeDNSServers: 
+      - "1.1.1.1" # use this DNS server as a main resolver to enable cross k8gb DNS based communication
```
ytsarev, line 18: How does it behave with the old value? What customers will see when they realize we broke backward compatibility?

jkremser: good point, it's a breaking change so it's going to fail during the operator startup during the validation with:

```
2021-09-13T12:22:35Z ERR workspace/main.go:66 > can't resolve environment variables error="'EDGE_DNS_SERVERS' is empty"
```

So I might add some logic to `depresolver_config.go` that would produce a better message in case the `EDGE_DNS_SERVER` is set but `EDGE_DNS_SERVERS` is not.  We may also accept the old format and internally transform it to a list with just one entry and produce some warning htat this has been deprecated and use the new format because it may be removed in the new version. On the other hand it's still under `1.0.0` so breaking is allowed according to sem-ver :D

Another thing to consider is the helm chart discrepancy. Meaning when deploying newer version of k8gb w/ older helm chart by explicitly specifying the `k8gb.imageTag=somethingNewer` or vice-versa (updated chart and deploying some old image that had different "env. variable contract"). Handling it in the helm chart with conditions would make it quite unreadable and pain to maintain, imho

ytsarev: depresolver to fail fast is a must imho. We can try internal conversion, that might be classy and actually very user friendly. Helm chart version might be irrelevant as the custom values are orthogonal to it and usually kept somewhere in gitops-like repo not directly connected to specific chart version.

controllers/internal/utils/dns.go
jkremser: I've added some defensive programming-fu for the Dig itself in https://github.com/k8gb-io/k8gb/commit/ea4c3c950edc6862f310d0bce06b49dad741c8e6#diff-2b429a0ff271b1687026b627ed409f0973f341a47d1f3c4a3f711b42adbc11ccR58 I am still not convinced about failing fast in the `DigWithFallback` in case of one empty sever because it would kind of defeat the purpose of being robust variant of Dig that wouldn't fail if the server is, say, unreachable

But if others also think this is better, I can change that. It's 1 line of code anyway :D

kuritka:
> > @jkremser, please squash commits into one , align PR message with squashed commit.
> 
> why? iirc, github should do that automatically during the merge, including the messages. Isn't it easier to review couple of independent commits in the PR rather than 1 big blob? I can do that, but I am really interested why anybody would want to do it.

so we can easily track the changes. One PR, one commit. Besides, when I do a local review, I have to diff in individual commits although I need to see all changes at once. 
The same goes for description, which is not part of commit message, but exists only on github.

controllers/internal/utils/dns.go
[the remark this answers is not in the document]
jkremser: hmm, you are right, I will change that to one single `Dig` implementation that will do the fallback

jkremser:
> One PR, one commit

but that's exactly what git hub does, when you hit that `squash and merge` button... check some commits on k8s upstream for instance: https://github.com/kubernetes/kubernetes/pull/104923 & https://github.com/kubernetes/kubernetes/pull/104925 , etc.

I mean I can do that, but is this really some kind of policy how to contribute to k8gb? If yes, I am asking about the reasoning behind it and if it makes any sense, because if there is a contribution from a community are you also going to ask them to squash everything even though, it's not kind of standard and github can do that for you, including the messages part.. It just merges all the commit messages into a list and put it to the squashed commit. As for the description in the PR, there is a correspondence going from the commit to PR link so why having it twice, I like things "DRY"

chart/k8gb/values.yaml
jkremser: did that in https://github.com/k8gb-io/k8gb/pull/605/commits/f19dcaca3b857b22b5a2097bb28ea659a7646712

then it looks like this in the logs:

```
...
2021-09-13T14:45:17Z INF workspace/controllers/logging/logr.go:54 > metrics server is starting to listen {}
2021-09-13T14:45:17Z WRN workspace/main.go:88 > 'EDGE_DNS_SERVER' has been deprecated, use EDGE_DNS_SERVERS instead. Details: Pass the hostname or IP address as comma-separated list
2021-09-13T14:45:17Z INF workspace/main.go:91 > Registering components
...
```

So it doesn't fail, but transforms the old format into a new one + the warning in the logs in case of using the deprecated format, so we are now backward-compatible

controllers/internal/utils/dns.go
jkremser: I've changed that

controllers/internal/utils/dns.go
```diff
@@ -20,30 +20,92 @@ package utils
...
+		}
+		sort.Strings(ips)
+		return ips, nil
 	}
-	a, err := dig.A(fqdn)
-	if err != nil {
-		err = fmt.Errorf("dig error: can't dig fqdn(%s) with error(%s)", fqdn, err)
-		return nil, err
+	return nil, fmt.Errorf("dig error: all dns servers were tried and none of them were able to resolve")
+}
```
kuritka, line 95: can't we just do that simpler way ? 
(find first  server responding, if ok than, provide iplist ? 
As benefit we can completelly remove dependency on `"github.com/lixiangzhong/dnsutil"`

```go
func Dig(fqdn string, edgeDNSServers ...DNSServer) (ips []string, err  error) {
	var ack *dns.Msg
	if len(edgeDNSServers) == 0 {
		return nil, fmt.Errorf("empty edgeDNS servers")
	}

	var processAck = func(ack *dns.Msg) (ips []string) {
		for _, a := range ack.Answer {
			ips = append(ips, a.(*dns.A).A.String())
		}
		sort.Strings(ips)
		return
	}

	for _, ns := range edgeDNSServers {
		r := new(dns.Msg)
		r.SetQuestion(fqdn+".", dns.TypeA)
		ack, err = dns.Exchange(r, fmt.Sprintf("%s:%v",ns.Host, ns.Port))
		if err == nil {
			return processAck(ack), nil
		}
	}

	err = fmt.Errorf("dig error: can't dig fqdn(%s) with error(%s)", fqdn, err)
	return
}
```

controllers/internal/utils/dns_test.go
```diff
@@ -46,21 +54,51 @@ func TestEmptyFQDNButValidEdgeDNS(t *testing.T) {
...
-	edgeDNSServer := ""
 	fqdn := "whatever"
 	// act
-	result, err := Dig(edgeDNSServer, fqdn)
+	result, err := Dig(fqdn, DNSServer{Host: "", Port: 53})
+	// assert
+	assert.Error(t, err)
+	assert.Nil(t, result)
+}
+
```
kuritka, line 74: Happy new tests were added, please cover corner cases for: 

- TestEmptyEdgeDNSInTheList - empty list testing guardian condition 
- TestMultipleEdgeDNSInTheList - we are testing only one and zero valid edgeDNS servers, this test wouldcover more

controllers/internal/utils/dns.go
```diff
@@ -20,30 +20,92 @@ package utils
...
+func (l DNSList) String() string {
+	var aux []string
+	for _, el := range l {
+		aux = append(aux, el.String())
+	}
+	return strings.Join(aux, ",")
+}
+
+// Dig returns list of tuple <IP address, A record > for a given FQDN by using the dns servers from edgeDNSServers
+// dns servers are tried taking always two of them (setting one as a primary and the other as a backup one)
```
kuritka, line 49: not sure here, the result is the faster one, see https://github.com/lixiangzhong/dnsutil/blob/e61fde722b0a86b89243c686efc8c32fac5223d0/dig.go#L127-L129

kuritka (changes requested):

controllers/internal/utils/dns_test.go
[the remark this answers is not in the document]
jkremser: added both

controllers/internal/utils/dns.go
jkremser: I've changed the implementation to something that looks close to yours, but using internally the `Exchange` function in the same file. It's being tried 1-by-1 from the slice -> algorithm should be easier to read

kuritka: @jkremser , thx,  I didn't check, but please clean `go.mod `from dependency on `github.com/lixiangzhong/dnsutil` if you didn't already

jkremser: gone, what about `terratest/utils/extensions.go:345` and `terratest/go.mod`, currently it depends on `github.com/AbsaOSS/gopkg v0.1.2` so it has the old/current implementation of Dig.. I assume it's kind of a chicken-egg problem, it needs to depend on a released version, but first this change needs to go in and only then we can bump the implementation also in the `terratest/utils/extensions.go`

kuritka: terratest is technically different project. After checking dependencies I see:
```
# github.com/lixiangzhong/dnsutil
k8gbterratest/utils
github.com/AbsaOSS/gopkg/dns
github.com/lixiangzhong/dnsutil
```
^^^ there is no direct dependency to `github.com/lixiangzhong/dnsutil`. 

Creating an issue in github.com/AbsaOSS/gopkg package.  We can move implementation there in the future.

controllers/internal/utils/dns.go
```diff
@@ -20,30 +20,69 @@ package utils
...
+		return nil, fmt.Errorf("empty edgeDNSServers, provide at least one")
 	}
-	a, err := dig.A(fqdn)
+	if len(fqdn) == 0 {
+		return
+	}
+
+	if !strings.HasSuffix(fqdn, ".") {
+		fqdn += "."
+	}
```
kuritka, line 59: thx a ton, utils/dns package looks good. please extend tests with fqdn having `.` suffix

jkremser: done, 12 new test cases btw.

kuritka (changes requested):

kuritka:
@jkremser , is PR name `s/edgeDNSServer/edgeDNSServers/g` relevant ?

controllers/depresolver/depresolver_config.go
```diff
@@ -257,13 +343,13 @@ func parseLogOutputFormat(value string) LogFormat {
 func (c *Config) GetExternalClusterNSNames() (m map[string]string) {
 	m = make(map[string]string, len(c.ExtClustersGeoTags))
 	for _, tag := range c.ExtClustersGeoTags {
-		m[tag] = getNsName(tag, c.DNSZone, c.EdgeDNSZone, c.EdgeDNSServer)
+		m[tag] = getNsName(tag, c.DNSZone, c.EdgeDNSZone, c.EdgeDNSServers[0].Host)
 	}
 	return
 }
```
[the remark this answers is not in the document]
kuritka: EdgeDNS server has no influence on NS name (see [this](https://github.com/k8gb-io/k8gb/blob/master/controllers/depresolver/depresolver_config.go#L295-L297)), maybe it would be enough to add validation that would look for EDGE_DNS_SERVERS can be only 127.0.0.1 or localhost, or server list (except 127.0.0.1, localhost)

controllers/depresolver/depresolver_config.go
```diff
@@ -257,13 +343,13 @@ func parseLogOutputFormat(value string) LogFormat {
...
 
 func (c *Config) GetClusterNSName() string {
-	return getNsName(c.ClusterGeoTag, c.DNSZone, c.EdgeDNSZone, c.EdgeDNSServer)
+	return getNsName(c.ClusterGeoTag, c.DNSZone, c.EdgeDNSZone, c.EdgeDNSServers[0].Host)
 }
```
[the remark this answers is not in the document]
kuritka: EdgeDNS server has no influence on NS name (see [this](https://github.com/k8gb-io/k8gb/blob/master/controllers/depresolver/depresolver_config.go#L295-L297)), maybe it would be enough to add validation that would look for EDGE_DNS_SERVERS can be only 127.0.0.1 or localhost, or server list (except 127.0.0.1, localhost)

controllers/depresolver/depresolver_config.go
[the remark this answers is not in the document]
k0da: It would simply pick the first one and never fallback

controllers/depresolver/depresolver_config.go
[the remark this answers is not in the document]
k0da: @kuritka GetClusterNSName should return name of NS record to be created in external dns provider

controllers/depresolver/depresolver_config.go
```diff
@@ -55,9 +57,26 @@ const (
...
+	"EDGE_DNS_SERVER": NewVar{
+		Name: "EDGE_DNS_SERVERS",
+		Msg:  "Pass the hostname or IP address as comma-separated list",
+	},
+	"EDGE_DNS_PORT": NewVar{
+		Name: "EDGE_DNS_SERVERS",
+		Msg:  "Port is an optional in the comma-separated list of dns edge servers, in following form: dns1:53,dns2",
+	},
+}
+
```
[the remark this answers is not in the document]
jkremser: > OldVar is practically n ot used, map[string]NewVar

it's not, but it's more readable this way, imho... if anyone sees the type it's quite obvious that it's a mapping from old variable name to the new variable name

> Also NewVar, OldVar are not exposed, so can be private

good point

> EDGE_DNS_SERVER,EDGE_DNS_PORT, should be constants as it is defined in the rest of the package & EDGE_DNS_SERVER is used twice

whatever, I am putting them back, I wanted to prepare the stuff for their removal in the future. That's how deprecation logic should work in general, if you think it needs to be there forever, I can add them back and group them together under some deprecated comment.. or under their own const block

> but can't we just define public function retreiving deprecated stuffs

nah, this approach would print the message even though the deprecated env var was not being used. There must be some data (map, slice whatever) saying this particular env var is deprecated and we need to check if it's set by the user. I guess, I can move it to the `findDeprecations` function itself, but when called multiple times it will be doing the same stuff over and over again (not big problem in our context, because it's called only once during the startup), like this:

```golang
func (dr *DependencyResolver) findDeprecations(config *Config) (deprecations []string) {
	type oldVar = string
	type newVar struct {
		Name string
		Msg  string
	}
	var deprecated = map[oldVar]newVar{
		"EDGE_DNS_SERVER": newVar{
			Name: "EDGE_DNS_SERVERS",
			Msg:  "Pass the hostname or IP address as comma-separated list",
		},
		"EDGE_DNS_PORT": newVar{
			Name: "EDGE_DNS_SERVERS",
			Msg:  "Port is an optional in the comma-separated list of dns edge servers, in following form: dns1:53,dns2",
		},
	}

	for k, v := range deprecated {
		if os.Getenv(k) != "" {
			deprecations = append(deprecations, fmt.Sprintf("'%s' has been deprecated, use %s instead. Details: %s", k, v.Name, v.Msg))

			// remove the following code once the EDGE_DNS_SERVER and EDGE_DNS_PORT are not supported
			if k == "EDGE_DNS_SERVER" {
				// old contract, let's transform it to the new api -> will be also validated
				edgeDNSHost := os.Getenv(k)
				edgeDNSPort, _ := env.GetEnvAsIntOrFallback("EDGE_DNS_PORT", 53)
				config.EdgeDNSServers = []utils.DNSServer{{Host: edgeDNSHost, Port: edgeDNSPort}}
			}
		}
	}
	return
}
```

I don't like this ^ It's long and I don't like mixing static data w/ behavior so I'd leave it under the `consts` because it's more readable this way, imho. Having a function returning the map would also work and calling it from `findDeprecations` but what would be the point if it's returning the same data over and over again.

controllers/depresolver/depresolver_config.go
```diff
@@ -67,8 +86,7 @@ func (dr *DependencyResolver) ResolveOperatorConfig() (*Config, error) {
 		dr.config.route53Enabled = env.GetEnvAsBoolOrFallback(Route53EnabledKey, false)
 		dr.config.ns1Enabled = env.GetEnvAsBoolOrFallback(NS1EnabledKey, false)
 		dr.config.CoreDNSExposed = env.GetEnvAsBoolOrFallback(CoreDNSExposedKey, false)
-		dr.config.EdgeDNSServer = env.GetEnvAsStringOrFallback(EdgeDNSServerKey, "")
-		dr.config.EdgeDNSServerPort, _ = env.GetEnvAsIntOrFallback(EdgeDNSServerPortKey, 53)
+		dr.config.EdgeDNSServers = parseEdgeDNSServers(env.GetEnvAsStringOrFallback(EdgeDNSServersKey, ""))
```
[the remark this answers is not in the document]
jkremser: This is actually my initial approach :) that's exactly what I have, but then I was thinking about the log messages for the user and the logger is not being initialized before the config (chicken egg) so I ended up setting the value in the `findDeprecations`, but I think you are right and we can use the findDeprecations (or `GetDeprecations`) merely for getting the messages and calling it on demand... instead of returning the 3-tuple 👍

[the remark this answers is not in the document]
jkremser: actually, not exactly, this is what I have there before:
```golang
		dr.config.EdgeDNSServers = parseEdgeDNSServers(env.GetEnvAsStringOrFallback(EdgeDNSServersKey,
				fmt.Sprintf("%s:%v", env.GetEnvAsStringOrFallback(EdgeDNSServerKey, ""),
				env.GetEnvAsStringOrFallback(EdgeDNSServerPortKey, "53"))))
```

> ^^ consider EdgeDNSServer, EdgeDNSServerPort private until deprecations will be removed (edgeDNSServer, edgeDNSServerPort)

nah, too messy. The api would be pretty confusing having a plural form, singular form... list of values in one thing, single value in the other... what should be used etc. definitely not going this route, it should be unified to the new version during the validation phase to make our lives easier

controllers/depresolver/depresolver_config.go
```diff
@@ -223,6 +269,46 @@ func parseMetricsAddr(metricsAddr string) (host string, port int, err error) {
...
+		if err != nil {
+			port = 53
+		}
+		r = append(r, utils.DNSServer{
+			Host: host,
+			Port: port,
+		})
+	}
+	return r
+}
```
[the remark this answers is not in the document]
jkremser: done in `TestResolveConfigWithMultipleEdgeDnsServersMalformed1` added also `TestResolveConfigWithMultipleEdgeDnsServersMalformed2`

controllers/depresolver/depresolver_config.go
```diff
@@ -223,6 +269,46 @@ func parseMetricsAddr(metricsAddr string) (host string, port int, err error) {
...
+				portStr = "53"
+			}
+		default: // ipv6
+			if net.ParseIP(chunk).To16() != nil {
+				host = chunk
+			} else {
+				host = chunk[:strings.LastIndex(chunk, ":")]
+				portStr = chunk[strings.LastIndex(chunk, ":")+1:]
+			}
+		}
```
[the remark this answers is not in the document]
jkremser: done

controllers/depresolver/depresolver_config.go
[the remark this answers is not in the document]
kuritka: Ok, I understand your reasoning, `EDGE_DNS_SERVER` is a dynamic variable and the deprecation list should be returned depending on how the individual variables are deprecated in the future.

`EDGE_DNS_SERVER`, `EDGE_DNS_SERVER_PORT` can be used as a before, and there is no need to delete them (to make thing deprecated, ).  This implies several things:
 - we won't have to deal with breaking change after all, at least not at the moment
 - we have added a new variable EDGE_DNS_SERVERS ; if EDGE_DNS_SERVER, EDGE_DNS_SERVER_PORT are filled in, we just extend EDGE_DNS_SERVERS with another value 
 - `findDeprecations` can be simplified a lot, like this:
```go
func (dr *DependencyResolver) findDeprecations(config *Config) (deprecations []string) {
	if dr.config.EdgeDNSServer != "" {
		deprecations = append(deprecations, fmt.Sprintf("%s has been deprecated, use %s instead .... ", EdgeDNSServerKey, EdgeDNSServerKey)
		deprecations = append(deprecations, fmt.Sprintf("%s has been deprecated, use %s instead ... ", EdgeDNSServerPortKey, EdgeDNSServerKey)
		}
	}
       // if something else is deprecated, deprecations list will be altered here...
return
}
```
 
`EDGE_DNS_SERVER, EDGE_DNS_SERVER_PORT, EDGE_DNS_SERVERS `- now we have three live envvars, we don't have breaking changes, maybe we need to change validation and tests a bit.


Usually, when something is deprecate din golang, we use `Deprecated comments`, see: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/fake/client.go#L65-L71

controllers/depresolver/depresolver_config.go
```diff
@@ -223,6 +278,44 @@ func parseMetricsAddr(metricsAddr string) (host string, port int, err error) {
...
+		if host != "" {
+			r = append(r, utils.DNSServer{
+				Host: host,
+				Port: port,
+			})
+		}
+	}
+	return r
+}
```
kuritka, line 317: consider input `"127.0.0.1:5 053"` goes to valid `127.0.0.1:53`. 

Parsing servers is very complex operation in comparison with the rest, maybe we can retrieve errors from parser function and quit resolving immediatelly; or pass returning err into validation funcion which fails as well, but failing logic will be on the single place ( validation chunks in directly in a  parser may help a lot). 

 `if error_from_parsing != nil;  fail error message.`


Nit: ipv6 dismissed, do we need switch/case syntax now ?

kuritka (changes requested):
please check my comments, 
 - depresolver implementation progress well
 - consider more EdgeDNSServersKey corner cases and cover by tests.

controllers/depresolver/depresolver_config.go
jkremser: this is actually called only if the value of that env variable passes the reg exp check, which will not for `127.0.0.1:5 053`. I am adding a new test case w/ the `127.0.0.1:5 053` input so that you can see that it is not considered as a valid input (test called `TestResolveConfigWithInvalidEdgeDnsServerPort2`)

kuritka: perefect, there is large space of possible valid/invalid inputs, maybe, rather than create new tests,  you can iterate all invalid values you have on mind, like this. But decision is yours. 

```
for _, invalid := range []string{"127.0.0.1:5 053", "565:11", "host.cloud.example.com:5353:21".....} {
     // acting
     //. assertion
} 
```

controllers/depresolver/depresolver_config.go
```diff
@@ -55,6 +57,32 @@ const (
...
+var deprecated = map[oldVar]newVar{
+	EdgeDNSServerKey: newVar{
+		Name: EdgeDNSServersKey,
+		Msg:  "Pass the hostname or IP address as comma-separated list",
+	},
+	EdgeDNSServerPortKey: newVar{
+		Name: EdgeDNSServersKey,
+		Msg:  "Port is an optional in the comma-separated list of dns edge servers, in following form: dns1:53,dns2",
+	},
+}
```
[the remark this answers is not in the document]
jkremser: I didn't want the function to do it over and over again, but I guess that would be my cognitive baggage from my previous job on graalvm compiler (premature optimisation, yada yada..) I've learnt that package scoped vars are evil so I will move that to that function as u suggested because in our case, the function is called only once and even if it's called bazzilion times, I assume the golang compiler can do inlining and escape analysis these days. 

As for the suggested new form: I need to also keep track of the "old form" -> "new form" so that user is informed about what is the remedy for the deprecation (check `depresolver_config.go:L264`).

controllers/depresolver/depresolver_config.go
```diff
@@ -67,8 +95,9 @@ func (dr *DependencyResolver) ResolveOperatorConfig() (*Config, error) {
 		dr.config.route53Enabled = env.GetEnvAsBoolOrFallback(Route53EnabledKey, false)
 		dr.config.ns1Enabled = env.GetEnvAsBoolOrFallback(NS1EnabledKey, false)
 		dr.config.CoreDNSExposed = env.GetEnvAsBoolOrFallback(CoreDNSExposedKey, false)
-		dr.config.EdgeDNSServer = env.GetEnvAsStringOrFallback(EdgeDNSServerKey, "")
-		dr.config.EdgeDNSServerPort, _ = env.GetEnvAsIntOrFallback(EdgeDNSServerPortKey, 53)
+		dr.config.EdgeDNSServers = parseEdgeDNSServers(env.GetEnvAsStringOrFallback(EdgeDNSServersKey,
+			fmt.Sprintf("%s:%v", env.GetEnvAsStringOrFallback(EdgeDNSServerKey, ""),
+				env.GetEnvAsStringOrFallback(EdgeDNSServerPortKey, "53"))))
```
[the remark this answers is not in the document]
jkremser: right, in this case the `EdgeDNSServersKey` takes the precedence which is imho reasonable behavior

controllers/depresolver/depresolver_config.go
jkremser: done

kuritka (approved):

kuritka:
👍

controllers/depresolver/depresolver_config.go
```diff
@@ -257,13 +350,13 @@ func parseLogOutputFormat(value string) LogFormat {
 func (c *Config) GetExternalClusterNSNames() (m map[string]string) {
 	m = make(map[string]string, len(c.ExtClustersGeoTags))
 	for _, tag := range c.ExtClustersGeoTags {
-		m[tag] = getNsName(tag, c.DNSZone, c.EdgeDNSZone, c.EdgeDNSServer)
+		m[tag] = getNsName(tag, c.DNSZone, c.EdgeDNSZone, c.EdgeDNSServers[0].Host)
```
ytsarev, line 353: What will happen if first ([0]) edge dns server is not accessible?

jkremser: inside the `getNsName()` the dns server is used only in case it's equal to `localhost` or `127.0.0.1` (then it's returned -> dummy for tests) otherwise the dns server is ignored and the return value is constructed using `EDGE_DNS_ZONE`, geotag and what not.

After the validation it's guaranteed that there will be at least 1 element in that array so `c.EdgeDNSServers[0].Host` can't fail.

kuritka: @ytsarev , @jkremser  - here is the discussion about the topic: https://github.com/k8gb-io/k8gb/pull/605#discussion_r709180510

controllers/depresolver/depresolver_config.go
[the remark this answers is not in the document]
ytsarev: edgedns server has a direct influence on NS name resolution. Sticking with only one defeats the idea of multiple edgeDNS Servers to rely on unless I miss something

controllers/depresolver/depresolver_config.go
```diff
@@ -212,6 +209,64 @@ func (dr *DependencyResolver) validateConfig(config *Config, recognizedDNSTypes
...
+	}
+
+	var deprecated = map[oldVar]newVar{
+		EdgeDNSServerKey: newVar{
+			Name: EdgeDNSServersKey,
+			Msg:  "Pass the hostname or IP address as comma-separated list",
+		},
+		EdgeDNSServerPortKey: newVar{
+			Name: EdgeDNSServersKey,
+			Msg:  "Port is an optional in the comma-separated list of dns edge servers, in following form: dns1:53,dns2",
```
ytsarev, line 258: does it default to `53` in case of `dns2`?  if yes, then the message is not totally clear

controllers/depresolver/depresolver_config.go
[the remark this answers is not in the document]
ytsarev: @jkremser actually i recalled the `getNSName` logic, it's fine to evaluate the first one for local scenario. Not sure how it is obvious for external guys though

ytsarev:
@jkremser one last thing, could you please highlight the places in code where we actually increasing reliability with the multiple edgedns servers? https://github.com/k8gb-io/k8gb/pull/605/files#diff-b93d6284f641cd9c665d0167b418250059f2ac5438acf1fe775001c13973a368R295 here and https://github.com/k8gb-io/k8gb/pull/605/files#diff-b93d6284f641cd9c665d0167b418250059f2ac5438acf1fe775001c13973a368R313 here? anywhere else?

controllers/depresolver/depresolver_config.go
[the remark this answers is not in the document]
jkremser: I've added the check that would fail if localhost or 127.0.0.1 appears on some other position in the list + test called `TestResolveConfigWithMultipleEdgeDnsServersLocalhostNotFirst`

controllers/depresolver/depresolver_config.go
jkremser: added better message here and also to `values.yaml`

jkremser:
> @jkremser one last thing, could you please highlight the places in code where we actually increasing reliability with the multiple edgedns servers? https://github.com/k8gb-io/k8gb/pull/605/files#diff-b93d6284f641cd9c665d0167b418250059f2ac5438acf1fe775001c13973a368R295 here and https://github.com/k8gb-io/k8gb/pull/605/files#diff-b93d6284f641cd9c665d0167b418250059f2ac5438acf1fe775001c13973a368R313 here? anywhere else?

- also utils.Exchange(..) called [here](https://github.com/k8gb-io/k8gb/pull/605/files#diff-b93d6284f641cd9c665d0167b418250059f2ac5438acf1fe775001c13973a368R277) but that's part of the dnsQuery function so impl detail
- utils.Dig's signature now uses the list as well. Dig is being called from tests, but also from:
  - [gslb.go:104](https://github.com/k8gb-io/k8gb/pull/605/files#diff-b93d6284f641cd9c665d0167b418250059f2ac5438acf1fe775001c13973a368R104)
  - [gslb.go:140](https://github.com/k8gb-io/k8gb/pull/605/files#diff-b93d6284f641cd9c665d0167b418250059f2ac5438acf1fe775001c13973a368R140)
  
and that should be it

ytsarev (approved):
Ok, I guess we can merge this epic PR :)

@jkremser impressive job, thanks!

@kuritka thanks a ton for careful review

@jeffhelps it's all your fault man :D

controllers/depresolver/depresolver_config.go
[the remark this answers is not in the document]
jkremser: right, the above example will pass the validation (as I've done it) and I am not sure if it should or should not be supported as a valid use-case. Either way I am not sure how the `getNSName` should behave in this case. Perhaps instead of passing the `c.EdgeDNSServers[0].Host` to that method, we can pass a boolean called `isLocal` or something similar with the intention of not doing the string concat logic in that method, but returning localhost instead... well and that boolean can be calculated as true iff the list contains only 1 element and the `element.Host == localhost / 127.0.0.1`

wdyt? that way we will still have the fallback mechanism for other methods (Dig, Exchange)

PR #617, opened 2021-09-13
Optimize kubelinter pipeline config

ytsarev:

### DIFF
--- a/.github/workflows/kube-linter.yaml
+++ b/.github/workflows/kube-linter.yaml
@@ -2,19 +2,13 @@ name: KubeLinter
 
 on:
   push:
-    paths-ignore:
-      - '**.md'
-      - '**.svg'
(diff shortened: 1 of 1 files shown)


kuritka (approved):

.github/workflows/kube-linter.yaml
```diff
@@ -2,19 +2,13 @@ name: KubeLinter
 
 on:
   push:
-    paths-ignore:
-      - '**.md'
```
somaritane, line 6: I think this was added for reason, if md and txt are not omitted - we might get unnecessary warnings

somaritane: Actually, I'm not correct, this config is not related to kube-linter settings.

somaritane (approved):
lgtm

Issue #423, opened 2021-04-02
core dns load balancer does not get the correct targetPort


raffaelespazzoli:
the loadbalancer service deployed by setting the option:
```
exposeCoreDNS: true
```
is configured with the target port statically set to 53.
In some systems this requires privileges permissions, so user will change this to a different port via the following config:
```
coredns:
  servers:
    - zone: .
    port: 5353
```
the load balancer service template be changed to pick this setting.

k0da:
@raffaelespazzoli atm we maintain CoreDNS service ourselves so coredns chart service settings have no effect

raffaelespazzoli:
@k0da that is exactly my point. I'm asking to make some changes to the chart to guarantee consistency.

k0da:
@raffaelespazzoli I think k8gb currently doesn't support running CoreDNS on non standard port. Unfortunately coredns doesn't support setting containerPorts separately. So we can't really support configuration you mentioned.

raffaelespazzoli:
as of now this prevents deploying k8gb on OpenShift. Maybe we should ask coredns to support this feature in their chart?

k0da:
Sure, once CoreDNS supports it we can adapt

PR #626, opened 2021-09-22
Run coredns on unpriveleged port

k0da:
Fixes https://github.com/k8gb-io/k8gb/issues/423

### DIFF
--- a/chart/k8gb/Chart.lock
+++ b/chart/k8gb/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: coredns
   repository: https://absaoss.github.io/coredns-helm
-  version: 1.15.2
-digest: sha256:454120d311ca4b67ecbf6a0d33ea3282b795d69b4bee7ba0eed2488949908927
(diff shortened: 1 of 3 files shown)


jkremser (changes requested):
could you pls try that the 

`make deploy-full-local-setup` still works as described in https://www.k8gb.io/docs/local.html ? I think we need to upgrade the ingress and service also here: 
- https://github.com/k8gb-io/k8gb/blob/master/deploy/ingress/nginx-ingress-values.yaml#L29
- https://github.com/k8gb-io/k8gb/blob/master/deploy/ingress/udp-services.yaml#L9

otherwise lgtm

k0da:
@jkremser terratest setup in CI spins up pretty much the same setup as `deploy-full-local-setup`. From ingress/service perspective it is still the same port 53, the change is Deployment runs on port 5353 and svc's targetPort specifies that. This is transparrent to the upper stack

jkremser (approved):

PR #629, opened 2021-09-29
Replace Infoblox.Override.FakeInfobloxEnabled by func IsFakeInfoblox() bool

kuritka:
I'd like to test the dns package more - currentlhy 20.4% coverage (of which infoblox is a part), and I'll use mock instead of FakeInfoblox class.

At the moment no test hits the fakeInfoblox section (https://github.com/k8gb-io/k8gb/blob/master/controllers/providers/dns/infoblox-ibclient.go#L41-L49).
To enter this section we use envar `FAKE_INFOBLOX` - in my opinion unnecessary. Entering the section (which is not technically used) is the only reason why we keep this variable.

I created a temporary function on Infoblox{}.IsFakeInfobloxEnabled() that says it is a fake infobox, instead of using a special variable in config.
 ATM I can't say on 100% but probably delete IsFakeInfobloxEnabled() completely in a following commit when I inject IBRConnection mock in controller tests instead.

### DIFF
--- a/controllers/depresolver/depresolver.go
+++ b/controllers/depresolver/depresolver.go
@@ -103,10 +103,8 @@ type Infoblox struct {
 	HTTPPoolConnections int
 }
 
-// Override configuration
-type Override struct {
-	// FakeInfobloxEnabled if true than Infoblox connection FQDN=`fakezone.example.com`; default = false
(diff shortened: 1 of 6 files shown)


kuritka:
will reopen after successfully mock Infoblox connection

PR #632, opened 2021-09-30
Fix #631: helm - remove new lines from resulting yaml when using conditionals

### LINKED ISSUE #631: Helm chart produces empty lines in yamls
When using `helm template ... chart/k8gb` the resulting list of yamls contains couple of empty lines. It's more an aesthetic thing, because semantically it's a valid format, but still. It's caused by the helm chart conditions and that missing dash symbols.

example:

```yaml
...
template:
    metadata:
      labels:
        name: k8gb
    spec:

      serviceAccountName: k8gb
      containers:
        - name: k8gb
          ...
          env:
              ...
            - name: DNS_ZONE
              value: cloud.example.com
            - name: RECONCILE_REQUEUE_SECONDS
              value: "30"




            - name: LOG_FORMAT
              value: "simple"
            - name: LOG_LEVEL
              value: "info"
            - name: NO_COLOR
              value: "true"
              ...
```

jkremser:
As described in [here](https://helm.sh/docs/chart_template_guide/control_structures/#controlling-whitespace), by adding the dash right after the double curly braces it will chomp whitespaces on the left (including new lines) so this should fix the unnecessary new lines in the rendered yamls (if the things are not enabled)

### DIFF
--- a/chart/k8gb/templates/external-dns/external-dns.yaml
+++ b/chart/k8gb/templates/external-dns/external-dns.yaml
@@ -1,4 +1,4 @@
-{{ if or .Values.ns1.enabled .Values.route53.enabled }}
+{{- if or .Values.ns1.enabled .Values.route53.enabled }}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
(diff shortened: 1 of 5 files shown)


kuritka (approved):

ytsarev (approved):
nice catch

k0da (approved):

PR #633, opened 2021-09-30
Make securityContext of the deployment fully configurable

jkremser:
context: aligning the logic w/ the coredns helm chart (upstream [version](https://github.com/coredns/helm/blob/70519f776a9175db2b429b7b39625262c273180d/charts/coredns/templates/deployment.yaml#L130:L132)), this will allow for more flexibility on platforms like OpenShift (changing fsGroup, selinux, seccomp profile and what not)

btw. if one wants to remove the `runAsUser: 1000` (which is set by default) from the resulting yamls, they can set the value explicitly to `null`

### DIFF
--- a/chart/k8gb/values.yaml
+++ b/chart/k8gb/values.yaml
@@ -29,12 +29,16 @@ k8gb:
     level: info # log level (panic,fatal,error,warn,info,debug,trace)
   splitBrainCheck: false
   metricsAddress: "0.0.0.0:8080"
-  runAsUser: 1000
+  securityContext: # For more options consult https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#securitycontext-v1-core
+    runAsNonRoot: true
+    readOnlyRootFilesystem: true
+    allowPrivilegeEscalation: false
+    runAsUser: 1000
 
 externaldns:
   image: k8s.gcr.io/external-dns/external-dns:v0.9.0
   interval: "20s"
-  securityContext:
+  securityContext: # For more options consult https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#securitycontext-v1-core
     fsGroup: 65534 # For ExternalDNS to be able to read Kubernetes and AWS token files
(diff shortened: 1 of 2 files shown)


chart/k8gb/values.yaml
```diff
@@ -29,12 +29,16 @@ k8gb:
...
+    runAsNonRoot: true
+    readOnlyRootFilesystem: true
+    allowPrivilegeEscalation: false
+    runAsUser: 1000
 
 externaldns:
   image: k8s.gcr.io/external-dns/external-dns:v0.9.0
   interval: "20s"
-  securityContext:
+  securityContext: # For more options consult https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#podsecuritycontext-v1-core
```
ytsarev, line 41: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#securitycontext-v1-core is actually more precise here

ytsarev (changes requested):

chart/k8gb/values.yaml
```diff
@@ -29,12 +29,16 @@ k8gb:
     level: info # log level (panic,fatal,error,warn,info,debug,trace)
   splitBrainCheck: false
   metricsAddress: "0.0.0.0:8080"
-  runAsUser: 1000
+  securityContext:
```
ytsarev, line 32: Not sure why are we making clarification comment below only for external-dns?

ytsarev (changes requested):

chart/k8gb/values.yaml
jkremser: good catch, the spec is almost the same :) also the link needs to be updated, because they will shutdown the old api docs eventually.. currently they host like 5 versions

k0da (approved):
LGTM

chart/k8gb/values.yaml
jkremser: fixed, it was by mistake

ytsarev (approved):

PR #634, opened 2021-09-30
Remove FAKE_INFOBLOX (2/3)

kuritka:
This is a follow-up PR to #630 (1/3), do not review until #630 is merged.

Everything is now being mocked up, there is no need to distinguish between local and production infoblox.
Bool switch no longer makes sense, removing

### DIFF
--- a/controllers/depresolver/depresolver.go
+++ b/controllers/depresolver/depresolver.go
@@ -104,12 +104,6 @@ type Infoblox struct {
 	HTTPPoolConnections int
 }
 
-// Override configuration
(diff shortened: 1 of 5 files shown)


ytsarev (changes requested):
Super happy to get rid of fake infoblox dead code, thanks :) It just looks like it should be rebased with #630 after its merge

ytsarev:
Ah, you mentioned it, sorry :)

ytsarev (approved):

k0da (approved):

PR #646, opened 2021-10-06
Fix terratest CVE-2021-41103, CVE-2020-27813, CVE-2020-26160

kuritka:
follow-up #639, gruntwork-io/terratest#1006
this is terratest, expecting fixing k8gb alerts as well  

CVE-2021-41103  github.com/containerd/containerd, only v1.4.11, v1.5.7 are patched
CVE-2020-27813  github.com/gorilla/websocket >= 1.4.1 is patched
CVE-2020-26160  github.com/dgrijalva/jwt-go, There is no patch available and users of jwt-go are advised to migrate to golang-jwt at version 3.2.1

Terratest package is the top of a tree that uses many packages in different versions that are often not updated. After github issued an alert, it turns out that some packages deep inside are referencing long unsupported packages.
In order to solve the problems with the alerts, I had to specify which versions of the packages to use and which ones should not be downloaded (excluded, their versions are in the go.mod files).
The final solution works well if the assumption that our tests are not dependent on older versions of containerd or azure is met. I assume that in the future the problems will be solved on the terratest package side


ytsarev (approved):
@kuritka thanks for taking care of that

PR #653, opened 2021-10-13
Add grafana including example dashboard for 'podinfoes'

jkremser:
```
make deploy-prometheus
make deploy-grafana
make demo
```

Grafana requires to do port-forwarding, perhaps in the future PR we could improve it a bit and open the port in k3d config

### DIFF
--- a/deploy/grafana/dashboard-cm.yaml
+++ b/deploy/grafana/dashboard-cm.yaml
@@ -85,32 +85,7 @@ data:
                 ]
               }
             },
-            "overrides": [
-              {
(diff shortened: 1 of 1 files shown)


jkremser:
I've added one more commit (the previous is the same) that removed pod name from the PR. The json was exported from the web ui, so I missed it, nice catch @kuritka 

I wanted to split it to make the review easier (old part has already been approved)

pls check also https://github.com/k8gb-io/k8gb/pull/653/commits/94c5105b47cb40bef1b9015f4c4b20e8538176f5

if we use squash and merge, this will end up as 1 commit

PR #671, opened 2021-10-20
Use PNG image as chart icon

k0da:
SVG image we used have media type text/xml. We should have either svg or
png type.


k0da:
@somaritane sorry, if I took this wrong, but your concern was a PNG size, this was addressed. If you're not ok with that I have no problem to open a PR with revert.

Issue #651, opened 2021-10-13
Gh action for pushing the up-to-date image of k8gb-demo-curl


jkremser:
Dockerfile:
https://github.com/k8gb-io/k8gb/blob/master/deploy/test-apps/curldemo/Dockerfile

... or setup the auto build feature on the dockerhub side with this path in the repo (if it's still there)

PR #709, opened 2021-10-30
Push image to docker hub #651

AugustasV:
Edited github CI yaml file to upload artifact into dockerhub during build stage.

Issue #651


ytsarev (changes requested):
Thanks a lot for the contribution 👍 Curldemo is not connected to main k8gb release so it's better to create a separate flow for it. Also `env.GITHUB_JOB` will probably create unrelated tag name for the image

.github/workflows/release.yaml
```diff
@@ -49,3 +49,11 @@ jobs:
...
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and push
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          context: ./deploy/test-apps/curldemo/
+          file: ./Dockerfile
+          push: true
+          tags: ${{ secrets.DOCKER_USER }}/k8gb:${{ env.GITHUB_JOB }}
```
ytsarev, line 59: should be different docker repo as well.
```suggestion
          tags: ${{ secrets.DOCKER_USER }}/k8gb-demo:${{ env.GITHUB_JOB }}
```

ytsarev (changes requested):

.github/workflows/curldemo.yaml
```diff
@@ -0,0 +1,45 @@
+name: curldemo
+
+on:
+  push:
+    tags:
+      - "v*"
```
ytsarev, line 6: It's not connected to main release, so it is better to execute as reaction to file patch change, see https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-including-paths

.github/workflows/curldemo.yaml
```diff
@@ -0,0 +1,45 @@
...
+      DOCKER_CLI_EXPERIMENTAL: "enabled"
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Get tag
+        id: get_tag
+        run: |
+          previous_tag=$(git tag --sort=v:refname | tail -2 | head -1)
+          echo "::set-output name=previous_tag::${previous_tag}"
```
ytsarev, line 21: Should not be required here as it is not connected with the release

.github/workflows/curldemo.yaml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.16.9
+      - name: Login to Dockerhub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USER }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: Run GoReleaser
```
ytsarev, line 31: no goreleaser required, we are not building any binary with curl demo

.github/workflows/release.yaml
```diff
@@ -48,4 +48,4 @@ jobs:
           version: latest
           args: release --rm-dist --release-notes=changes
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
ytsarev, line 51: please revert newline change

ytsarev (changes requested):
Please address suggested changes.

Please also use --signoff to satisfy DCO https://github.com/k8gb-io/k8gb/pull/709/checks?check_run_id=4060681862 .

We are CNCF project and it is a strong requirement.

jkremser (changes requested):

.github/workflows/curldemo.yaml
```diff
@@ -0,0 +1,45 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and push
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          context: ./deploy/test-apps/curldemo/
+          file: ./Dockerfile
+          push: true
+          tags: ${{ secrets.DOCKER_USER }}/k8gb-demo:latest
```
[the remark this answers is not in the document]
AugustasV: Yes I know that using latest is evil, which env variable you would recommend to use? I tried right now with  GITHUB_RUN_NUMBER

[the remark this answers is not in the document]
ytsarev: @jkremser `latest` is evil but imho exactly what we need for the trivial demo. Obviously open to any better suggestions but the main goal is to keep it simple.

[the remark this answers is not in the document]
jkremser: make sense, kets keep it simple

.github/workflows/release.yaml
ytsarev, line 51: Please keep new line around

.github/workflows/curldemo.yaml
```diff
@@ -0,0 +1,29 @@
...
+          username: ${{ secrets.DOCKER_USER }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: Build and push
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          context: ./deploy/test-apps/curldemo/
+          file: ./Dockerfile
+          push: true
+          tags: ${{ secrets.DOCKER_USER }}/k8gb-demo-curl:$GITHUB_RUN_NUMBER
```
ytsarev, line 29: Let's stick with `latest` as we agreed in https://github.com/k8gb-io/k8gb/pull/709#discussion_r740856251

ytsarev (changes requested):

.github/workflows/curldemo.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+jobs:
+  curldemo:
+    runs-on: ubuntu-20.04
+    env:
+      DOCKER_CLI_EXPERIMENTAL: "enabled"
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
jkremser, line 16:
```suggestion
```

nit: `fetch-depth: 0` [indicates](https://github.com/actions/checkout/blob/ec3a7ce113134d7a93b817d10a8272cb61118579/README.md?plain=1#L92) that all history, branches and tags will be fetched, but we are not doing any fancy git stuff here, so the default shallow fetch should be ok. If omitted the `1` is used

jkremser (changes requested):

jkremser (approved):

k0da (approved):
LGTM

ytsarev (approved):
@AugustasV looks great, thanks for all the fixes :)

PR #716, opened 2021-11-08
Fix upgrade-candidate

kuritka:
`upgrade-candidate` - the target for deploying the current version of
k8gb to local k3d clusters tried to deploy a local image that did not exist.
Here is the fix.

### DIFF
--- a/.github/workflows/upgrade-testing.yaml
+++ b/.github/workflows/upgrade-testing.yaml
@@ -46,11 +46,11 @@ jobs:
           cluster-name: "test-gslb2"
           args: -c k3d/test-gslb2.yaml
 
-      - name: K8GB deploy stable
-        run: make deploy-stable
+      - name: K8GB deploy stable version
+        run: make deploy-stable-version
 
-      - name: K8GB upgrade candidate
-        run: make upgrade-candidate
+      - name: K8GB deploy test version
+        run: make deploy-test-version
 
       - name: Terratest
         run: make terratest
--- a/Makefile
+++ b/Makefile
@@ -119,8 +119,8 @@ deploy-full-local-setup: ## Deploy full local multicluster setup (k3d >= 4.2.0)
 	$(call deploy-local-cluster,$(CLUSTER_GSLB1),$(CLUSTER_GSLB2),$(VERSION),,'k8gb/k8gb')
 	$(call deploy-local-cluster,$(CLUSTER_GSLB2),$(CLUSTER_GSLB1),$(VERSION),$(CLUSTER_GSLB2_HELM_ARGS),'k8gb/k8gb')
 
-.PHONY: deploy-stable
-deploy-stable:
+.PHONY: deploy-stable-version
+deploy-stable-version:
 	@echo "\n$(YELLOW) import $(CYAN)k8gb:$(STABLE_VERSION) $(YELLOW)to $(CYAN)$(CLUSTER_GSLB1), $(CLUSTER_GSLB2) $(NC)"
 	$(call deploy-local-cluster,$(CLUSTER_GSLB1),$(CLUSTER_GSLB2),$(STABLE_VERSION),,'k8gb/k8gb')
 	$(call deploy-local-cluster,$(CLUSTER_GSLB2),$(CLUSTER_GSLB1),$(STABLE_VERSION),$(CLUSTER_GSLB2_HELM_ARGS),'k8gb/k8gb')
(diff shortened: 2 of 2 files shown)


ytsarev (dismissed):

somaritane (dismissed):

ytsarev (commented):
It probably duplicates the build step in the pipeline

kuritka:
> It probably duplicates the build step in the pipeline

going to fix

k0da:
Yes.     
 - name: Build artifacts
        uses: goreleaser/goreleaser-action@v2
        with:
          version: latest
          args: release --rm-dist --skip-publish --skip-validate --snapshot
        env:

kuritka:
All green, can  be reviewed again...

Makefile
```diff
@@ -128,8 +128,8 @@ deploy-stable:
 	$(call list-running-pods,$(CLUSTER_GSLB1))
 	$(call list-running-pods,$(CLUSTER_GSLB2))
 
-.PHONY: upgrade-candidate
-upgrade-candidate: ## Upgrade k8gb to the test version on existing clusters
+.PHONY: deploy-semver
+deploy-semver: ## Upgrade k8gb to the test version on existing clusters
```
ytsarev, line 132:
```suggestion
deploy-test-version: ## Upgrade k8gb to the test version on existing clusters
```
I might be missing something but `deploy-semver` is a bit confusing target name here

Makefile
```diff
@@ -128,8 +128,8 @@ deploy-stable:
 	$(call list-running-pods,$(CLUSTER_GSLB1))
 	$(call list-running-pods,$(CLUSTER_GSLB2))
 
-.PHONY: upgrade-candidate
-upgrade-candidate: ## Upgrade k8gb to the test version on existing clusters
+.PHONY: deploy-semver
```
ytsarev, line 131:
```suggestion
.PHONY: deploy-test-version
```

.github/workflows/upgrade-testing.yaml
```diff
@@ -49,8 +49,8 @@ jobs:
       - name: K8GB deploy stable
         run: make deploy-stable
 
-      - name: K8GB upgrade candidate
-        run: make upgrade-candidate
+      - name: K8GB deploy semver
```
ytsarev, line 52:
```suggestion
      - name: K8GB deploy test version
```

.github/workflows/upgrade-testing.yaml
```diff
@@ -49,8 +49,8 @@ jobs:
...
+        run: make deploy-semver
```
ytsarev, line 53:
```suggestion
        run: make deploy-test-version
```
I suggest target renaming, please see below

ytsarev (changes requested):

Makefile
kuritka: The main idea for naming is we operate with `stable`[[1](https://github.com/k8gb-io/k8gb/blob/master/Makefile#L122)], [[2](https://github.com/k8gb-io/k8gb/blob/master/.github/workflows/upgrade-testing.yaml#L50)] [[3](https://github.com/k8gb-io/k8gb/blob/master/Makefile#L70)] and `semver` [[4](https://github.com/k8gb-io/k8gb/blob/master/Makefile#L67)] terms. and already use `stable`. I was trying to keep same semantics. 

Maybe  can rename both (deploy-stable, deploy-semver)

jkremser: Agreed, using the word `semver` in this context is quite misleading. Semantic versioning talks about the versioning scheme, when to raise each component of that `x.y.z` triplet. So even the `stable` actually deploys a version (latest released in dockerhub) that conforms the semver.

 The makefile also operate with word "candidate", this is much closer to what it is, imho. Or `test-version` or `latest` vs `latest-released`(=stable).

The version/container tag it actually produces looks like `x.y.z-gitsha` and from what I've seen it's pretty much the same semantics as the `x.y.z-SNAPSHOT` versions from the Java world

kuritka:
@ytsarev , @jkremser,   fixed

k0da:
It still duplicates build, GHA builds image alreadt

kuritka:
@k0da are you sure? `upgrade-testing.yaml ` runs exactly same code as before , just target renamed from `upgrade-candidate` to `deploy-test-version`.

`upgrade-candidate` now builds locally + still fits with all documentation.

ytsarev (approved):
lgtm, thanks!

k0da:
> @k0da are you sure? `upgrade-testing.yaml ` runs exactly same code as before , just target renamed from `upgrade-candidate` to `deploy-test-version`.
> 
> `upgrade-candidate` now builds locally + still fits with all documentation.

https://github.com/k8gb-io/k8gb/blob/master/.github/workflows/upgrade-testing.yaml#L29

k0da (approved):

PR #722, opened 2021-11-09
Multicluster setup (n > 2)

jkremser:
- change the Makefile targets so that they run in a for loop
- add tests for terratest for round-robin @ 3 clusters (failover is still pending)
- change the helm chart to be able to set up more than one hostAlias entry for a pod
- add simple script that generates the yaml configs for k3d clusters so that we are DRY

to verify the change locally i suggest:
```bash
# deploy 5 clusters
make deploy-full-local-setup CLUSTERS_NUMBER=5

# do some testing
for i in {0..20} ; do make test-round-robin | grep mess ; done

# cleanup
make destroy-full-local-setup CLUSTERS_NUMBER=5
```

another consequence of this PR is that when trying to incorporate a change in local helm chart (that hasn't been published yet), one can do `make deploy-full-local-setup CHART=./chart/k8gb`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,6 @@ chart/k8gb/charts/*.tgz
 
 # gh-pages | Jekyll generated files
 _site
+
+# generated k3d configs
+k3d/test-gslb[!1-3].yaml
--- a/Makefile
+++ b/Makefile
@@ -24,8 +24,11 @@ endif
 ###############################
 #		CONSTANTS
 ###############################
-CLUSTER_GSLB1 = test-gslb1
-CLUSTER_GSLB2 = test-gslb2
+CLUSTERS_NUMBER ?= 2
+CLUSTER_IDS = $(shell seq $(CLUSTERS_NUMBER))
+CLUSTER_NAME ?= test-gslb
+CLUSTER_GEO_TAGS ?= eu us cz af ru ap uk ca
+CHART ?= k8gb/k8gb
 CLUSTER_GSLB_NETWORK = k3d-action-bridge-network
 GSLB_DOMAIN ?= cloud.example.com
 REPO = absaoss/k8gb
@@ -33,7 +36,6 @@ VALUES_YAML ?= ""
 PODINFO_IMAGE_REPO ?= ghcr.io/stefanprodan/podinfo
 HELM_ARGS ?=
 K8GB_COREDNS_IP ?= kubectl get svc k8gb-coredns -n k8gb -o custom-columns='IP:spec.clusterIP' --no-headers
-CLUSTER_GSLB2_HELM_ARGS ?= --set k8gb.clusterGeoTag='us' --set k8gb.extGslbClustersGeoTags='eu' --set k8gb.hostAlias.hostnames='{gslb-ns-eu-cloud.example.com}'
 LOG_FORMAT ?= simple
 LOG_LEVEL ?= debug
 CONTROLLER_GEN_VERSION  ?= v0.7.0
@@ -52,6 +54,7 @@ DEMO_DELAY ?=5
 ifndef NO_COLOR
 YELLOW=\033[0;33m
 CYAN=\033[1;36m
+RED=\033[31m
 # no color
 NC=\033[0m
 endif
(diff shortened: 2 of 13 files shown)


kuritka (commented):
probably requires additional investigation, 
please consider native iterators (see example):

```Makefile
# search, based on value from first list retrieves value in second list
# l1=[a b c] l2=[A D E] -> search(b, l1,l2) -> D 
search = $(subst $1_,,$(filter $1_%, $(join $(addsuffix _,$2),$3)))

k3ds ?= $(shell seq -f 'k3d-%g.yaml' 1 10)
k3dgeo ?= us de za cz es sp nl ru cn eu
https ?= $(shell seq 8053 8063)

k3d-%.yaml:
	@echo anycommand $@ $(*) \
	$(call search,$@,$(k3ds),$(k3dgeo)) \
	$(call search,$@,$(k3ds),$(https))

.PHONY: k3dgen
k3dgen: $(k3ds)

```

```shell
❯ make k3dgen
anycommand k3d-1.yaml 1 us 8053
anycommand k3d-2.yaml 2 de 8054
anycommand k3d-3.yaml 3 za 8055
anycommand k3d-4.yaml 4 cz 8056
anycommand k3d-5.yaml 5 es 8057
anycommand k3d-6.yaml 6 sp 8058
anycommand k3d-7.yaml 7 nl 8059
anycommand k3d-8.yaml 8 ru 8060
anycommand k3d-9.yaml 9 cn 8061
anycommand k3d-10.yaml 10 eu 8062
```

k3d/test-gslb4.yaml
```diff
@@ -0,0 +1,25 @@
...
+    - agent[0]
+- port: 5056:53/udp
+  nodeFilters:
+  - agent[0]
+options:
+  k3d:
+    disableLoadbalancer: true
+  k3s: # options passed on to K3s itself
+    extraServerArgs: # additional arguments passed to the `k3s server` command; same as `--k3s-server-arg`
+      - --no-deploy=traefik,servicelb,metrics-server
```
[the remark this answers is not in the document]
jkremser: good point, I wanted to do it actually, but didn't want to complicate the makefile even further. It's done now and it's not that bad. So that .gitignore ignores all `/k3d/test-gslb*.yaml` except 1,2 and 3

Makefile
```diff
@@ -113,54 +123,87 @@ demo: ## Execute end-to-end demo
...
 
 .PHONY: deploy-stable-version
 deploy-stable-version:
-	@echo "\n$(YELLOW) import $(CYAN)k8gb:$(STABLE_VERSION) $(YELLOW)to $(CYAN)$(CLUSTER_GSLB1), $(CLUSTER_GSLB2) $(NC)"
-	$(call deploy-local-cluster,$(CLUSTER_GSLB1),$(CLUSTER_GSLB2),$(STABLE_VERSION),,'k8gb/k8gb')
-	$(call deploy-local-cluster,$(CLUSTER_GSLB2),$(CLUSTER_GSLB1),$(STABLE_VERSION),$(CLUSTER_GSLB2_HELM_ARGS),'k8gb/k8gb')
-
-	$(call list-running-pods,$(CLUSTER_GSLB1))
-	$(call list-running-pods,$(CLUSTER_GSLB2))
+	@for c in $(shell seq $(CLUSTERS_NUMBER)); do \
```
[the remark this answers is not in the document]
jkremser: amended

jkremser:
> please consider native iterators (see example):

I know about it, but I wanted to avoid 'dynamic targets' or pattern rules, or what is it called to make it simple. I like the way how you combine two strings together and then allow for "random access" in it so I've borrowed that for the geo tags to have it as one-liner (space separated) instead of those 10 lines and 10 different variables

.gitignore
```diff
@@ -123,3 +123,9 @@ _site
 # OLM
 /olm/bundle
 /olm/olm-bundle
+
+# generated k3d configs
+/k3d/test-gslb*.yaml
+!/k3d/test-gslb1.yaml
+!/k3d/test-gslb2.yaml
+!/k3d/test-gslb3.yaml
```
kuritka, line 131:
```suggestion
k3d/test-gslb[!1-3].yaml
```

kuritka (dismissed):

terratest/utils/test_settings.go
```diff
@@ -25,7 +25,10 @@ type TestSettings struct {
 	Port1           int
 	DNSServer2      string
 	Port2           int
+	DNSServer3      string
```
k0da, line 28: So basically this just fixes number of clusters to 3? You can't run more without modyfing terratest code?

jkremser: no, you can deploy as many as you like w/ `make deploy-full-local-setup CLUSTERS_NUMBER=7` but in terratests for round robin 3 clusters are created and some testing is done (start them all, then scale podinfo to 0 on one and check if the IPs of the other two clusters are correctly propagated to gslb's cr's status, etc.)

Makefile
```diff
@@ -112,59 +115,94 @@ demo: ## Execute end-to-end demo
...
-	@echo "\n$(YELLOW)import k8gb docker image to $(CYAN)$(CLUSTER_GSLB1), $(CLUSTER_GSLB2) $(NC)"
+	@echo "\n$(YELLOW)import k8gb docker image to all $(CLUSTERS_NUMBER) clusters$(NC)"
+
+	@for c in $(CLUSTER_IDS); do \
+		echo "\n$(CYAN)$(CLUSTER_NAME)$$c:$(NC)" ;\
+		k3d image import $(REPO):$(SEMVER)-amd64 -c $(CLUSTER_NAME)$$c ;\
+	done
+
+	@for c in $(CLUSTER_IDS); do \
+		$(MAKE) deploy-local-cluster CLUSTER_ID=$$c VERSION=$(SEMVER)-amd64 CHART='./chart/k8gb' ;\
```
somaritane, line 142: IIRC, we've tried to get rid of that practice of calling external `make` as it's hard to debug and decided to use custom functions as means of code reuse. Is there something preventing us from following this approach?

Makefile
```diff
@@ -112,59 +115,94 @@ demo: ## Execute end-to-end demo
...
+		$(MAKE) create-local-cluster CLUSTER_NAME=$(CLUSTER_NAME)$$c ;\
+	done
 
-	$(call deploy-local-cluster,$(CLUSTER_GSLB1),$(CLUSTER_GSLB2),$(VERSION),,'k8gb/k8gb')
-	$(call deploy-local-cluster,$(CLUSTER_GSLB2),$(CLUSTER_GSLB1),$(VERSION),$(CLUSTER_GSLB2_HELM_ARGS),'k8gb/k8gb')
-
-	@echo "\n$(YELLOW)Deploy test apps $(NC)"
-	$(call deploy-test-apps,$(CLUSTER_GSLB1))
-	$(call deploy-test-apps,$(CLUSTER_GSLB2))
+	$(MAKE) deploy-stable-version
```
somaritane, line 124: Same question here, why use external make call instead of function?

Makefile
jkremser: yes, good point, I didn't know that you were actually doing it before, couple of reasons:

1) I think it's actually easier to debug, because one can call `make foobar` directly from terminal with arbitrary parameters w/o having to introduce special makefile target just to verify if the function/macro works

2) named arguments vs positional arguments, good example is this:
  ```bash
  $(call deploy-k8gb-with-helm,$(CLUSTER_GSLB1),$(CLUSTER_GSLB2),$(SEMVER)-amd64,,'./chart/k8gb')
  ```
  
  vs just (as it's called in the PR)
  ```bash
  $(MAKE) deploy-k8gb-with-helm
  ```
  ok little bit of cheating, because stuff got passed implicitly as vars, but thats good actually.. no need to pass everything over and over again, only those things that are different than defaults. Check like 129 for instance, here the `CHART` is not "overiden" and the default (`k8gb/k8gb`) is taken.

3) most importantly the `$(call foobar,arg1,arg2)` do not work reliably in for loops. If makefile contains for loop or a conditional, the commands in the body need to be separated by some non-whitespace separator (you know, `;`, `&&`,`||`, etc.) but on the other hand when commands are listed one by one under makefile target, each command is executed in its own subshell and those commands are not separated.
With these two different context we have no reliable way to call the macro from both, the loop context (requires separator) and from the regular target line (no separator). There are [workarounds](https://stackoverflow.com/a/12605404) and I went down that route, but it would complicate stuff even more.

4) crossplane/upbound also do that and I found their Makefile machinery pretty good and easy to use, I think we should actually split our huge Makefile into smaller ones (one with functions only, one for CI perhaps, one for local development, observability targets can live on their own etc.) but that's different story. 
My first interaction with k8gb as an external guy was, after cloning the repo, going to the Makefile and trying to figure out what that `make deploy-full-local-setup` or `make test-round-robin` do and sometimes it would be actually use the kubectl command directly instead of hiding it behind Makefile target (introducing unnecessary level of indirection).
I can imagine how it was done, that you wanted to have the Makefile DRY so that it was encapsulated as a funciton and then referrenced from the local development tutorial, but I find much easier to follow what this does
  ```bash
  kubectl scale deployment frontend-podinfo -n test-gslb --replicas=0
  ```
  ..rather than inspecting Makefile and learning that `make start-test-app` -> `$(call testapp-set-replicas,0)` -> `kubectl scale deploy ..`. 
  btw. I didn't touch this part at all

Makefile
jkremser: `deploy-stable-version` was actually a "standalone target" even before (so it can't be a function), it's being called from `upgrade-testing.yaml`. What I did here was little bit of code de-duplication. The thing is that some targets needs to be self-contained from start to end (`deploy-full-local-setup` - one liner that should do everything including creation of k3d clusters) but some (`deploy-stable-version`) assume the cluster to be created (in our testing the cluster is created by k3d action)

it's a negative side-effect of mixing tasks that might be convenient for a user with targets that are required by CI. But if those two concerns live in the same file, this is a reusable (/"DRY") thing to do.

ytsarev (changes requested):
Concern:

The default number of clusters produced by Makefile is 2.

The hardcoded number of clusters in terratest is becoming 3 with this change.

 It will break local `make terratest` invocation when CLUSTER_NUMBER != 3

We probably should fail fast in `make terratest` target if CLUSTER_NUMBER !=3 with a clear message that tests are expecting 3 clusters to avoid confusion in future.

In the long run we might extend terratest suite to be more dynamic in this regards but it will be probably overkilling.

Makefile
somaritane: > 1. I think it's actually easier to debug, because one can call make foobar directly from terminal with arbitrary parameters w/o having to introduce special makefile target just to verify if the function/macro works
You can usually run `make -d` or `make --debug=FLAGS` to generate advanced debug tracing during make target run. When you reinvoke `make` from make targets, you are losing this setting (and tracing as result) unless you explicitly pass that option again.

somaritane: > most importantly the $(call foobar,arg1,arg2) do not work reliably in for loops. If makefile contains for loop or a conditional, the commands in the body need to be separated by some non-whitespace separator (you know, ;, &&,||, etc.) but on the other hand when commands are listed one by one under makefile target, each command is executed in its own subshell and those commands are not separated.

just a thought: can make's `foreach` function help here? https://www.gnu.org/software/make/manual/html_node/Foreach-Function.html#Foreach-Function

somaritane: > ok little bit of cheating, because stuff got passed implicitly as vars, but thats good actually.. no need to pass everything over and over again, only those things that are different than defaults. 

IMHO explicit is better than implicit magic in this case, because env vars are globals and it's hard to trace sometimes what implicit value they have at the moment of call.

jkremser: > You can usually run make -d or make --debug=FLAGS...

These actually propagate correctly to recursive calls, check:
Makefile:
```makefile
foo:
	echo "hello from foo"
	echo $(MAKEFLAGS)
	$(MAKE) bar

bar:
	echo "hello from bar"
	echo $(MAKEFLAGS)
```

`make -d foo`
both targets will be run w/ that `-d` option and both will echo `-d`
Some options are not [propagated](https://www.gnu.org/software/make/manual/html_node/Options_002fRecursion.html#Options_002fRecursion):
> "The options ‘-C’, ‘-f’, ‘-o’, and ‘-W’ are not put into MAKEFLAGS; these options are not passed down."

but this is pretty edge case to me and it can be controlled by altering that `MAKEFLAGS` var

jkremser: >  can make's foreach function help here?

it can, I've actually had a link to the [workaround](https://stackoverflow.com/a/12605404) in the original comment and I was doing it at the beginning, but it was too complicated wrapping it in this one-liner. Also only 1 type of command separator can be used here

jkremser:
> We probably should fail fast in make terratest target if CLUSTER_NUMBER !=3 with a clear message that tests are expecting 3 clusters to avoid confusion in future.

it's there

ytsarev (approved):
thanks for the fixes, LGTM

somaritane (approved):

jkremser:
Thanks for reviews!

terrascan & CodeQL checks are being stucked, but these were passing before, so merging..

chart/k8gb/.helmignore
```diff
@@ -20,3 +20,6 @@
 .idea/
 *.tmproj
 .vscode/
+
+LICENSE
+README.md
```
somaritane, line 25: @jkremser was late to spot, does this mean that the k8gb helm chart packed and published by next release won't have associated license and readme included in here https://artifacthub.io/packages/helm/k8gb/k8gb ?

jkremser: thanks for catching this, it's fixed by https://github.com/k8gb-io/k8gb/pull/761

PR #723, opened 2021-11-10
Increase the default timeout for Infoblox http requests to 60 sec

jkremser:
Because recently the 20 seconds was not enough

### DIFF
--- a/chart/k8gb/values.yaml
+++ b/chart/k8gb/values.yaml
@@ -87,7 +87,7 @@ infoblox:
   # -- use SSL
   sslVerify: true
   # -- Request Timeout in secconds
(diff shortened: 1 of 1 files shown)


chart/k8gb/values.yaml
```diff
@@ -87,7 +87,7 @@ infoblox:
   # -- use SSL
   sslVerify: true
   # -- Request Timeout in secconds
-  httpRequestTimeout: 20
+  httpRequestTimeout: 60
```
somaritane, line 90: @jkremser I don't think we should increase the default value because of some local cases, this value can be overridden in a custom values file.

somaritane (changes requested):

PR #725, opened 2021-11-10
Switch version to the latest tag

k0da:
There are cases when Chart version is bumped but tag is is not yet set,
this breaks semantics of preparing a release and relying on the data is
not true yet. E.g. appVersion have no docker image pushed yet.


kuritka (approved):

ytsarev (changes requested):
Please provide well rounded reasoning.

ytsarev (approved):

PR #726, opened 2021-11-10
switch hostAlias to real edgeDNS

k0da:
This runs rfc2136 provider and remove all local hacks


deploy/edge/zone.yaml
```diff
@@ -0,0 +1,33 @@
...
+      };
+    };
+  k8s.zone: |
+    $TTL 30
+    @         IN SOA  example.com. root.example.com. (
+                                    16         ; serial
+                                    30         ; refresh (1 minute)
+                                    30         ; retry (1 minute)
+                                    30         ; expire (1 minute)
+                                    30         ; minimum (1 minute)
```
jkremser, line 30: nit: 30s  != 1 min

in the worst case, the TTL will be 30(bind)+30(core dns) = 60, right? 

those `cloud.example.com` records will be created by DNSendpoint CRs?

k0da: Yes

Makefile
```diff
@@ -209,6 +215,9 @@ deploy-k8gb-with-helm:
 		--set k8gb.imageTag=${VERSION:"stable"=""} \
 		--set k8gb.log.format=$(LOG_FORMAT) \
 		--set k8gb.log.level=$(LOG_LEVEL) \
+		--set rfc2136.enabled=true \
+		--set k8gb.edgeDNSServers[0]=$(CLUSTER_GSLB_GATEWAY):1053 \
+		--set externaldns.image=absaoss/external-dns:rfc-ns1 \
```
jkremser, line 220: what were the changes in `absaoss/external-dns:rfc-ns1`, where is it being built (/dockerfile)? Is it the upstream repo + this PR https://github.com/kubernetes-sigs/external-dns/pull/2439 ?

Makefile
```diff
@@ -30,6 +30,7 @@ CLUSTER_NAME ?= test-gslb
 CLUSTER_GEO_TAGS ?= eu us cz af ru ap uk ca
 CHART ?= k8gb/k8gb
 CLUSTER_GSLB_NETWORK = k3d-action-bridge-network
+CLUSTER_GSLB_GATEWAY = $(shell docker network inspect ${CLUSTER_GSLB_NETWORK} | jq -r .[].IPAM.Config[0].Gateway)
```
jkremser, line 33: this adds the dependency on `jq`, we should probably mention it somewhere otherwise it would fail silently. Well, `CLUSTER_GSLB_GATEWAY` would contain some err msg

chart/k8gb/templates/operator.yaml
```diff
@@ -15,10 +15,6 @@ spec:
       labels:
         name: k8gb
     spec:
-      {{- if .Values.k8gb.hostAliases }}
-      hostAliases:
-        {{- toYaml .Values.k8gb.hostAliases | nindent 8 }}
-      {{- end }}
```
jkremser, line 21: Can't we leave this here and also in `values.yaml` as an alternative option? These still can co-exist, right? 0 support on the "makefile level", but leaving the option there would be a safe thing to do, imho

jkremser (changes requested):
pls, leave the support for host aliases in helm

We should probably update the docs later on (local playground needs to be changed/improved to make podinfo working in the browser) and also mention somewhere that it's not meant to be used in production, unless the credentials are changed, like in the [original tutorial](https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/rfc2136.md#server-credentials).

otherwise looks good to me. I've tested the branch locally by:

```
make deploy-full-local-setup CLUSTERS_NUMBER=2 CHART='./chart/k8gb'
# CHART='./chart/k8gb' is required here, because the changes in helm haven't been released yet

dig @localhost -p 1053 roundrobin.cloud.example.com +short
172.19.0.8
172.19.0.7
172.19.0.3
172.19.0.4
172.19.0.5
172.19.0.6
k scale deployment frontend-podinfo -n test-gslb --replicas 0 --context k3d-test-gslb1

dig @localhost -p 1053 roundrobin.cloud.example.com +short
172.19.0.8
172.19.0.7
172.19.0.5
172.19.0.6

k scale deployment frontend-podinfo -n test-gslb --replicas 0 --context k3d-test-gslb2
dig @localhost -p 1053 roundrobin.cloud.example.com +short
172.19.0.8
172.19.0.7

k scale deployment frontend-podinfo -n test-gslb --replicas 1 --context k3d-test-gslb1
dig @localhost -p 1053 roundrobin.cloud.example.com +short
172.19.0.3
172.19.0.4
172.19.0.5
172.19.0.6
```

good stuff @k0da 👍

chart/k8gb/templates/operator.yaml
k0da: I'd like to drop fall back from controller as well, right now it is quite confusing

Makefile
k0da: Yes, exactly. This is built from branch make build.mini

Makefile
k0da: Added check for `CLUSTER_GSLB_GATEWAY` emptynes

somaritane: No need for `jq` dependendcy here, it can be solved with `docker` `--format` option, i.e.:
```suggestion
CLUSTER_GSLB_GATEWAY = $(shell docker network inspect ${CLUSTER_GSLB_NETWORK} -f '{{ (index .IPAM.Config 0).Gateway }}')
```

Makefile
```diff
@@ -30,6 +30,7 @@ CLUSTER_NAME ?= test-gslb
 CLUSTER_GEO_TAGS ?= eu us cz af ru ap uk ca
 CHART ?= k8gb/k8gb
 CLUSTER_GSLB_NETWORK = k3d-action-bridge-network
+CLUSTER_GSLB_GATEWAY = $(shell docker network inspect ${CLUSTER_GSLB_NETWORK} -f '{{ (index .IPAM.Config 0).Gateway }}')
```
somaritane, line 33: In case of nonexistent `k3d-action-bridge-network` this evaluation will fail:
```
❯ make                  
Error: No such network: k3d-action-bridge-network
```
I think it makes sense to have it as text for a shell command and postpone the evaluation until the make target.

k0da: @somaritane fixed

jkremser (dismissed):

k0da:
rebased against master to include k3d-action@v2 and drop temporary k3d hack

Makefile
```diff
@@ -30,6 +30,7 @@ CLUSTER_NAME ?= test-gslb
 CLUSTER_GEO_TAGS ?= eu us cz af ru ap uk ca
 CHART ?= k8gb/k8gb
 CLUSTER_GSLB_NETWORK = k3d-action-bridge-network
+CLUSTER_GSLB_GATEWAY = docker network inspect ${CLUSTER_GSLB_NETWORK} -f '{{ (index .IPAM.Config 0).Gateway }}'
```
kuritka, line 33: Where is CLUSTER_GSLB_NETWORK created ?

k0da: by k3d

kuritka (approved):
👍  nice job!

jkremser (approved):
👍

PR #732, opened 2021-11-15
Update docs about automatic tag during release process + different token

jkremser:
Updating docs.

[Previously](https://github.com/k8gb-io/k8gb/actions/runs/1462539270) the tag has been pushed, but didn't trigger the release probably because of [this](https://github.community/t/github-actions-workflow-not-triggering-with-tag-push/17053/2 ). So let's use a different token

edit: added one more thing - `contains` function was changed to `startsWith` because when reverting a commit, it takes the original msg and prepends "Revert" which would trigger the event (good catch @ytsarev)

### DIFF
--- a/.github/workflows/cut_release.yaml
+++ b/.github/workflows/cut_release.yaml
@@ -20,7 +20,7 @@ jobs:
           echo "::set-output name=new_tag::${new_tag}"
       - name: Push Tag
         uses: mathieudutour/github-tag-action@v5.6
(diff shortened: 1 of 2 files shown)


ytsarev (approved):

kuritka (dismissed):

ytsarev (approved):
LGTM

k0da (approved):

PR #738, opened 2021-11-15
Deploy test-gslb namespace only in local cluster

k0da:
We don't really need podinfo during terratest run in pipelines. To save
up some compute resources and CI run time.


ytsarev:
Just to clarify - we actually need podinfo but it is getting deployed by terratest itself into the dynamically created namespace, correct?

ytsarev (commented):
Just to clarify - we actually need podinfo but it is getting deployed by terratest itself into the dynamically created namespace, correct?

k0da:
Yes, deploy-test-app moved to deploy-full-local-setup. Terratesst helm installs podinfo on its own

ytsarev (approved):

PR #740, opened 2021-11-15
cleaning .gitignore

kuritka:
please try the local branch to see if everything works as expected.
Because things in .gitignore were repetitive and didn't make much sense, I refactored .gitignore.

- We stayed in denyList mode.
- ~~It has two sections~~
  - ~~the first section is generated by [toptal.com](https://www.toptal.com/developers/gitignore) for vim,go,goland,visualstudiocode,emacs,jekyll~~
  - ~~the second section follows the first one and is for custom rules of our project.~~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,125 +1,23 @@
-
-# Binaries for programs and plugins
-*.exe
-*.exe~
-*.dll
-*.so
-*.dylib
-bin
-
-# Test binary, build with `go test -c`
-*.test
-
-# Output of the go coverage tool, specifically when used with LiteIDE
-*.out
-
-# Kubernetes Generated files - skip generated files, except for vendored files
-
-!vendor/**/zz_generated.*
-
-# editor and IDE paraphernalia
-.idea
-*.swp
-*.swo
-*~
-
-# Temporary Build Files
-build/_output
-build/_test
-# Created by https://www.gitignore.io/api/go,vim,emacs,visualstudiocode
-### Emacs ###
-# -*- mode: gitignore; -*-
-*~
-\#*\#
-/.emacs.desktop
-/.emacs.desktop.lock
-*.elc
-auto-save-list
-tramp
-.\#*
-# Org-mode
-.org-id-locations
-*_archive
-# flymake-mode
-*_flymake.*
-# eshell files
-/eshell/history
-/eshell/lastdir
-# elpa packages
-/elpa/
-# reftex files
-*.rel
-# AUCTeX auto folder
-/auto/
-# cask packages
-.cask/
 dist/
-# Flycheck
-flycheck_*.el
-# server auth directory
-/server/
-# projectiles files
-.projectile
-projectile-bookmarks.eld
-# directory configuration
-.dir-locals.el
-# saveplace
-places
-# url cache
-url/cache/
-# cedet
-ede-projects.el
-# smex
-smex-items
-# company-statistics
-company-statistics-cache.el
-# anaconda-mode
-anaconda-mode/
-### Go ###
-# Binaries for programs and plugins
-*.exe
-*.exe~
-*.dll
-*.so
-*.dylib
-# Test binary, build with 'go test -c'
-*.test
-# Output of the go coverage tool, specifically when used with LiteIDE
 *.out
-### Vim ###
-# swap
-.sw[a-p]
-.*.sw[a-p]
-# session
-Session.vim
-# temporary
-.netrwhist
-# auto-generated tag files
-tags
-### VisualStudioCode ###
-.vscode/*
-.history
-# End of https://www.gitignore.io/api/go,vim,emacs,visualstudiocode
-
-### JetBrains IDE ###
-.idea
-# End of idea tools
-# Exclude chart tar.gz files
-chart/k8gb/charts/*.tgz
-
-# Ignore testing kubeconfigs
-kubeconfig*
-
-# Ignore changes from github_changelog_generator action
-changes
-
-# dotenv file
+*.sw[ap]
+k8gb
+k8gb.exe
 .env
 
-# gh-pages | Jekyll generated files
-_site
-.jekyll-metadata
+# IDE specific files (debug config etc...)
+.vscode/
+.idea/
+Session.vim
+.netrwhist
+tags
+
+# Helm charts
+chart/k8gb/charts/*.tgz
 
 # OLM
 /olm/bundle
 /olm/olm-bundle
+
+# gh-pages | Jekyll generated files
+_site


.gitignore
```diff
@@ -37,89 +11,195 @@ build/_test
...
 
-# gh-pages | Jekyll generated files
-_site
-.jekyll-metadata
+#### Custom section
+# put custom changes here
 
-# OLM
-/olm/bundle
-/olm/olm-bundle
```
k0da, line 125: This is needed when running olm-bundle locally

.gitignore
```diff
@@ -37,88 +11,199 @@ build/_test
...
+com_crashlytics_export_strings.xml
+crashlytics.properties
+crashlytics-build.properties
+fabric.properties
+
+# Editor-based Rest Client
+.idea/httpRequests
+
+# Ignores the whole .idea folder and all .iml files
+.idea/
```
jkremser, line 150: am I missing something? What's the point of lines 80-117 and 123, 132, 138, 147 and 153 if we then ignore the whole directory? Ignoring the whole `.idea` is ok btw, but doing it in a way like, hey ignore this file, this and this and btw. ignore everything (which is a super-set rule) looks odd.

.gitignore
```diff
@@ -37,88 +11,199 @@ build/_test
...
+atlassian-ide-plugin.xml
+
+# Cursive Clojure plugin
+.idea/replstate.xml
+
+# Crashlytics plugin (for Android Studio and IntelliJ)
+com_crashlytics_export_strings.xml
+crashlytics.properties
+crashlytics-build.properties
+fabric.properties
```
jkremser, line 144: not sure if we need to add stuff for Android Studio, Clojure plugin, Mongo Explorer, JIRA plugin and what not. Am I missing some hidden features of k8gb here? I assume you want to make it in sync with whatever the toptal.com produces for some pre-selected list of tools/IDEs that we use. I don't know, it seems too many entries to my taste, but if others use these tools, I am ok with it.

kuritka:
Good points @jkremser, gitignore in this case is generated by https://www.toptal.com/developers/gitignore (which was partly last time too). Only the end of the file are custom rules.

k0da:
just for curioucity, is it just IDE runs git add . when handling git? Never had a usecase for git ignore except goreleaser when it requires clean state to perform a release (but IIRC this only relates to tracked files)

somaritane (commented):
@kuritka I think it makes sense to follow the YAGNI (You Ain't Gonna Need It) principle here and leave only those filters that are relevant to the k8gb project nature and the tooling it uses. 
So I would not add auto-generated filters at all, but instead, even further remove unnecessary ones, if we talk about `.gitignore` cleaning.

ytsarev (changes requested):
I will second @somaritane here. Let's put only project-related entries into `.gitignore`

kuritka:
.gitignore optimized

.gitignore
```diff
@@ -1,125 +1,20 @@
...
 
-### JetBrains IDE ###
-.idea
-# End of idea tools
-# Exclude chart tar.gz files
+# Helm charts
 chart/k8gb/charts/*.tgz
 
-# Ignore testing kubeconfigs
-kubeconfig*
```
ytsarev, line 111: Can we please leave it around? I'm using it for testing over multiple clusters from time to time avoiding tampering the main kubeconfig

ytsarev (changes requested):

.gitignore
somaritane: @ytsarev This looks like a user-specific, not project-specific thing :) i.e. you can `export KUBECONFIG=` to point to any `kubeconfig` path of your like, shouldn't be necessary lying in the k8gb working copy. Or am I missing something?

.gitignore
```diff
@@ -1,125 +1,20 @@
...
-/eshell/lastdir
-# elpa packages
-/elpa/
-# reftex files
-*.rel
-# AUCTeX auto folder
-/auto/
-# cask packages
-.cask/
+vendor/
```
somaritane, line 1: are we still using Go vendoring?

kuritka: I subconsciously think someone might try. But we don't vendoring. 

Will ammend

.gitignore
```diff
@@ -1,125 +1,19 @@
...
 chart/k8gb/charts/*.tgz
 
-# Ignore testing kubeconfigs
-kubeconfig*
-
-# Ignore changes from github_changelog_generator action
-changes
-
-# dotenv file
-.env
```
somaritane, line 117: @kuritka we support env var overrides in local `.env` file for project testing and debugging purposes in `Makefile`:
https://github.com/k8gb-io/k8gb/blob/master/Makefile#L19

.gitignore
```diff
@@ -1,125 +1,19 @@
...
-
-# gh-pages | Jekyll generated files
-_site
```
somaritane, line 120: Not sure what happens if we switch from `gh-pages` branch back to `master` after k8gb.io site authoring. Local jekyll site run is dockerized now, but the working directory is mounted to container, so generated site resources are still in the mentioned dir.

.gitignore
ytsarev: @somaritane look like project creator specific thing yes ;) it is just two lines to speed up my productivity, I am not asking for too much, I think :)
Re kubeconfigs - I prefer to temporary keep them to the project under test, the matter of convenience

ytsarev: @somaritane following the same logic,  you could also use vim and we could drop `.vscode` and `.idea` ;)

.gitignore
kuritka: @somaritane , nice finding, I overlooked .env in makefile and took it as user specific, amending

.gitignore
ytsarev: Also kubeconfigs are somehow related to k8s operator project :D Sorry, now I will not give it up so easily :D

somaritane: @ytsarev I actually all for dropping IDE-specific files, they can simply reside in a personal `.gitignore` file.

ytsarev: @somaritane works for me, i will sacrifice my kubeconfigs for that :D

kuritka: I am also in favor of removing IDE specific settings, I left it there more because for most users this is too conflicting.

jkremser: pls keep `.idea` & `.vscode` also there, I have both in my global `.gitignore`, but still if it's not there it makes contributing (for external ppl w/ random editors) harder. The [chances](https://insights.stackoverflow.com/survey/2021#worked-with-vs-want-to-work-with-new-collab-tools-worked-want-prof) of using either vs code or idea are pretty high these days 

I don't see anything wrong w/ putting ide specific one-liners for IGNORING the configs (like `.idea`) what seemed weird to me were those `!.vscode/tasks.json` "keep-ide-specific-stuff-in-repo" rules

btw. I had to add those two specific rules to my `.gitignore` to be able to work comfortly w/ k8gb:

```
...
/.custom-env/*               ... custom env files that can be consumed by debug/run configurations in goland
/k8gb                        ... binary file we produce when doing `go build`
```

kuritka: just a reminder, if we leave the DenyListing .gitignore and switch to AllowListing .gitignore instead, we will only allow what is actually part of the project.

ytsarev (approved):
image
Looks like good cleanup. LGTM

.gitignore
```diff
@@ -1,125 +1,23 @@
...
-*.test
-# Output of the go coverage tool, specifically when used with LiteIDE
 *.out
-### Vim ###
-# swap
-.sw[a-p]
-.*.sw[a-p]
-# session
+*.sw[ap]
+k8gb
```
jkremser, line 4: ❤️

jkremser (approved):

.gitignore
somaritane: > The chances of using either vs code or idea are pretty high these days
@jkremser This statement is subjective without stats :).

I just hope that If we leave `.gitignore` for vim, idea, vscode, then we're ready to accept PRs to add emacs, sublime, atom, eclipse, komodo etc (sits pretty well in random editor category with Golang support available) on request in the future, denying such requests would be unfair. 

But as I said, if I have my IDE of choice, asking every single project to include it in its `.gitignore` won't scale.

somaritane (changes requested):
> btw. I had to add those two specific rules to my .gitignore to be able to work comfortly w/ k8gb:

@kuritka I think we should add `k8gb` to `.gitignore` because this is project-specific thing.

somaritane (approved):
@kuritka sorry, didn't see `k8gb` binary is already included.
Thanks a lot for the tremendous effort, LGTM! 👍

PR #745, opened 2021-11-16
Extend cut_release GHA workflow with release revert mechanism

ytsarev:
* Delete tag and release on associated revert commit
* Conditional expression fix/simplification according to
  https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idif

### DIFF
--- a/.github/workflows/cut_release.yaml
+++ b/.github/workflows/cut_release.yaml
@@ -14,8 +14,8 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - name: Get New Tag
-        id: get_new_tag
+      - name: Get Desired Tag
+        id: get_desired_tag
         run: |
           tag=$(awk '/appVersion:/ {print $2}' chart/k8gb/Chart.yaml)
           echo "Version to release: ${tag}"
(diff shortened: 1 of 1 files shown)


jkremser (approved):
looks good, but might be dangerous if someone deletes the tag manually and then also does the revert. That would delete the last but one release.

We might be paranoid and check if the version we are downgrading into in the Chart.yaml == git describe --abbrev=0 --tags ${tag}^ (previous tag in git)

Probably overkill, though + we should be able to catch this in the PR review

ytsarev:
@jkremser really good point, let me try to amend

ytsarev:
@jkremser amended, please check

k0da (dismissed):

.github/workflows/cut_release.yaml
```diff
@@ -12,17 +12,39 @@ jobs:
...
+        run: |
+          tag=$(git describe --tags --abbrev=0)
+          echo "Version to revert: ${tag}"
+          echo "::set-output name=tag::${tag}"
+      - name: Get Previous Tag
+        if: startsWith(github.event.head_commit.message, 'Revert "RELEASE:')
+        id: get_previous_tag
+        run: |
+          tag=$(git describe --tags --abbrev=0 $(git describe --tags --abbrev=0)^)
+          echo "Previos tag: ${tag}"
```
jkremser, line 43: nit: Previos

.github/workflows/cut_release.yaml
```diff
@@ -12,17 +12,39 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
-      - name: Get Tag
-        id: get_tag
+      - name: Get New Tag
```
jkremser, line 17: (nit) now that we can travel both: to the future and to the past, perhaps "desired tag" would better capture the intent

jkremser (changes requested):

ytsarev:
@jkremser make sense, amended

jkremser (approved):

k0da (approved):

kuritka (approved):

Issue #720, opened 2021-11-09
Deployment configuration for geotags


jkremser:
Currently we specify 
`operator:yaml:`
```yaml
            - name: CLUSTER_GEO_TAG
              value: {{ quote .Values.k8gb.clusterGeoTag }}
            - name: EXT_GSLB_CLUSTERS_GEO_TAGS
              value: {{ quote .Values.k8gb.extGslbClustersGeoTags }}
```
So for 5 cluster setup it can look like:
```
cluster 1:
CLUSTER_GEO_TAG: eu
EXT_GSLB_CLUSTERS_GEO_TAGS: us, cz, af, ap

cluster 2:
CLUSTER_GEO_TAG: us
EXT_GSLB_CLUSTERS_GEO_TAGS: eu, cz, af, ap

cluster 3:
CLUSTER_GEO_TAG: cz
EXT_GSLB_CLUSTERS_GEO_TAGS: eu, us, af, ap
..
```

This is not ideal because for each cluster `EXT_GSLB_CLUSTERS_GEO_TAGS` needs to be calculated (so that `CLUSTER_GEO_TAG` is not in the list). 

optimisation 1)
If we allowed the `EXT_GSLB_CLUSTERS_GEO_TAGS` to contain all the geo tags, including the "myself" (~`CLUSTER_GEO_TAG`), the deployment burden would be easier and we can easily perform this operation internally in the gslb controller (even check if `EXT_GSLB_CLUSTERS_GEO_TAGS ∈ CLUSTER_GEO_TAG` during validation phase)

optimisation 2)
Another optimisation that would make the life easier for those who deploy us would be inferring the `CLUSTER_GEO_TAG` from the Node's label/annotation. This shifts the problem to some other level, but it would allow for the same helm command (or the same crossplane XRD / whatever) to be run against each cluster.

optimisation 3)
[...]

PR #750, opened 2021-11-16
Allow the EXT_GSLB_CLUSTERS_GEO_TAGS to contain CLUSTER_GEO_TAG

jkremser:
It's addressing the first point in the [#720](https://github.com/k8gb-io/k8gb/issues/720) and basically we can ignore the value of `CLUSTER_GEO_TAG` if it's in the `EXT_GSLB_CLUSTERS_GEO_TAGS` list during the start of the application so that no other changes to code are needed (=> backward compatible change)

### DIFF
--- a/controllers/depresolver/depresolver_config.go
+++ b/controllers/depresolver/depresolver_config.go
@@ -82,6 +82,7 @@ func (dr *DependencyResolver) ResolveOperatorConfig() (*Config, error) {
 		fallbackDNS := fmt.Sprintf("%s:%v", dr.config.fallbackEdgeDNSServerName, dr.config.fallbackEdgeDNSServerPort)
 		edgeDNSServerList := env.GetEnvAsArrayOfStringsOrFallback(EdgeDNSServersKey, []string{fallbackDNS})
 		dr.config.EdgeDNSServers = parseEdgeDNSServers(edgeDNSServerList)
+		dr.config.ExtClustersGeoTags = excludeGeoTag(dr.config.ExtClustersGeoTags, dr.config.ClusterGeoTag)
 		dr.config.Log.Level, _ = zerolog.ParseLevel(strings.ToLower(dr.config.Log.level))
 		dr.config.Log.Format = parseLogOutputFormat(strings.ToLower(dr.config.Log.format))
 		dr.config.EdgeDNSType, recognizedDNSTypes = getEdgeDNSType(dr.config)
@@ -124,7 +125,7 @@ func (dr *DependencyResolver) validateConfig(config *Config, recognizedDNSTypes 
 	}
 	for i, geoTag := range config.ExtClustersGeoTags {
 		err = field(fmt.Sprintf("%s[%v]", ExtClustersGeoTagsKey, i), geoTag).
-			isNotEmpty().matchRegexp(geoTagRegex).isNotEqualTo(config.ClusterGeoTag).err
+			isNotEmpty().matchRegexp(geoTagRegex).err
 		if err != nil {
 			return err
 		}
@@ -319,6 +320,17 @@ func parseEdgeDNSServers(serverList []string) (r []utils.DNSServer) {
 		}
 	}
 	return r
+}
+
+// excludeGeoTag excludes the clusterGeoTag from external geo tags
+func excludeGeoTag(tags []string, tag string) (r []string) {
+	r = []string{}
+	for _, t := range tags {
+		if tag != t {
+			r = append(r, t)
+		}
+	}
+	return
 }
 
 // getEdgeDNSType contains logic retrieving EdgeDNSType.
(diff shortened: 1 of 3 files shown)


controllers/depresolver/depresolver_test.go
```diff
@@ -603,15 +603,39 @@ func TestResolveInvalidExtGeoTags(t *testing.T) {
...
 	defer cleanup()
 	tag := "us-west1"
 	expected := predefinedConfig
 	expected.ClusterGeoTag = tag
-	expected.ExtClustersGeoTags = []string{"us-east1", tag}
-	// act,assert
-	arrangeVariablesAndAssert(t, expected, assert.Error)
+	expected.ExtClustersGeoTags = []string{}
+	configureEnvVar(expected)
+	os.Setenv(ExtClustersGeoTagsKey, tag)
```
kuritka, line 614: `configureEnvVar(expected)` maps the expected entity to envvars, you can use something like this:

```go
	expected.ExtClustersGeoTags = []string{ExtClustersGeoTagsKey}
	configureEnvVar(expected)
```

Please can you change in all tests ?

controllers/depresolver/depresolver_test.go
```diff
@@ -603,15 +603,39 @@ func TestResolveInvalidExtGeoTags(t *testing.T) {
...
+func TestResolveGeoTagExistsWithinExtGeoTags(t *testing.T) {
+	// arrange
+	defer cleanup()
+	tag := "us-west1"
+	for _, arr := range [][]string{{"good-tag"}, {"us-east1", "eu"}} {
+		expected := predefinedConfig
+		expected.ClusterGeoTag = tag
+		expected.ExtClustersGeoTags = arr
+		configureEnvVar(expected)
+		os.Setenv(ExtClustersGeoTagsKey, strings.Join(append(arr, tag), ","))
```
kuritka, line 630:
```go
		expected.ExtClustersGeoTags = arr
		expected.ExtClustersGeoTags = append(expected.ExtClustersGeoTags, tag)
		configureEnvVar(expected)
```
 After resolve, config.ExtClusterGeoTags will contain ExtClusterGeoTag cleaned from ClusterGeoTag - which is value that is expected on this place.
 
 Probably assertion phase needs to be change / extended .

kuritka (changes requested):
good job, please do change in tests.

controllers/depresolver/depresolver_test.go
jkremser: I am not sure I get you correctly. In the assert phase we call deep equals on `expected` and `config`.

it's in the loop and
`config. ExtClustersGeoTags` will contain:
1. iter: `{"good-tag"}`
2. iter: `{"us-east1", "eu"}`
however on line 630 we override the os variable `EXT_GSLB_CLUSTERS_GEO_TAGS` to contain whatever is in the `config. ExtClustersGeoTags` + `"us-west1"`
`"us-west1"` is also `== config.ClusterGeoTag` so it should be excluded after config resolution so that `assert.Equal(t, expected, *config)` will hold

controllers/depresolver/depresolver_config.go
```diff
@@ -321,6 +322,17 @@ func parseEdgeDNSServers(serverList []string) (r []utils.DNSServer) {
 	return r
 }
 
+// parseExtClusterGeoTags ignores/excludes the clusterGeoTag from external geo tags
+func parseExtClusterGeoTags(tags []string, tag string) (r []string) {
```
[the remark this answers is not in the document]
jkremser: changed

controllers/depresolver/depresolver_test.go
jkremser: but the expected value should be an empty array, no? 
The idea is that we set the `EXT_GSLB_CLUSTERS_GEO_TAGS` to `tag `, `ClusterGeoTag` is also = `tag` => `config. ExtClustersGeoTags` should be empty. Test is quite artificial, perhaps I should change it in a way that at least one tag will remain in `ExtClustersGeoTags`, wdyt?

 same thing as it https://github.com/k8gb-io/k8gb/pull/750#discussion_r753044252

the current `configureEnvVar(expected) & NewDependencyResolver().ResolveOperatorConfig()` combo is not designed on this situation where we "change" the value during the validation process

kuritka: sorry, menat it this :

```go
	expected.ExtClustersGeoTags = []string{tag}
	configureEnvVar(expected)
```

controllers/depresolver/depresolver_test.go
kuritka: ok I see, was bit confused by setitng envvar twice 
```go
configureEnvVar(expected)
os.Setenv(ExtClustersGeoTagsKey, strings.Join(append(arr, tag), ","))
```

kuritka (approved):

PR #755, opened 2021-11-18
Add missing displayName + remove existing dir

jkremser:
missing `displayName: DNSEndpoint` was causing this:
Screenshot 2021-11-18 at 09 45 58

Deleting the directory in the GHA is there because when trying to modify the existing bundle version (bundleVersion = "master") it would copy the content to that existing directory. But we want it to be replaced.

### DIFF
--- a/chart/k8gb/Chart.yaml
+++ b/chart/k8gb/Chart.yaml
@@ -46,6 +46,7 @@ annotations:
     - kind: DNSEndpoint
       name: dnsendpoints.externaldns.k8s.io
       version: v1alpha1
+      displayName: DNSEndpoint
       description: Using ExternalDNS it synchronizes exposed Kubernetes Services and Ingresses with DNS providers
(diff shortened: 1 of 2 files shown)


chart/k8gb/Chart.yaml
```diff
@@ -46,6 +46,7 @@ annotations:
     - kind: DNSEndpoint
       name: dnsendpoints.externaldns.k8s.io
       version: v1alpha1
+      displayName: Gslb
```
k0da, line 49: Should be DNSEndpoint

k0da (changes requested):

chart/k8gb/Chart.yaml
jkremser: thanks, I've actually fixed this before opening the PR, but forgot to "push-f" again :D

k0da (approved):

Issue #712, opened 2021-11-02
Use pod's dnsConfig for our demo


jkremser:
Currently we [do](https://github.com/k8gb-io/k8gb/blob/v0.8.3/Makefile#L468):

```bash
"echo 'nameserver `$(K8GB_COREDNS_IP)`' > /etc/resolv.conf && wget "roundrobin.cloud.example.com"
```
The same goal can be achieved by using the dnsConfig under pod's spec:
https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-dns-config

PR #764, opened 2021-11-19
Drop resolv.conf hack

k0da:
This commit drops resolv.conf hack and uses POD's dnsConfig instead

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ REPO = absaoss/k8gb
 VALUES_YAML ?= ""
 PODINFO_IMAGE_REPO ?= ghcr.io/stefanprodan/podinfo
 HELM_ARGS ?=
-K8GB_COREDNS_IP ?= kubectl get svc k8gb-coredns -n k8gb -o custom-columns='IP:spec.clusterIP' --no-headers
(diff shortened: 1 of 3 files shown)


ytsarev (approved):
looks great

Makefile
```diff
@@ -478,9 +478,9 @@ kubectl config use-context $1 > /dev/null
 endef
 
 define hit-testapp-host
-	kubectl run -it --rm busybox --restart=Never --image=busybox -- sh -c \
-	"echo 'nameserver `$(K8GB_COREDNS_IP)`' > /etc/resolv.conf && \
-	wget -qO - $1"
+	kubectl run -it --rm busybox --restart=Never --image=busybox --command \
+	--overrides='{"spec": {"dnsConfig": {"nameservers": ["$(K8GB_COREDNS_IP)"]}, \"dnsPolicy\": \"None\"}}' \
+	-- wget $1
```
jkremser, line 483: `-qO -` is missing (we can use curl actually instead) (`-q` ... quiet (no wget stuff), `-O -` print to stdout)

jkremser (changes requested):

Makefile
k0da: @jkremser good catch! Amended

jkremser:
Fixes https://github.com/k8gb-io/k8gb/issues/712 (nice change)

jkremser (approved):

PR #767, opened 2021-11-22
Invoke the OLM pipeline from release pipeline

jkremser:
manually tested on my fork: https://github.com/jkremser/k8gb/actions/workflows/test.yaml

this action will send [HTTP post](https://docs.github.com/en/rest/reference/actions#create-a-workflow-dispatch-event) to trigger the workflow dispatch type of pipeline. This way we will still have a way to run the pipeline manually to hot-fix issues in OLM's manifests (no need to make a release for this).

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -49,3 +49,9 @@ jobs:
           args: release --rm-dist --release-notes=changes
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Invoke workflow for OLM
+        uses: benc-uk/workflow-dispatch@v1
+        with:
+          workflow: OLM bundle and PR
+          token: ${{ secrets.CR_TOKEN }}
+          inputs: '{ "bundleVersion": "master" }'


.github/workflows/release.yaml
```diff
@@ -49,3 +49,9 @@ jobs:
           args: release --rm-dist --release-notes=changes
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Invoke workflow for OLM
+        uses: benc-uk/workflow-dispatch@v1
+        with:
+          workflow: OLM bundle and PR
+          token: ${{ secrets.CR_TOKEN }}
+          inputs: '{ "bundleVersion": "master" }'
```
k0da, line 57: shouldn't we have a `tag` here?

jkremser: both would work, if `master` is set, it will take the code from master (which is the same as the release tag). The current tag is not available anywhere and I didn't want to do something like [this](https://stackoverflow.com/a/58178121) or [this](https://github.com/k8gb-io/k8gb/blob/master/.github/workflows/cut_release.yaml#L31:L37) again. But I can add that if you think it can help with readability.

k0da (approved):
LGTM

PR #768, opened 2021-11-22
Makefile: surround vars with quotes in conditional expressions

jkremser:
[This](https://github.com/k8gb-io/k8gb/runs/4288253217?check_suite_focus=true#step:8:168) went in under the radar.

It produces those `/bin/sh: 1: [: =: unexpected operator
16:17` during the deployment. Even though, the logic is sound (by accident :D) it is not nice.

It happens when the variable is empty and then check like this is done:
```
@if [ = true ]; then echo 1 ; fi
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -155,7 +155,7 @@ create-local-cluster:
 
 .PHONY: deploy-local-cluster
 deploy-local-cluster:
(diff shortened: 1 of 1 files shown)


k0da (approved):

ytsarev (approved):

jkremser:
Thanks @k0da for finding this!

PR #769, opened 2021-11-23
Running all the tests on two clusters and only full-rr on 3 clusters

jkremser:
Making `TestFullRoundRobin` dynamic in a way that it can run the assertions properly for any number of clusters (any ~ [2,8])

In makefile we should pass the build tag `all` xor `rr_multicluster` based on the number of clusters that currently runs. `all` will run the all the tests (same as before). `rr_multicluster` will run only the TestFullRoundRobin that depends on `CLUSTERS_NUMBER` env variable (which defaults to `2`).

This way, we can also have a on dispatch GH action that can run the tests on-demand w/ the number of clusters as an input parameter of the action (not done in this PR).

Some data:
I've tried 4 attempts for these pipelines:
- https://github.com/k8gb-io/k8gb/actions/runs/1495842167
- https://github.com/k8gb-io/k8gb/actions/runs/1495842172
- https://github.com/k8gb-io/k8gb/actions/runs/1495842169

All 12 passed so I am assuming it's better than current master where it's pretty fragile.


terratest/test/k8gb_abstract_full_roundrobin_test.go
```diff
@@ -0,0 +1,120 @@
...
+	for _, f := range from {
+		if _, found := aux[f]; !found {
+			diff = append(diff, f)
+		}
+	}
+	return
+}
+
+// helper function for de-duplicating elements in slice (in O(n))
+func dedupe(s []string) (uniq []string) {
```
kuritka, line 111: please rename the function to something more meaningful  (e.g. in SQL wa have distinct keyword for that)

terratest/test/k8gb_abstract_full_roundrobin_test.go
```diff
@@ -0,0 +1,120 @@
...
+	"testing"
+)
+
+func AbstractTestFullRoundRobin(t *testing.T, n int) {
+	if n < 2 || n > 8 {
+		t.Logf("Use value of n that represents the number of clusters from interval [2,8]")
+		t.FailNow()
+	} else {
+		t.Logf(fmt.Sprintf("Running TestFullRoundRobin for %d clusters", n))
+	}
```
kuritka, line 33: avoid if else statements if you can, just invert condition and fail fast (e.g: [crossplane doc](https://github.com/crossplane/crossplane/blob/master/CONTRIBUTING.md#return-early)).

kuritka (changes requested):
`.github/workflows/terratest-more-clusters.yaml` will be executed simultaneousy with terratest.yaml workflow ?

terratest/test/k8gb_abstract_full_roundrobin_test.go
jkremser: right, the else branch is totally useless here :D (done)

terratest/test/k8gb_abstract_full_roundrobin_test.go
jkremser: changed to `distinct`

jkremser:
> .github/workflows/terratest-more-clusters.yaml will be executed simultaneousy with terratest.yaml workflow ?

y, that's how it's done in this PR. But I am open to also run it either on demand by dispatch action or nightly/"morningly" (cron), but as it was mentioned on the call, this way nobody will pay attention to the result and it may be hard to track the change that introduced some issue. Btw. it runs only `TestFullRoundRobin` nothing else.

terratest/test/k8gb_abstract_full_roundrobin_test.go
```diff
@@ -0,0 +1,119 @@
...
+package test
+
+import (
+	"fmt"
+	"github.com/stretchr/testify/require"
+	"k8gbterratest/utils"
+	"testing"
+)
+
+func AbstractTestFullRoundRobin(t *testing.T, n int) {
```
ytsarev, line 27: Do we need to export this function? Isn't it the same test package where it is accessed?

jkremser: right, changed to lower case

terratest/test/k8gb_lifecycle_test.go
```diff
@@ -1,3 +1,5 @@
+// +build basic all
```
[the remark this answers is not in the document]
jkremser: amended

.licignore
```diff
@@ -11,3 +11,6 @@
 /api/**
 main.go
 gslb_controller.go
+
+# Build tags in tests...
+/terratest/test/**
```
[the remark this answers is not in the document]
jkremser: hmm, yes, but currently it's 8/10 files that have a build tag on them in that dir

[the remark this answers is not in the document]
ytsarev: ok, i guess we can live with that

ytsarev (approved):
LGTM, thanks for the fixes!

.licignore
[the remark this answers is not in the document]
kuritka: I created [issue for that](https://github.com/AbsaOSS/golic/issues/4). Later we can set it up exactly as we need.

kuritka (approved):

PR #774, opened 2021-11-26
Reorder terratest Workflow

k0da:
Right now workflow creates GSLB awaits for Ingress to come up, only then
deploys test app.

This represents corner case, while common case is to have and app to be
deployed at the time exposure object is supposed to be created.

This commit reorders Workflow in a way, podinfo deployed first, gslb
created afterwards.

### DIFF
--- a/terratest/utils/extensions.go
+++ b/terratest/utils/extensions.go
@@ -164,21 +164,6 @@ func (w *Workflow) Start() (*Instance, error) {
 	k8s.CreateNamespace(w.t, w.k8sOptions, w.namespace)
 	w.state.namespaceCreated = true
 
-	// gslb
-	if w.settings.gslbResourcePath != "" {
-		w.t.Logf("Create ingress %s from %s", w.state.gslb.name, w.settings.gslbResourcePath)
-		k8s.KubectlApply(w.t, w.k8sOptions, w.settings.gslbResourcePath)
-		k8s.WaitUntilIngressAvailable(w.t, w.k8sOptions, w.state.gslb.name, 60, 1*time.Second)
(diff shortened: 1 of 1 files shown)


terratest/utils/extensions.go
```diff
@@ -211,6 +196,22 @@ func (w *Workflow) Start() (*Instance, error) {
...
+		w.t.Logf("Create ingress %s from %s", w.state.gslb.name, w.settings.gslbResourcePath)
+		k8s.KubectlApply(w.t, w.k8sOptions, w.settings.gslbResourcePath)
+		k8s.WaitUntilIngressAvailable(w.t, w.k8sOptions, w.state.gslb.name, 60, 1*time.Second)
+		ingress := k8s.GetIngress(w.t, w.k8sOptions, w.state.gslb.name)
+		require.Equal(w.t, ingress.Name, w.state.gslb.name)
+		w.settings.ingressName = w.state.gslb.name
+	}
+
+	// ingress
+	if w.settings.ingressResourcePath != "" {
```
somaritane, line 211: is this intentional "NOOP" or am I missing something?

k0da: Dunno, it was there so I'm leaving it as it is. Maybe @kuritka have an idea?

somaritane: Yep, good to check. To me, this looks like a dead code.
Also surprised it got skipped the linters somehow, staticcheck would definitely treat this as an empty branch:
https://staticcheck.io/docs/checks/#SA9003

somaritane (approved):
@k0da LGTM

kuritka (approved):

PR #778, opened 2021-11-30
.gitignoring file called 'changes' that's produced and consumed by goreleaser

jkremser:
This entry was removed from the .gitignore recently, but it needs to be there for the release process to finish w/o errors.

We create the file here: https://github.com/k8gb-io/k8gb/blob/master/.github/workflows/release.yaml#L27

and consume it within the same action couple of steps later here: https://github.com/k8gb-io/k8gb/blob/master/.github/workflows/release.yaml#L49

But if it's not ignored, the goreleaser fails on this issue: https://github.com/k8gb-io/k8gb/runs/4372363917?check_suite_focus=true#step:8:23
because the git is in dirty state (there are untracked files). Other workaround would be copying that file somewhere else on that filesystem, but it's imho better to `.gitignore` it as it was before.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,6 @@ _site
 
 # generated k3d configs
 k3d/test-gslb[!1-3].yaml
+
+# Required by goreleaser
+changes


kuritka (approved):

jkremser:
upgrade testing will be failing until the release is not reverted (the 0.8.5 tag is there in the git repo, but container image is not there on dockerhub)

PR #789, opened 2021-12-06
Use font awesome anchor

jkremser:
follow-up of https://github.com/k8gb-io/k8gb/pull/692#issuecomment-951822359

looks like this:

Screenshot 2021-12-06 at 15 12 33

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -26,6 +26,7 @@
     <meta name="msapplication-TileColor" content="#b91d47">
     <meta name="msapplication-config" content="/assets/images/browserconfig.xml">
     <link rel="stylesheet" href="{{ '/assets/css/style.css?v=' | append: site.github.build_revision | relative_url }}">
+    <link href="//maxcdn.bootstrapcdn.com/font-awesome/4.6.0/css/font-awesome.min.css" rel="stylesheet">
   </head>
   <body>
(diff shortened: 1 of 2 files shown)


_layouts/default.html
```diff
@@ -45,7 +46,7 @@ <h2 class="project-tagline">{{ page.description | default: site.description | de
     </header>
 
     <main id="content" class="main-content" role="main">
-      {% include anchor_headings.html html=content anchorBody="§" anchorClass="heading-anchor" anchorAttrs="aria-labelledby=\"%html_id%\"" headerAttrs="class=heading" %}
+      {% include anchor_headings.html html=content anchorClass="heading-anchor fa fa-anchor" anchorAttrs="aria-labelledby=\"%html_id%\"" headerAttrs="class=heading" %}
```
somaritane, line 49: I suggest to use `fa-link` symbol
image

instead of `fa-anchor`
image

Most documentation-centric themes and CMS are using "link" symbol for header anchors and it became a de-facto standard recognizable by users.

jkremser: well, both are better than `§`. I like the "anchor" anchor better b/c it's aligned w/ the k8s nomenclature but I guess we can be ~boring~ predictable in our docs ;) so changing..

somaritane (approved):

PR #792, opened 2021-12-09
Gracefully finishing deferred calls 

kuritka:
We recently solved how to exit k8gb and make sure all deferred calls are done. The solution was to run a defer at the very end of the program, which decided whether to run os.Exit(0) or os.Exit(1).

Although this solution works, it is still somewhat difficult to understand and quite error prone from the point of view that from the very start of the program, we have to maintain the state of how the program exits.

The following refactoring uses a different pattern - which is stateless and maight be potentionally accepted by the kubebuilder when the PR is created.

```go
package main

func main() {
	if err := run(); err != nil {
		os.Exit(1)
	}
}

func run() error {

	// Current main() code with logs <-- logging, initializing etc.. is here

	return nil
}

```

### DIFF
--- a/main.go
+++ b/main.go
@@ -50,10 +50,12 @@ func init() {
 }
 
 func main() {
-	var exitCode = 1
-	defer func() {
-		os.Exit(exitCode)
-	}()
+	if err := run(); err != nil {
+		os.Exit(1)
+	}
+}
+
+func run() error {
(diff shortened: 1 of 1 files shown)


ytsarev (approved):
Looks very clean 👍

k0da (approved):

PR #615, opened 2021-09-10
Fix helm linting error for coredns.serviceType

somaritane:
This PR fixes helm linting error for `coredns.serviceType` value when k8gb chart dependency (coredns) is not yet updated:
```
[ERROR] templates/: template: k8gb/templates/operator.yaml:116:18: executing "k8gb/templates/operator.yaml" at <eq "LoadBalancer" .Values.coredns.serviceType>: error calling eq: incompatible types for comparison
```
It also fixes `kube-linter` security context errors in the service-delete-hook helm template
Signed-off-by: Timofey Ilinykh <timofey.ilinykh@absa.africa>

PR #799, opened 2021-12-13
Fix coredns based exposure

ytsarev:
The issue was detected with AWS reference setup

* The condition was not working as expected
* It led to degraded state when we expected `k8gb-ns-*` DNSEndpoint
  IP addresses to be populated with exposed coredns service ip addresses
  Instead it was not detecting the rule and was populated by Ingress IPs
  creating non-obvious mess in the configuration.

### DIFF
--- a/chart/k8gb/templates/operator.yaml
+++ b/chart/k8gb/templates/operator.yaml
@@ -99,7 +99,7 @@ spec:
             - name: EXTDNS_ENABLED
               value: "true"
             {{- end }}
-            {{- if eq "LoadBalancer" ( quote .Values.coredns.serviceType ) }}
+            {{- if eq "LoadBalancer" quote .Values.coredns.serviceType }}
             - name: COREDNS_EXPOSED
               value: "true"
             {{- end }}


kuritka (approved):

k0da (approved):

somaritane:
Related to #615

somaritane (approved):
@ytsarev LGTM, but I think we have to find the way how to also make `kube-linter` happy about it as well (see #615). Right now the check is [useless](https://github.com/k8gb-io/k8gb/runs/4508833083?check_suite_focus=true#step:3:24) anyways

somaritane:
@ytsarev this code renders `COREDNS_EXPOSED` template properly and also doesn't cause kube-linter errors:

```
            {{- if eq "LoadBalancer" ( .Values.coredns.serviceType ) }}
```

jkremser:
weirdly enough, the `kube-lint` installed from brew behaves differently than the [one](https://github.com/stackrox/kube-linter/releases/download/0.2.5/kube-linter-darwin.tar.gz) downloaded from internets. Even though it's the same version. 

`{{- if eq "LoadBalancer" ( .Values.coredns.serviceType ) }}` this works w/ kube-lint from brew, but fails w/ the one that's used in the action


```
 ~/Downloads/kube-linter version
0.2.5
~/Downloads/kube-linter lint -v chart
Warning: failed to load object from chart/k8gb: failed to render: template: k8gb/templates/operator.yaml:102:19: executing "k8gb/templates/operator.yaml" at <eq "LoadBalancer" (.Values.coredns.serviceType)>: error calling eq: incompatible types for comparison
Warning: no valid objects found.
```
while brew one:
```
kube-linter version
0.2.5
kube-linter lint -v chart
Warning: failed to load object from chart/k8gb/templates/external-dns/rbac.yaml: failed to decode: Object 'Kind' is missing in '---'
KubeLinter 0.2.5

No lint errors found!
```

this expression works on both versions, but I can't check if it's ok in the aws context:
```
{{- if eq "LoadBalancer" ( toString .Values.coredns.serviceType ) }}
```

somaritane:
> this expression works on both versions, but I can't check if it's ok in the aws context:
> 
> ```
> {{- if eq "LoadBalancer" ( toString .Values.coredns.serviceType ) }}
> ```
@jkremser thanks for finding the working solution! I've checked the proper rendering with `helm template` and "aws-enabled" custom values yaml file.
I've made the related fix here: https://github.com/k8gb-io/k8gb/pull/801

PR #801, opened 2021-12-13
kube-linter pipeline and helm template fix

### LINKED ISSUE #800: dev: Fix kube-linter github workflow
Currently k8gb [kube-linter](https://github.com/k8gb-io/k8gb/blob/master/.github/workflows/kube-linter.yaml) produces [results](https://github.com/k8gb-io/k8gb/blob/master/.github/workflows/kube-linter.yaml) that are not on par with local `kube-linter` command execution:
```
Warning: failed to load object from chart/k8gb: failed to render: template: k8gb/templates/operator.yaml:102:19: executing "k8gb/templates/operator.yaml" at <eq "LoadBalancer" (.Values.coredns.serviceType)>: error calling eq: incompatible types for comparison
Warning: no valid objects found.
```
```
Warning: failed to load object from chart/k8gb/templates/external-dns/rbac.yaml: failed to decode: Object 'Kind' is missing in '---'
``` 

It should be corrected so that:
- it uses a recent version of kube-linter-action (1.0.4)
- it points to the correct chart repository (`chart\k8gb`)
- it generates an output format that can be visualized by GitHub (sarif?)
- it uses `.kube-linter.yaml` config format for both local and GH workflow experience

somaritane:
Fixes #800 

- Updated kuber-linter-actions to v1 (=v1.0.4)
- Fixed kube-linter version to 0.2.5
- Added support for [sarif analysis](https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/uploading-a-sarif-file-to-github)
- Fixed problematic helm templates

kudos to @jkremser for finding the [proper fix for `COREDNS_EXPOSED` env var rendering](https://github.com/k8gb-io/k8gb/pull/799#issuecomment-992956154)!

Co-authored-by: Yury Tsarev <yury.tsarev@absa.africa>

### DIFF
--- a/chart/k8gb/templates/external-dns/rbac.yaml
+++ b/chart/k8gb/templates/external-dns/rbac.yaml
@@ -1,5 +1,5 @@
----
 {{- if or .Values.ns1.enabled .Values.route53.enabled .Values.rfc2136.enabled }}
+---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
--- a/.github/workflows/kube-linter.yaml
+++ b/.github/workflows/kube-linter.yaml
@@ -15,8 +15,20 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Scan repo
+
+      - name: Create ../results directory for sarif files
+        shell: bash
+        run: mkdir -p ../results
+
+      - name: Scan k8gb chart
         id: kube-lint-repo
-        uses: stackrox/kube-linter-action@v1.0.2
+        uses: stackrox/kube-linter-action@v1
         with:
-          directory: chart
+          directory: chart/k8gb
+          version: 0.2.5
+          format: sarif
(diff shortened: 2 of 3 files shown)


kuritka (approved):

chart/k8gb/templates/external-dns/rbac.yaml
```diff
@@ -1,4 +1,3 @@
----
 {{- if or .Values.ns1.enabled .Values.route53.enabled .Values.rfc2136.enabled }}
```
ytsarev, line 1:
```suggestion
{{- if or .Values.ns1.enabled .Values.route53.enabled .Values.rfc2136.enabled }}
---
```

ytsarev (changes requested):
I think it is cleaner to move yaml separator under condition instead of total removal

somaritane:
> I think it is cleaner to move yaml separator under condition instead of total removal

Agree, amended :)

k0da (approved):

ytsarev (approved):

Issue #757, opened 2021-11-19
k8gb controller's pid fails to start correctly when deployed by OLM


jkremser:
repro steps:

```
k3d cluster create -c k3d/test-gslb1.yaml
curl -sL https://github.com/operator-framework/operator-lifecycle-manager/releases/download/v0.19.1/install.sh | bash -s v0.19.1
kubectl create -f https://operatorhub.io/install/k8gb.yaml
```

```
kubectl get pods -n operators
NAME                            READY   STATUS                       RESTARTS   AGE
k8gb-566cd87ddc-xhrzf           0/1     CreateContainerConfigError   0          3m52s
k8gb-coredns-755bb648dc-stgs4   1/1     Running
```

```
kubectl describe pod k8gb-566cd87ddc-xhrzf -n operators
...
  Warning  Failed     3m54s (x12 over 6m9s)  kubelet            Error: container has runAsNonRoot and image has non-numeric user (nonroot), cannot verify user is non-root (pod: "k8gb-566cd87ddc-xhrzf_operators(ba777db9-dca3-4924-a485-f8b9eeeda398)", container: k8gb)
```

jkremser:
Removing the `runAsNonRoot: true` from live deployment fixes the issue so we should make this field null when generating the manifests for OLM

PR #809, opened 2021-12-15
Put OLM bundle under release publish

k0da:

### DIFF
--- a/.github/workflows/helm_publish.yaml
+++ b/.github/workflows/helm_publish.yaml
@@ -28,3 +28,9 @@ jobs:
           helm repo add k8gb https://k8gb.io/
           helm repo update
           helm -n k8gb upgrade -i k8gb k8gb/k8gb --wait --create-namespace --version=$(make version)
+      - name: Invoke workflow for OLM
+        uses: benc-uk/workflow-dispatch@v1
+        with:
+          workflow: OLM bundle and PR
+          token: ${{ secrets.CR_TOKEN }}
+          inputs: '{ "bundleVersion": "master" }'
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -49,9 +49,3 @@ jobs:
           args: release --rm-dist --release-notes=changes
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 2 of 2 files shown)


somaritane (changes requested):
@k0da it probably makes sense to rename WF file to `publish_artifacts.yaml`, and WF to `Publish Artifacts` to reflect the WF contents,  as OLM bundle is not technically a helm chart.

k0da:
> @k0da it probably makes sense to rename WF file from `helm_publish` to `publish_artifacts.yaml`, and WF to `Publish Artifacts` to reflect the WF contents, as OLM bundle is not technically a helm chart.

Arguable, it just triggers other workflow which uses Helm as source for manifests. From my POV it still quite pure helm task.

somaritane (changes requested):
> > Arguable, it just triggers other workflow which uses Helm as source for manifests. From my POV it still quite pure helm task. 

@k0da what about changing the event type in `olm_pr.yaml` job from:
```
on:
  workflow_dispatch:
```
to
```
on:
  release:
    types:
      - published
```
This way it is triggered by release event in parallel with `help_publish`.

jkremser:
> what about changing the event type in olm_pr.yaml job from

we would lose a way to hot-fix existing bundle by manually triggering the action. I had to that for instance for [this issue](https://github.com/k8gb-io/k8gb/issues/757) using [this pipeline](https://github.com/k8gb-io/k8gb/runs/4253469608?check_suite_focus=true#step:8:8) (resulting [pr](https://github.com/k8s-operatorhub/community-operators/pull/420/files))

jkremser (approved):

somaritane (approved):

k0da:
> > > Arguable, it just triggers other workflow which uses Helm as source for manifests. From my POV it still quite pure helm task.
> 
> @k0da what about changing the event type in `olm_pr.yaml` job from:
> 
> ```
> on:
>   workflow_dispatch:
> ```
> 
> to
> 
> ```
> on:
>   release:
>     types:
>       - published
> ```
> 
> This way it is triggered by release event in parallel with `help_publish`.

Making it separate would release it in parallel, I think we still want to smoke test helm itself and only after use its artifacts to build OLM

PR #812, opened 2021-12-20
Upgrade licenses

kuritka:
- update `.licignore` to contain `/terratests`, `main.go`, and `gslb_controller.go`.
- remove old license from source code and terratests
- place license on the new place in the go files  (under package *\n)
- update build pipeline (`build.yml`)
- bump golic to `v0.7.2`
```shell
# golic@v0.5.0
golic remove -t apache2
# golic@v0.7.2
golic inject -t apache2
```


ytsarev (commented):
> place license on the new place in the go files (under package *\n)

curious about this change. Was it breaking something with the license header before `package`?

kuritka:
@ytsarev, thx for question.

A valid go file structure can be :
```go
 /* 
   optional comment
 */
 
 // Package xyz optional comment,  if it exists, it must start with 'Package xyz' otherwise linters may shout. e.g: idea goland always warns. 
 package xyz
 ...
``` 

the problem starts when go directives are introduced, which must be in a certain place in the file - for example at the beginning:
```go
//go:build !ignore_autogenerated
// +build !ignore_autogenerated

// ^^ must be on thbe top...
// Package xyz defines....
package xyz
```
In such case, finding a place to inject the licence is more difficult. 

finding a place for the license would be significantly harder, but not impossible:
 - we would have to specify files in `k8gb/.golic.yaml` that would generate the license under `//go:build` clauses (otherwise on top). With new files, we would have to edit the file from time to time. 
- Edit golic again and iterate through the comments preceding the package definition until a space is found between the comments and the GO directives.

ytsarev:
@kuritka that makes total sense, for the explanation

ytsarev (approved):

PR #814, opened 2021-12-21
Upgrade all GHA to checkout v2

ytsarev:


k0da (approved):

kuritka (approved):
It was so obvious that everyone overlooked it.

jkremser (changes requested):
CI failing

perhaps we need to do
```
with:
   fetch-depth: 0
```
so that this works properly: https://github.com/k8gb-io/k8gb/blob/master/Makefile#L69 

related: https://github.com/actions/checkout/issues/338

jkremser (approved):
👍

PR #818, opened 2021-12-28
helm: finer control over what is going to be deployed

jkremser:
* Respect the coredns values (`coredns.deployment.enabled` & `coredns.rbac.create`)
 * control whether crds (`k8gb.deployCrds`) and/or rbac (`k8gb.deployRbac`) for k8gb should be deployed

Main reason for this change is that I would like to add a terratest for a new feature I am working on (optimisation 2 described in https://github.com/k8gb-io/k8gb/issues/720) and currently it is not possible to deploy the k8gb AGAIN in the same cluster to a different namespace. I know that the behavior might be undef, because two controllers will be fighting against each other for the C~R~UD events. But I want to test only the startup of the operator and if it's able to recognize the labels on node and set the geo tag appropriately. Unfortunatelly, I can't do that w/ current helm chart, because some resources are created globaly in the default namespace (crd, cluster role and cluster role binding). This PR aims to make the deployment of these "global" resources optional. We choose not to be opinionated about the deployment mechanisms people can have so making it more configurable can't hurt.

### DIFF
--- a/chart/k8gb/templates/coredns-cm.yaml
+++ b/chart/k8gb/templates/coredns-cm.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.coredns.deployment.enabled }}
 kind: ConfigMap
 metadata:
   labels:
@@ -19,3 +20,4 @@ data:
             negttl {{ .Values.k8gb.dnsZoneNegTTL }}
         }
     }
+{{- end }}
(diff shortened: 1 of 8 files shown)


chart/k8gb/templates/coredns-cm.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if and .Values.coredns.deployment.enabled }}
```
ytsarev, line 1: why do we have `and` here?

jkremser: it worked, but it's a mistake, amended

ytsarev (approved):
LGTM

PR #823, opened 2022-01-04
Getting cluster geo tag from nodes labels and annotations

jkremser:
more details in issue #720 

Unfortunately, the `clusterGeoTag` was also used in helm chart for calculating the `extdnsOwnerID` which is then passed to `external-dns` deployment as an argument for the main binary/container. However we don't know it upfront so I had to add a logic around it. It (`external-dns` deployment) takes the param as the env variable from a configmap called `external-dns-env`. If the cm is not there it won't get deployed (-> no race conditions), we create this configmap after k8gb start with the correctly calculated value for `EXTERNAL_DNS_TXT_OWNER_ID` (same logic as before)

k8gb now has to be able to read the nodes in order to get the annotations or labels on them

I've created a dummy rbac for tests in just to be able to read the nodes as well (we are not using the global one called "k8gb" because there would be a risk that operator would be handling the gslb resources). In the terratest we deploy another k8gb instance using helm with crippled RBAC and test if it fails if geo tag is not provided by env var nor by label/annotations on the nodes and then put the label on nodes, kill the pod to enforce the restart and check that this time it's ok, since it gets the geo tag from a node.


terratest/utils/utils.go
```diff
@@ -268,3 +273,110 @@ func RunBusyBoxCommand(t *testing.T, options *k8s.KubectlOptions, dns string, co
...
+	if verb != "apply" && verb != "delete" {
+		require.Failf(t, "you should call the K8gbRBAC with verb = apply | delete, it was: '%s'", verb)
+	}
+	rbacPath, err := filepath.Abs("../examples/rbac-geo-tags-read-nodes.yaml")
+	require.NoError(t, err)
+	cmd := shell.Command{
+		Command: "bash",
+		Args:    []string{"-c", fmt.Sprintf("cat %s | envsubst | kubectl %s -n %s -f -", rbacPath, verb, options.Namespace)},
+		Env:     map[string]string{
+			"namespace": options.Namespace,
```
k0da, line 288: Do we controll k8gb namespace in tests?

jkremser: y, get's created & deleted [here](https://github.com/k8gb-io/k8gb/pull/823/files#diff-ff1e9de1e313a77323dc5ce09022081630f3f965935e1632db43b5193c38a43cR44). They will look like `k8gb-test-geo-tag-1fuk1x` where the last chunk is random

k0da: We can throw away `envsubs` and use .Release.Namespace in original RBAC and ommit having separate rbac resource here.

main.go
```diff
@@ -133,6 +137,12 @@ func run() error {
 	reconciler.DNSProvider = f.Provider()
 	log.Info().Str("provider", reconciler.DNSProvider.String()).Msg("Started")
 
+	err = inferClusterGeoTag(config, mgr.GetAPIReader())
+	if err != nil {
+		log.Err(err).Msg("Can't infer the CLUSTER_GEO_TAG")
```
kuritka, line 142:
```suggestion
log.Err(err).Msgf("Can't infer the %s", depresolver.ClusterGeoTagKey)
```

main.go
```diff
@@ -144,6 +154,46 @@ func run() error {
...
+		case len(node.Labels[region]) != 0:
+			operatorConfig.ClusterGeoTag = node.Labels[region]
+		}
+		if len(operatorConfig.ClusterGeoTag) != 0 {
+			return nil
+		}
+	}
+	if len(operatorConfig.ClusterGeoTag) == 0 {
+		return fmt.Errorf("%v nodes were tried, but none of them were annotated."+
+			" Either set the CLUSTER_GEO_TAG explicitly as operator conviguraion using env variable,"+
```
kuritka, line 195: `depresolver.ClusterGeoTagKey`

main.go
kuritka, line 142: In order to get as many input values as possible as soon as possible (and fail on invalid inputs soon as possible), would it be possible  to make the initialization (`inferClusterGeotag`) as on the top of `main()` as possible? Maybe  somewhere after the manager initialization.

kuritka (changes requested):
Hi overall excellent idea and good job!

I have only three little things mentioned in review.

somaritane (changes requested):
@jkremser nice work! 
This feature might help a lot to simplify the GitOps-centric k8gb deployment workflows, with one set of k8gb deployment settings valid for all the clusters.
There might be a few conceptual comments/questions we might want to address before merging.

main.go
```diff
@@ -133,6 +137,12 @@ func run() error {
 	reconciler.DNSProvider = f.Provider()
 	log.Info().Str("provider", reconciler.DNSProvider.String()).Msg("Started")
 
+	err = inferClusterGeoTag(config, mgr.GetAPIReader())
```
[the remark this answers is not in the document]
jkremser: the check is in the function itself, I wanted to make it concise in the main. check: https://github.com/k8gb-io/k8gb/pull/823/commits/863ed69a2c7891d2d98cc46b04445f89acc506b7#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R162

main.go
```diff
@@ -144,6 +154,46 @@ func run() error {
...
+	for _, node := range nodeList.Items {
+		switch {
+		case len(node.Annotations[zone]) != 0:
+			operatorConfig.ClusterGeoTag = node.Annotations[zone]
+		case len(node.Labels[zone]) != 0:
+			operatorConfig.ClusterGeoTag = node.Labels[zone]
+		case len(node.Annotations[region]) != 0:
+			operatorConfig.ClusterGeoTag = node.Annotations[region]
+		case len(node.Labels[region]) != 0:
+			operatorConfig.ClusterGeoTag = node.Labels[region]
```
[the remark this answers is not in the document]
jkremser: Added logic that handles this. So taking only region into account + they must be the same on each node, otherwise fail fast

main.go
jkremser: done, well calling from main now, but sooner

controllers/depresolver/depresolver.go
```diff
@@ -139,6 +139,9 @@ type Config struct {
 	MetricsAddress string `env:"METRICS_ADDRESS, default=0.0.0.0:8080"`
 	// extDNSEnabled hidden. EdgeDNSType defines all enabled Enabled types
 	extDNSEnabled bool `env:"EXTDNS_ENABLED, default=false"`
+	// Route53HostedZoneID identifier of route53 hosted zone that's added (if not empty)
+	// for external-dns deployment as part of the txt-owner-id
+	Route53HostedZoneID string `env:"ROUTE53_HOSTED_ZONE_ID"`
```
k0da, line 144: What about NS1 or rfc2136 provider, IIUC controller now expects R53 hosted zone ID and will fail if string matching pattern won't be provided.

jkremser: It can be empty, the regexp check is done only if it's not an empty string. So for NS1 or rfc2136 the env var is not going to be set.

If it's empty the DNS_ZONE value is used instead (same as before in the _helpers.tpl)

k0da: Maybe we can use same owner across providers? Not sure mainaining separate case for r53 worth it.

controllers/post_start.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (r *GslbReconciler) PostStartHook(operatorConfig *depresolver.Config, mgr ctrl.Manager) error {
+	if err := inferClusterGeoTag(operatorConfig, mgr.GetAPIReader()); err != nil {
+		log.Err(err).Msgf("Can't infer the %s", depresolver.ClusterGeoTagKey)
+		return err
+	}
+	if err := createOrUpdateExternalDNSConfigMap(operatorConfig, mgr.GetClient()); err != nil {
+		log.Err(err).Msgf("Can't create/update config map for external-dns")
+		return err
+	}
```
kuritka, line 40: I cannot fully evaluate if it would not be better from the maintenance point of view to always create configMap declaratively (somewhere in yaml configurations) with default value `fmt.Sprintf("%s-%s-%s", txtOwnerPrefix, cfg.DNSZone, cfg.ClusterGeoTag)`. And only if `cfg.Route53HostedZoneID` was set, we would update the txtOwnerKey. 

@team WDYT?

controllers/post_start.go
```diff
@@ -0,0 +1,115 @@
...
+func createOrUpdateExternalDNSConfigMap(cfg *depresolver.Config, c client.Client) error {
+	const (
+		ns             = "k8gb"
+		cmName         = "external-dns-env"
+		txtOwnerKey    = "EXTERNAL_DNS_TXT_OWNER_ID"
+		txtOwnerPrexix = "k8gb"
+	)
+	var txtOwnerValue string
+	if len(cfg.Route53HostedZoneID) != 0 {
+		txtOwnerValue = fmt.Sprintf("%s-%s-%s", txtOwnerPrexix, cfg.Route53HostedZoneID, cfg.ClusterGeoTag)
```
kuritka, line 92: please fix typo `prexix` in  `txtOwnerPrexix`

controllers/post_start.go
```diff
@@ -0,0 +1,115 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func (r *GslbReconciler) PostStartHook(operatorConfig *depresolver.Config, mgr ctrl.Manager) error {
+	if err := inferClusterGeoTag(operatorConfig, mgr.GetAPIReader()); err != nil {
+		log.Err(err).Msgf("Can't infer the %s", depresolver.ClusterGeoTagKey)
+		return err
+	}
+	if err := createOrUpdateExternalDNSConfigMap(operatorConfig, mgr.GetClient()); err != nil {
+		log.Err(err).Msgf("Can't create/update config map for external-dns")
```
somaritane, line 38: Probably, leftover (Msgf->Msg)

controllers/post_start.go
jkremser: In the case when geo tag is not provided upfront, we have it only after k8gb starts (reading it from the node labels), so we can't have it in yaml in helm chart.

why not to create the cm before external-dns start and update the key from the k8gb?
..because the if the env var in Deployment is [referenced](https://github.com/k8gb-io/k8gb/pull/823/files#diff-c5b1eb253f3b95c48e2f9203764d56752b157636d69930e79e5e7e27a59567d0R39) from a ConfigMap and that cm is not there yet, the deployment will wait for the cm to be created. So it solves the race/data condition when external-dns could do its job (reconcile the dnsendpoints) with wrong ownerId + updating the CM will not make the deployment that uses it to restart iirc, we would need to restart it anyway.

I am currently looking into an option to throw away the geo tag from the `txt-owner-id` as we talked about it and use the `k8gb-{{ .Values.k8gb.dnsZone }}-$uuid` where `$uuid` is generated during the helm install (w/ the option to override it using helm values).. the problem with this approach is that helm upgrade will regenerate the uuid (as I've pointed out during the call) so it's quite cumbersome to do that... helm install will give you some random id, then we have to use it during the helm upgrade ... not very user (/dev-ops) friendly

kuritka: @jkremser , ok, thanks for the explanation

ytsarev (changes requested):
This is breaking change, we need documentation for it

ytsarev:
@jkremser do you plan to continue working on this PR?

ytsarev:
Stale, we can return to it if there is a demand by community. Thanks for the implementation attemt!

PR #824, opened 2022-01-05
Always use the bash as the default shell

jkremser:
If the SHELL variable is not specified in the makefile, the env var is used if it exist. If not,
then the /bin/sh alias will be used. This is unpredictable because each OS can have different default
shell implementaion. For instance on Ubuntu it is dash and it behaves differently than bash.

I've tried this in ubuntu container:

```
foo:
        @$(eval RUNNING_CLUSTERS := $(shell echo 3))
        @$(eval TEST_TAGS := $(shell [ $(RUNNING_CLUSTERS) == 2 ] && echo all || echo rr_multicluster))
        @if [ "$(RUNNING_CLUSTERS)" -lt 2 ] ; then \
        echo "$(RED)Make sure you run the tests against at least two running clusters$(NC)" ;\
        exit 1;\
        fi
        echo "running clusters=$(RUNNING_CLUSTERS)"
```
`make foo`
```
/bin/sh: 1: [: 3: unexpected operator
echo "running clusters=3"
running clusters=3
```

when I prepended
`SHELL=bash` in Makefile or changed the `/bin/sh` symlink from dash to bash, the issue was gone

It's probably [a good thing to do](https://github.com/upbound/build/blob/master/makelib/common.mk#L18L:L19) to have a predictable environment on different "OSes"

Sorry that I introduced this issue and missed that in the CI output, this should fix https://github.com/k8gb-io/k8gb/runs/4705421875?check_suite_focus=true#step:9:6


somaritane (approved):
@jkremser LGTM!

kuritka:
Hi, I know it's a detail but wouldn be:
 
```Makefile
# Set the shell to bash always
SHELL := /bin/bash
```

jkremser:
> Hi, I know it's a detail but wouldn be:

`¯\_(ツ)_/¯` the lazy evaluation of `=` should not play any role here, right? Because on the right hand side we have a string and nothing dynamic, I was adapting to the style of other assignments in that file... for instance the `REPO = absaoss/k8gb` 1 line above. But you are probably right that `:=` denotes more something like hey this is not gonna be changed in the future no mater what's on the right side.. no lazy expansion fanciness :D ~ constant

so I can change both lines

kuritka:
it was about /bin/bash x bash rather than assignment. But I say it's just a detail. Anyway A or B, it's good that you added it.   
https://github.com/crossplanebook/provider-template/blob/master/Makefile#L1-L2

kuritka (approved):

Issue #825, opened 2022-01-05
Consider making terratest output readable


jkremser:
Because of the fact that terratests are run in parallel and produce a lot of log messages when doing the polling mechanisms (those `waitUntilX` functions), it's almost impossible to find out the root cause of the issue why some test is red.

Gruntwork provides a tool that produces separate log files for each test w/o any interleaving, it's called `terratest_log_parser` They also provide an example how to use it together with CircleCI. Here is an example of circle.ci config: https://github.com/gruntwork-io/kubergrunt/blob/master/.circleci/config.yml#L42:L54

They upload the separate log files into artifact repository that pertains to the job. GHA has something [similar](https://github.com/actions/upload-artifact)

links:
- https://terratest.gruntwork.io/docs/testing-best-practices/debugging-interleaved-test-output/
- https://github.com/actions/upload-artifact

AugustasV:
Will take a look into that

jkremser:
@AugustasV thanks!

PR #826, opened 2022-01-05
Terratest readable output

AugustasV:
#825 Updated make file and terratest files to output terratest results.
In that way it will be easier to debug.


.github/workflows/terratest-more-clusters.yaml
```diff
@@ -70,6 +70,14 @@ jobs:
...
-        run: make terratest
+        run: mkdir -p ${{ github.workspace }}/tmp/terratest && make terratest | tee ${{ github.workspace }}/tmp/terratest/all.log
+
+      - name: Terratest log parser
+        run: terratest_log_parser --testlog ${{ github.workspace }}/tmp/terratest/all.log --outputdir ${{ github.workspace }}/tmp/terratest
+
+      - uses: actions/upload-artifact@v2
+        with:
+          name: terratest-logs
+          path: ${{ github.workspace }}/tmp/terratest
```
ytsarev, line 83: https://thoughtbot.com/blog/no-newline-at-end-of-file#history-lesson please keep the newline at the end of file and remove whitespace characters

.github/workflows/terratest.yaml
```diff
@@ -62,5 +62,13 @@ jobs:
           echo "Cluster 2 (us):"
           kubectl get no -owide --context=k3d-test-gslb2
 
-      - name: Terratest
-        run: make terratest
+      - name: Run Terratests
```
ytsarev, line 65: Please keep consistent naming 'Run Terratests' vs 'Run Terratest' above

.github/workflows/terratest.yaml
```diff
@@ -62,5 +62,13 @@ jobs:
...
+        run: mkdir -p ${{ github.workspace }}/tmp/terratest && make terratest | tee ${{ github.workspace }}/tmp/terratest/all.log
+
+      - name: Terratest log parser
+        run: terratest_log_parser --testlog ${{ github.workspace }}/tmp/terratest/all.log --outputdir ${{ github.workspace }}/tmp/terratest
+
+      - uses: actions/upload-artifact@v2
+        with:
+          name: terratest-logs
+          path: ${{ github.workspace }}/tmp/terratest
```
ytsarev, line 74: please add newline character as above

Makefile
```diff
@@ -426,8 +426,11 @@ terratest: # Run terratest suite
 		echo "$(RED)Make sure you run the tests against at least two running clusters$(NC)" ;\
 		exit 1;\
 	fi
-	cd terratest/test/ && go mod download && CLUSTERS_NUMBER=$(RUNNING_CLUSTERS) go test -v -timeout 15m -parallel=12 --tags=$(TEST_TAGS)
-
+	mkdir -p /tmp/terratest && cd terratest/test/ && go mod download && CLUSTERS_NUMBER=$(RUNNING_CLUSTERS) go test -v -timeout 15m -parallel=12 --tags=$(TEST_TAGS) | tee /tmp/terratest/all.log
+	curl --location --silent --fail --show-error -o terratest_log_parser https://github.com/gruntwork-io/terratest/releases/download/v0.38.8/terratest_log_parser_linux_amd64
+	chmod +x terratest_log_parser
+	sudo mv terratest_log_parser /usr/local/bin
+	terratest_log_parser --testlog /tmp/terratest/all.log --outputdir /tmp/terratest
```
ytsarev, line 433: locally static `/tmp/terratest` can quickly get messy after multiple invocations. Should we include a randomized string in directory name?

ytsarev (changes requested):
@AugustasV thanks a lot for your contribution.

We also have `.github/workflows/upgrade-testing.yaml` which involves terratest execution so it is a candidate for the enhancement.

Please also sign off the commits to satisfy DCO

Makefile
```diff
@@ -426,8 +426,11 @@ terratest: # Run terratest suite
 		echo "$(RED)Make sure you run the tests against at least two running clusters$(NC)" ;\
 		exit 1;\
 	fi
-	cd terratest/test/ && go mod download && CLUSTERS_NUMBER=$(RUNNING_CLUSTERS) go test -v -timeout 15m -parallel=12 --tags=$(TEST_TAGS)
-
+	mkdir -p /tmp/terratest && cd terratest/test/ && go mod download && CLUSTERS_NUMBER=$(RUNNING_CLUSTERS) go test -v -timeout 15m -parallel=12 --tags=$(TEST_TAGS) | tee /tmp/terratest/all.log
+	curl --location --silent --fail --show-error -o terratest_log_parser https://github.com/gruntwork-io/terratest/releases/download/v0.38.8/terratest_log_parser_linux_amd64
+	chmod +x terratest_log_parser
+	sudo mv terratest_log_parser /usr/local/bin
```
jkremser, line 432: sudo call in makefile isn't a good idea imho

Makefile
```diff
@@ -426,8 +426,11 @@ terratest: # Run terratest suite
 		echo "$(RED)Make sure you run the tests against at least two running clusters$(NC)" ;\
 		exit 1;\
 	fi
-	cd terratest/test/ && go mod download && CLUSTERS_NUMBER=$(RUNNING_CLUSTERS) go test -v -timeout 15m -parallel=12 --tags=$(TEST_TAGS)
-
+	mkdir -p /tmp/terratest && cd terratest/test/ && go mod download && CLUSTERS_NUMBER=$(RUNNING_CLUSTERS) go test -v -timeout 15m -parallel=12 --tags=$(TEST_TAGS) | tee /tmp/terratest/all.log
+	curl --location --silent --fail --show-error -o terratest_log_parser https://github.com/gruntwork-io/terratest/releases/download/v0.38.8/terratest_log_parser_linux_amd64
```
jkremser, line 430: downloading the file all the time isn't optimal on the other hand, having something like [this](https://github.com/k8gb-io/k8gb/blob/master/olm/generate.sh#L17) in here would make it even more complicated. What about doing only the ` | tee ..` & `mkdir ..` part here and the rest in the gh action?
Btw. a platform specific version is hard-coded (I am developing on mac for instance) -> another conditional or Makefile magic... having this hidden in the action yaml would be better imho

Makefile
jkremser, line 433: It's not very obvious, but 
.github/workflows/terratest-more-clusters.yaml
currently runs only the [full_round_robin](https://github.com/k8gb-io/k8gb/blob/master/terratest/test/k8gb_full_roundrobin_test.go) tests. So there is no log interleaving coming from this output.

it would make more sense to have it for
.github/workflows/terratest.yaml

.. which combines multiple tests together and the output is mixed

Btw. the logic that controls when certain tests are run is done using build tags

jkremser (changes requested):
Thanks for your contribution! pls see the comments, otherwise it looks good

Makefile
AugustasV: Understand, but how to have terratest_log_parser executable otherwise?

Makefile
AugustasV: As I removed `terratest_log_parser` from Make file, right now logs will be overwriten every time into `all.log` file

jkremser:
cool 💯 , I can see the uploaded zip file with the separated log files: https://github.com/k8gb-io/k8gb/actions/runs/1664623772
Bummer, that it's not reachable directly from link under pull-request, but I had to go via actions tab.

Could you pls add one more thing? The link that would lead to that web page where it can be get.

this should be the filed - `${{ github.run_id }}`
so pls put something like

`echo Separated log files will be available under artifact section at https://github.com/k8gb-io/k8gb/actions/runs/${{ github.run_id }}`

to Terratest log parser step

example for PR msg:
https://github.com/k8gb-io/k8gb/blob/master/.github/workflows/olm_pr.yaml#L94

and it then looks like this:
https://github.com/k8s-operatorhub/community-operators/pull/499

Another future improvement could be actually printing these logs right away during the build, but let's not complicate it even more. There is a way to make the collapsible log sections like in TravisCi - https://github.community/t/has-github-action-somthing-like-travis-fold/16841 

so then having something like
```bash
Terratest logs:
▶ TestK8gbSplitFailoverExample
▶ TestK8gbIngressAnnotationRR
..
```
would be awesome

AugustasV:
I tried to implement collapsible logs but without success. It collapsed everything. It is probably not possible to collapse one by one. Any ideas?
image

.github/workflows/upgrade-testing.yaml
```diff
@@ -60,5 +60,23 @@ jobs:
       - name: K8GB deploy test version
         run: make deploy-test-version list-running-pods
 
-      - name: Terratest
-        run: make terratest
+      - name: Run Terratest
+        run: |
+          mkdir -p ${{ github.workspace }}/tmp/terratest 
+          echo "::group::Terratest logs"
```
jkremser, line 66:
```suggestion
```

jkremser:
right, nice so it works almost :D, I think we need to iterate through the files one by one and "cat"em within each collapsible section, check the suggestion comments with the diff I've made

as for the DCO check, could you pls sign them all? It's still possible with interactive rebase([tutorial](https://duan.ca/2016/12/24/rebase-multiple-git-commits-with-signing/)). Or you can make your life easier and squash all the commits to just one, up to you, or we can squash it during the PR merge using "github button".

Otherwise some tips for signing are also here: https://github.com/k8gb-io/k8gb/blob/master/CONTRIBUTING.md#signature
All projects under CNCF umbrella have to have this policy.

jkremser (changes requested):
thanks, DCO is passing, we are almost there

AugustasV:
I see that my all DCO signed with ```git log``` command, but here I see it's not

jkremser (approved):

ytsarev (approved):

PR #827, opened 2022-01-11
Make make colorful

jkremser:
what:
`s/@echo/@echo -e/g` in Makefile

why:

the `SHELL := bash` on line 36 is causing this:

```
make deploy-full-local-setup
\n\033[0;33mCreating 2 k8s clusters\033[0m
/Library/Developer/CommandLineTools/usr/bin/make create-local-cluster CLUSTER_NAME=edge-dns
\n\033[0;33mCreate local cluster \033[1;36medge-dns \033[0m
k3d cluster create -c k3d/edge-dns.yaml
INFO[0000] Using config file k3d/edge-dns.yaml (k3d.io/v1alpha3#simple)
INFO[0000] Prep: Network
..
```

i.e. echo doesn't interpret the escape characters for colors, nor the new lines. Colors can be fixed [like this](https://github.com/upbound/build/blob/bd63a4167ae20a71788537217b022fced8f2f854/makelib/common.mk#L30), but new lines (`\n`) require the `echo -e ..` (which also fixes the colors)

after:
Screenshot 2022-01-11 at 16 41 26

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,7 +118,7 @@ demo: ## Execute end-to-end demo
 # spin-up local environment
 .PHONY: deploy-full-local-setup
 deploy-full-local-setup: ensure-cluster-size ## Deploy full local multicluster setup (k3d >= 5.1.0)
-	@echo "\n$(YELLOW)Creating $(CLUSTERS_NUMBER) k8s clusters$(NC)"
+	@echo -e "\n$(YELLOW)Creating $(CLUSTERS_NUMBER) k8s clusters$(NC)"
 	$(MAKE) create-local-cluster CLUSTER_NAME=edge-dns
 	@for c in $(CLUSTER_IDS); do \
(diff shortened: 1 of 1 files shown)


somaritane (approved):
LGTM 👍

kuritka:
@jkremser  would there be any advantages/disadvantages to using `SHELL := /bin/sh` instead of the `-e` argument ?

jkremser:
@kuritka  `SHELL := /bin/sh` would ran into the issue that on Ubuntu this points to the Dash implementation.. 

```
❯ docker run -ti --rm --entrypoint="" ubuntu ls -l /bin/sh
lrwxrwxrwx 1 root root 4 Jul 18  2019 /bin/sh -> dash
```

kuritka (approved):
@jkremser , Thank you for the explanation, lgtm

PR #830, opened 2022-01-14
Structured logging improvements

somaritane:
- Made use of native zerolog formatters where possible
- Unified logging fields (camelCase, no spaces)
- Updated logging messages


kuritka (approved):

controllers/dnsupdate.go
```diff
@@ -103,7 +103,7 @@ func (r *GslbReconciler) gslbDNSEndpoint(gslb *k8gbv1beta1.Gslb) (*externaldns.D
 					log.Info().
 						Str("gslb", gslb.Name).
 						Str("cluster", gslb.Spec.Strategy.PrimaryGeoTag).
-						Str("targets", fmt.Sprintf("%v", finalTargets)).
+						Interface("targets", finalTargets).
```
jkremser, line 106: `Interface  - > Strs`? So that we are consistent in this file. I assume the Strs is faster since we know the type and it's not doing the reflection on the argument.

jkremser (changes requested):

controllers/gslb_controller.go
```diff
@@ -265,10 +264,12 @@ func (r *GslbReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		strToInt := func(s string) int {
+			intValue, err := strconv.Atoi(s)
 			if err != nil {
-				log.Err(err).Msg("can't convert string to int")
+				log.Err(err).
+					Str("value", s).
+					Msg("Can't convert string to int")
 			}
 			return intValue
 		}
```
kuritka, line 275: I'm checking the code and see following:
```go
		strToInt := func(str string) int {
			intValue, err := strconv.Atoi(str)
			if err != nil {
				log.Err(err).Msg("can't convert string to int")
			}
			return intValue
		}

		for annotationKey, annotationValue := range a.GetAnnotations() {
			switch annotationKey {
			case dnsTTLSecondsAnnotation:
				gslb.Spec.Strategy.DNSTtlSeconds = strToInt(annotationValue)
			case splitBrainThresholdSecondsAnnotation:
				gslb.Spec.Strategy.SplitBrainThresholdSeconds = strToInt(annotationValue)
			}
		}

```

`strToInt` is private embeded function parsing value from annotation. Thanks to the fact that the function is not used anywhere else and has one purpose (annotation parsing), we could better target the error message. Instead of the generic "Can't convert string" we could say something like, "I can't parse the annotation... " or something like that.

controllers/dnsupdate.go
somaritane: @jkremser well spotted, thanks! will amend

somaritane: Done

controllers/gslb_controller.go
somaritane: Done

kuritka (approved):

jkremser (approved):

PR #832, opened 2022-01-17
Fix deployCrds template rewrite in generated CRD

somaritane:
- Fixed the auto-removal of deployCrds logic from CRD manifest by
  controller-gen, the manifest is wrapped with template after generation
- Added extra helper messages to affected make targets and functions

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,7 @@ DNS_ZONE ?= cloud.example.com
 DEMO_URL ?= http://failover.cloud.example.com
 DEMO_DEBUG ?=0
 DEMO_DELAY ?=5
+GSLB_CRD_YAML ?= chart/k8gb/templates/crds/k8gb.absa.oss_gslbs.yaml
 
 ifndef NO_COLOR
 YELLOW=\033[0;33m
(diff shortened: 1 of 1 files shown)


jkremser (approved):
thanks, I missed that in #818

Makefile
```diff
@@ -506,16 +511,21 @@ define wait-for-ingress
...
 endef
 
-define manifest
-	$(call controller-gen,crd:crdVersions=v1 paths="./..." output:crd:artifacts:config=chart/k8gb/templates/crds)
+define crd-manifest
+	$(call install-controller-gen)
+	@echo -e "\n$(YELLOW)Generating the CRD manifests$(NC)"
+	@echo -n "{{- if .Values.k8gb.deployCrds }}" > $(GSLB_CRD_YAML)
+	$(GOBIN)/controller-gen crd:crdVersions=v1 paths="./..." output:crd:stdout >> $(GSLB_CRD_YAML)
+	@echo "{{- end }}"  >> $(GSLB_CRD_YAML)
```
ytsarev, line 524: looks like it will tamper the git status?

somaritane: @ytsarev not sure I'm following

ytsarev: @somaritane i've meant from the PoV that it will affect `git status` after regeneration but it is actually valid in case of API changes. And if there are no API changes in the working tree the GSLB_CRD_YAML file will stay intact. So all good, approving

ytsarev (approved):

PR #839, opened 2022-01-21
PATCH: downgrade k3d version

kuritka:
In our build pipe, k3d started crashing quite often when image-import was being executed.
Process locked up during image import and produced error `write unix @->/run/docker.sock: use of closed network connection` on manual stop.
I checked that the image actually exists in the virtual environment and it exists.
Since we encountered this bug with the latest version of k3d, I downgraded back to
the latest stable k3d (but still latest k3d-action).

**TEST:**
The bug was so frequent that I did a rerun of our workflow five times to get the bug to show up. The bug didn't occured.

### DIFF
--- a/.github/workflows/helm_publish.yaml
+++ b/.github/workflows/helm_publish.yaml
@@ -21,6 +21,7 @@ jobs:
       - name: Create k3s cluster
         uses: AbsaOSS/k3d-action@v2
         with:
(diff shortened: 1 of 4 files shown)


ytsarev (approved):

k0da (approved):

somaritane (approved):
@kuritka LGTM! 👍

Issue #840, opened 2022-01-24
Optimize GitHub Actions execution


ytsarev:
We have quite an amount of duplication in action triggering given the parallel reaction to both `push` and `pull_request`

We can try to optimize it with https://github.com/fkirc/skip-duplicate-actions

PR #844, opened 2022-01-26
workflow optimization

kuritka:
related to #840

skipping already finished workflows, NOT concurrent (different PR). See [skip detail](https://github.com/k8gb-io/k8gb/actions/runs/1751931140).

I've been doing investigations about canceling workflows (https://github.com/fkirc/skip-duplicate-actions). The action works by checking in pre_job whether the pipe has already been run. Then in the main job there is a condition that either continues processing or skips the job.

This results in ALWAYS running all workflows, even duplicate ones. The difference is that the duplicates will terminate at a different point. We will still see a lot of green runs.

_It's a bit different than I imagined (not starting or CANCEL the duplicate at all).
I found out that I can cancel the workflow programmatically via HTTP POST. 
So if we create custom action, we can cancel on top of workflow and probably delete 
cancelled job from workflows page. Such behaviour would be more expected._

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,9 +15,23 @@ on:
       - '**.drawio'
       - '.spelling'
 jobs:
+  skip-check:
+    runs-on: ubuntu-latest
+    name: Skip the job?
+    outputs:
+          should_skip: ${{ steps.skip_check.outputs.should_skip }}
+    steps:
+    - id: skip_check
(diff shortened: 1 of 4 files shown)


k0da (approved):
LGTM

ytsarev (approved):
that's pretty cool, thanks!

jkremser (approved):

PR #845, opened 2022-02-03
Expose CoreDNS over tcp + make Dig in terratests go via tcp

jkremser:
why: When we get rid of Docker Desktop and use for instance Lima, it doesn't have support for UDP port forwarding from host to containers running in the VM - https://github.com/lima-vm/lima/issues/366

So as a workaround, we can run the Dig in our tests using the tcp protocol, where the redirect works fine (it uses ssh tunnel). 

~tests will fail, it assumes a new release on `AbsaOSS/gopkg` with this https://github.com/AbsaOSS/gopkg/pull/11 merged in~

It adds a new service for coredns that uses NodePort (30053) and then adds this redirect in k3d manifests so that

```
dig @127.0.0.1 -p5053 failover.cloud.example.com +short +tcp
dig @127.0.0.1 -p5054 failover.cloud.example.com +short +tcp
```
both work

~todo: update docs for local playground~

that `AbsaOSS/gopkg` library contains a module called strings, that was renamed to string like a year ago, but only now it's [failing](https://github.com/k8gb-io/k8gb/runs/5069047324?check_suite_focus=true#step:6:7) because of the newly released version. Hence the `strings -> string` thing.

btw:
> DNS resolvers and recursive servers MUST support UDP, and SHOULD support TCP, for sending (non-zone-transfer) queries.
 --- https://datatracker.ietf.org/doc/html/rfc7766

### DIFF
--- a/chart/k8gb/templates/coredns-tcp-svc.yaml
+++ b/chart/k8gb/templates/coredns-tcp-svc.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.coredns.deployment.enabled }}
+{{- if and .Values.coredns.deployment.enabled (not .Values.route53.enabled) }}
 apiVersion: v1
 kind: Service
 metadata:
--- a/chart/k8gb/templates/coredns-tcp-svc.yaml
+++ b/chart/k8gb/templates/coredns-tcp-svc.yaml
@@ -1,0 +1,19 @@
+{{- if .Values.coredns.deployment.enabled }}
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+{{ include "chart.labels" . | indent 4  }}
+  name: {{ .Release.Name }}-coredns-tcp
+spec:
+  type: NodePort
+  selector:
+    app.kubernetes.io/instance: k8gb
+    app.kubernetes.io/name: coredns
+  ports:
+  - name: tcp-5353
+    port: 53
+    protocol: TCP
+    targetPort: 5353
+    nodePort: 30053
+{{- end }}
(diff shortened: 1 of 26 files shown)


ytsarev (commented):
Aren't we making here a strong assumption that k8gb contributor should use mac+lima?

chart/k8gb/templates/coredns-tcp-svc.yaml
```diff
@@ -0,0 +1,19 @@
...
+  name: {{ .Release.Name }}-coredns-tcp
+spec:
+  type: NodePort
+  selector:
+    app.kubernetes.io/instance: k8gb
+    app.kubernetes.io/name: coredns
+  ports:
+  - name: tcp-5353
+    port: 53
+    protocol: TCP
```
ytsarev, line 16: Was it tested outside of the local scenario? Main concern is overall TCP support/expectation in the DNS environment

somaritane (approved):
lgtm

kuritka (approved):

chart/k8gb/templates/coredns-tcp-svc.yaml
k0da: @jkremser could you please move this service definition out of helm and deploy it during test setup creation (around we deploy edgeDNS)

k0da (changes requested):

chart/k8gb/templates/coredns-tcp-svc.yaml
jkremser: it should be there

k0da:
This way it will be deployed into default namespace

jkremser:
> This way it will be deployed into default namespace

good point, added `k8gb` ns

k0da (dismissed):
LGTM

k0da (approved):

PR #850, opened 2022-03-17
RELEASE: v0.8.8

jkremser:
This is probably the last release that supports the ingress at version `v1beta1`. From now on, only the `Ingress@v1` will be supported.

note: `Ingress@v1beta1` has been deprecated and it's not present in k8s since version 1.22


jkremser:
We are not buildable at the moment because of [this](https://github.com/kubernetes-sigs/external-dns/issues/2634) (PR for this https://github.com/kubernetes-sigs/external-dns/pull/2635). Build currently fails on this dependency `hcloud-dns` that's brought transitively by `external-dns`. We could have possibly go-vendor it, but I'd wait until it's fixed upstream.. it must impact everyone who depends on `external-dns` so I expect it will not take long time.

Raffo:
hcloud-dns has been removed from ExternalDNS, so you are hopefully good to go.

somaritane (approved):
@jkremser lgtm!

Issue #847, opened 2022-02-17
Support for Ingress @ networking.k8s.io/v1


jkremser:
Since k8s@1.22 the Ingress with version `extensions/v1beta1` nor `networking.k8s.io/v1beta1` are no longer supported. The only supported version is now `networking.k8s.io/v1`

If k8gb is deployed to a kubernetes cluster with version 1.22 or higher, it keeps crash-looping on this error:

```
2022-02-16T16:03:02Z ERR github.com/k8gb-io/k8gb/controllers/logging/logr.go:60 > gslb: if kind is a CRD, it should be installed before calling Start {"reconciler group":"k8gb.absa.oss","reconciler kind":"Gslb"} error="no matches for kind \"Ingress\" in version \"networking.k8s.io/v1beta1\""
```

somaritane:
@jkremser, this is how it was done in [helm.sh](https://helm.sh/) for sample chart creation (with `helm create` command):
https://github.com/helm/helm/pull/9621/files#diff-92d0c93dff3a254f887d928f60f6b46ad1da1cbe7a5737dda4cd85cd72876092R10

jkremser:
right, in theory we can handle this issue also on the helm chart level. Having those `(semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion))` and deploy older version of k8gb for older k8s clusters, but this smells like some auto-magic behavior and I am not a big fan of this approach, the less logic in helm, the better, imho

I was also looking into that possibility of having multiple versions of CRDs present at the same time (because this change also impacts the shape under `.spec.ingress` in our gslbs) which can be handy also in the future. There is a nice tutorial how to do that on [kubebuilder web](https://book.kubebuilder.io/multiversion-tutorial/tutorial.html). But it requires a new webhook, cert-manager and what not. I've actually started this approach and then abandoned it after writing those conversion methods.

PR #854, opened 2022-03-21
Ingress v1

jkremser:
Fixes https://github.com/k8gb-io/k8gb/issues/847 (we can't deploy on k8s@1.22 and newer)

All the details are described in the issue ^

This is an API breaking change, because the ingress@v1 resource has a slightly different structure and we use that sub-tree in our own gslb CRD.

Nginx ingress-controller was updated because version 3 works w/ the old ingress (v1beta1)

The idea is that for Ingress@v1beta1 (kubernetes version 1.21 and lower) use k8gb in version `0.8.8` or lower and for Ingress@v1 (since k8s@1.19), use the `> 0.8.8`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,13 +36,13 @@ spec:
       - host: failover.test.k8gb.io # Desired GSLB enabled FQDN
         http:
           paths:
-          - backend:
-            path: /
+          - path: /
             pathType: Prefix
-            service:
-              name: frontend-podinfo # Service name to enable GSLB for
-              port:
-                name: http
+            backend:
+              service:
+                name: frontend-podinfo # Service name to enable GSLB for
+                port:
+                  name: http
   strategy:
     type: failover # Global load balancing strategy
     primaryGeoTag: eu-west-1 # Primary cluster geo tag
--- a/controllers/status.go
+++ b/controllers/status.go
@@ -59,6 +59,14 @@ func (r *GslbReconciler) getServiceHealthStatus(gslb *k8gbv1beta1.Gslb) (map[str
 	serviceHealth := make(map[string]k8gbv1beta1.HealthStatus)
 	for _, rule := range gslb.Spec.Ingress.Rules {
 		for _, path := range rule.HTTP.Paths {
+			if path.Backend.Service == nil || path.Backend.Service.Name == "" {
+				log.Warn().
+					Str("gslb", gslb.Name).
+					Interface("service", path.Backend.Service).
+					Msg("Malformed service definition")
+				serviceHealth[rule.Host] = k8gbv1beta1.NotFound
+				continue
+			}
 			service := &corev1.Service{}
 			finder := client.ObjectKey{
 				Namespace: gslb.Namespace,
--- a/Makefile
+++ b/Makefile
@@ -175,7 +175,7 @@ deploy-local-cluster:
 	helm repo add --force-update nginx-stable https://kubernetes.github.io/ingress-nginx
 	helm repo update
 	helm -n k8gb upgrade -i nginx-ingress nginx-stable/ingress-nginx \
-		--version 3.24.0 -f deploy/ingress/nginx-ingress-values.yaml
+		--version 4.0.15 -f deploy/ingress/nginx-ingress-values.yaml
 
 	@if [ "$(DEPLOY_APPS)" = true ]; then $(MAKE) deploy-test-apps ; fi
 
--- a/k3d/edge-dns.yaml
+++ b/k3d/edge-dns.yaml
@@ -1,7 +1,7 @@
 apiVersion: k3d.io/v1alpha3
 kind: Simple
 name: edgedns
-image: docker.io/rancher/k3s:v1.21.2-k3s1
+image: docker.io/rancher/k3s:v1.22.6-k3s1
 agents: 0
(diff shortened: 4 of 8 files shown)


kuritka (commented):
@jkremser, nice change. Does it make sense to consolidate import identifier `netv1` and `v1` into one ?  
 - `v1 "k8s.io/api/networking/v1"`
 -  `netv1 "k8s.io/api/networking/v1"`

jkremser:
@k0da another part of this should be also bumping the version of `external-dns` deployment, right? This might go in a separate PR afterwards.

Although we don't use the Ingresses nor Services for communicating the DNS records for ext-dns (we use CRs). I can imagine there might be some expectations baked in the external-dns binary. however the tests are green with `external-dns:rfc-ns1` img. 

Would it be hard to rebase our changes on the latest external-dns? What was the repo for this image `absaoss/external-dns:rfc-ns1`, is it this [branch](https://github.com/k0da/external-dns/tree/rfc2136_ns)? I can't find `external-dns` repo under `absaoss` GH org, nor under `k8gb-io` org.

side note: we might use similar compatibility matrix as ext-dns have in their readme: https://github.com/kubernetes-sigs/external-dns/pkgs/container/external-dns%2Fexternal-dns#kubernetes-version-compatibility

k0da:
@jkremser We simply can switch to latest released upstream. As change was finally merged a while ago

README.md
```diff
@@ -37,9 +37,12 @@ spec:
         http:
           paths:
           - backend:
```
kuritka, line 39: does `- backend:` suppose to be empty ? 
I wouldn't care but in README.md,  for easier orientation we can leave `path` as the first item https://github.com/k8gb-io/k8gb/pull/854/files#diff-524f9815d92e58f96cf2ca5d5cb3b2457e7dbf58b7b16d05b10bb2551eb1d041R30-R35

chart/k8gb/Chart.yaml
```diff
@@ -5,6 +5,7 @@ icon: https://www.k8gb.io/assets/images/icon-192x192.png
 type: application
 version: v0.8.8
 appVersion: v0.8.8
+kubeVersion: ">= 1.19.0-0"
```
kuritka, line 8: Maybe I didn't understand properly. In PR description you mentioned _Since k8s@1.22 the Ingress with version extensions/v1beta1 nor networking.k8s.io/v1beta1 are no longer supported_ . Is `>= 1.19.0-0` expected in Helm chart? shouldn't be `>= 1.22.0-0` ?

kuritka (changes requested):
Overall Looks good. check my comments / questions please.

chart/k8gb/Chart.yaml
jkremser: `Ingress@v1` was introduced in `1.19` and `v1Beta1` was deprecated, since `1.22` `v1beta1` is gone. There was an overlap. Check this [compatibility matrix](https://github.com/kubernetes-sigs/external-dns#kubernetes-version-compatibility), this pretty much holds also for us.

In our case, it's gonna be:

| k8gb                    |      <= 0.8.x      |     >= 0.9.0 (or 1.0.0)      |
| ------------------------------ | :----------------: | :----------------: |
| Kubernetes <= 1.18             | :white_check_mark: |        :x:         |
| Kubernetes >= 1.19 and <= 1.21 | :white_check_mark: | :white_check_mark: |
| Kubernetes >= 1.22             |        :x:         | :white_check_mark: |

README.md
jkremser: good catch, should be fixed now

controllers/status.go
```diff
@@ -62,7 +62,7 @@ func (r *GslbReconciler) getServiceHealthStatus(gslb *k8gbv1beta1.Gslb) (map[str
 			service := &corev1.Service{}
 			finder := client.ObjectKey{
 				Namespace: gslb.Namespace,
-				Name:      path.Backend.ServiceName,
+				Name:      path.Backend.Service.Name,
```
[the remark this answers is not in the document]
jkremser: right, backend itself can't be nil because it doesn't have the "omitempty" tag, but `path.Backend.Service` can be nil (it's `path.Service` XOR `path.Resource` we support only the Service ref)

If service is there, then the Name has to be specified (again not "omniempty" tag). But to stay on the safe side, I can be paranoid and check everything :)

chart/k8gb/Chart.yaml
kuritka: thx for clarification @jkremser !

controllers/status.go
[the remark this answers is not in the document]
jkremser: I've added some checks and logged the warning msg if it's malformed. Also marking the svc health as `k8gbv1beta1.NotFound`... in a sense: "the info user gave us was not sufficient to find the service"

Hopefully, it won't be that spammy in the logs. It's going to be there in each reconciliation, but on the other hand, if we "swallowed" the error, it's even worse.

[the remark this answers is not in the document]
kuritka: ok, lets keep it as it is and we will see afterwards. Possible candidate for validation would be depresolver_spec.

kuritka (approved):
lgtm

PR #869, opened 2022-04-04
Add opencontainers labels for container image

jkremser:
Although, we don't produce OCI compliant images at the moment. It would be nice to adapt their [metadata standards](https://github.com/opencontainers/image-spec/blob/main/annotations.md) for start. Some tools can already [interpret](https://github.com/jkremser/log2rbac-operator/pkgs/container/log2rbac/18136364?tag=v0.0.3) them even if provided as labels in the image json manifest and not as intended (which is OCI [annotations](https://github.com/imjasonh/ImJasonH/tree/main/articles/oci-base-image-annotations#when-can-i-use-it)).

I was looking into how to produce those OCI images and it's not that easy with goreleaser. It can support podman as enterprise feature, otherwise it uses docker. I was playing with `skopeo` that can save any image as tar in OCI format. Github registries are [ready](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#about-container-registry-support) for OCI images.

There are also `org.opencontainers.image.base.name` & `org.opencontainers.image.base.digest` annotations/labels, that can help to construct the full lineage of predecessor of the image (`FROM ...` relation). These can be added later if we decide that we or community need/want them.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -21,11 +21,23 @@ dockers:
   - "absaoss/k8gb:v{{ .Version }}-amd64"
   build_flag_templates:
   - "--platform=linux/amd64"
+  - &LABEL1
+    "--label=org.opencontainers.image.title=k8gb"
+  - &LABEL2
+    "--label=org.opencontainers.image.description=A cloud native Kubernetes Global Balancer"
+  - &LABEL3
+    "--label=org.opencontainers.image.revision={{ .FullCommit }}"
+  - &LABEL4
+    "--label=org.opencontainers.image.version={{ .Version }}"
 - image_templates:
   - "absaoss/k8gb:v{{ .Version }}-arm64"
   goarch: arm64
   build_flag_templates:
   - "--platform=linux/arm64"
+  - *LABEL1
+  - *LABEL2
+  - *LABEL3
+  - *LABEL4
 docker_manifests:
 - name_template: absaoss/k8gb:{{ .Tag }}
   image_templates:


ytsarev (commented):
Isn't it something that can potentially enhanced in operator-sdk build scripts upstream?

jkremser:
> Isn't it something that can be potentially enhanced in operator-sdk build scripts upstream?

right, can be, can be.. Although they don't use the goreleaser, but instead prepare some makefile target with `docker` command. 

Recently there was a  [long discussion](https://kubernetes.slack.com/archives/CAR30FCJZ/p1649068358040189) about the makefile on kubebuilder@slack. They want to make it more general so that one can plug any other tool instead of the docker... but this assumes the same api contract.. you know all the params need to be the same. It hasn't been merged yet, but if it is, I would rather not add it, because these labels need to be passed as `docker build --label ..` or as build-args as I do it [here](https://github.com/jkremser/log2rbac-operator/blob/master/Makefile#L84) and then interpret them in the Dockerfile.

.github/workflows/build.yml
```diff
@@ -33,12 +33,18 @@ jobs:
...
+        with:
+          go-version: 1.17.6
       - uses: actions/checkout@v2
       # see: https://golangci-lint.run/usage/configuration/#config-file
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v2.5.2
         with:
           version: v1.42.1
+          skip-go-installation: true
+          skip-pkg-cache: true
```
ytsarev, line 47: Could you briefly explain this option switch?

jkremser: https://github.com/golangci/golangci-lint-action/blob/master/README.md?plain=1#L62
There is an issue open for this https://github.com/k8gb-io/k8gb/issues/870 I will split the PR into two separate ones to make this cleaner.

jkremser: I've extracted those bits that deals with `golangci-lint` into a standalone [pr](https://github.com/k8gb-io/k8gb/pull/873) and the `skip-pkg-cache` switch actually was not necessary, so good catch. Now the build on this PR is obviously failing and needs to be re-run after the pr 873 is merged, actually whole master is now in troubles, but there were no other PRs going in so it looks ok :D

ytsarev (approved):

PR #874, opened 2022-04-22
Add json schema for helm chart values

jkremser:
Currently there is no validation in place when helm install/upgrade is running so that if some unexpected value is passed, it fails during the operator startup ending up in k8s crash loop. Let's catch some potential issues sooner by introducing json schema for the helm chart's values.

example:

```bash
λ helm -n k8gb upgrade -i k8gb ./chart/k8gb -f "" \
                --set k8gb.clusterGeoTag='eu' --set k8gb.extGslbClustersGeoTags='' \
                --set k8gb.reconcileRequeueSeconds=ahoj \
                --set k8gb.dnsZoneNegTTL=-10 \
                --set k8gb.imageTag="" \
                --set k8gb.log.format=sYmple \
                --set k8gb.log.level=deBUG \
                --set rfc2136.enabled=maybe \
                --set externaldns.image=""
walk.go:74: found symbolic link in path: /Users/ab017z6/workspace/k8gb/chart/k8gb/LICENSE resolves to /Users/ab017z6/workspace/k8gb/LICENSE
Error: UPGRADE FAILED: values don't meet the specifications of the schema(s) in the following chart(s):
k8gb:
- rfc2136.enabled: Invalid type. Expected: boolean, given: string
- k8gb.extGslbClustersGeoTags: String length must be greater than or equal to 1
- k8gb.reconcileRequeueSeconds: Invalid type. Expected: integer, given: string
- k8gb.log.level: k8gb.log.level must be one of the following: "panic", "fatal", "error", "warn", "info", "debug", "trace"
- k8gb.log.format: k8gb.log.format must be one of the following: "simple", "json"
- k8gb.imageTag: String length must be greater than or equal to 1
- k8gb.dnsZoneNegTTL: Must be greater than or equal to 0
- externaldns.image: String length must be greater than or equal to 1
```

note: I've used this https://jsonformatter.org/yaml-to-jsonschema to bootstrap the schema and then edited it manually. This can't be automated w/o the domain knowledge (allowed values, reg exps, enums, etc.)

### DIFF
--- a/chart/k8gb/values.schema.json
+++ b/chart/k8gb/values.schema.json
@@ -1,0 +1,394 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$ref": "#/definitions/All",
+    "definitions": {
+        "All": {
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+                "global": {
+                    "$ref": "#/definitions/Global"
+                },
+                "k8gb": {
+                    "$ref": "#/definitions/k8gb"
+                },
+                "externaldns": {
+                    "$ref": "#/definitions/Externaldns"
+                },
+                "coredns": {
+                    "$ref": "#/definitions/Coredns"
+                },
+                "infoblox": {
+                    "$ref": "#/definitions/Infoblox"
+                },
+                "route53": {
+                    "$ref": "#/definitions/Route53"
+                },
+                "ns1": {
+                    "$ref": "#/definitions/Ns1"
+                },
+                "openshift": {
+                    "$ref": "#/definitions/Openshift"
+                },
+                "rfc2136": {
+                    "$ref": "#/definitions/Rfc2136"
+                }
+            },
+            "required": []
+        },
+        "Coredns": {
+            "type": "object",
+            "additionalProperties": true,
+            "properties": {
+                "isClusterService": {
+                    "type": "boolean"
+                },
+                "deployment": {
+                    "$ref": "#/definitions/CorednsDeployment"
+                },
+                "image": {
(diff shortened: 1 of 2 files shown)


jkremser:
when working on this, I've figured out that `externaldns.interval` is not being used anywhere. Should we remove it / use it for the deployment of `external-dns` (`--interval=42s` [upstream chart](https://github.com/kubernetes-sigs/external-dns/blob/f3e04dfe8c46853966b878ec76f3eeb6b2ed1441/charts/external-dns/templates/deployment.yaml#L58))

kuritka:
@jkremser , hello , maybe I'm asking the wrong question , but json is only option? (no yaml?)

jkremser:
@kuritka 
yep, only json schema is supported https://helm.sh/docs/topics/charts/#schema-files

ytsarev (approved):
that looks great 👍

kuritka (approved):

somaritane:
> when working on this, I've figured out that `externaldns.interval` is not being used anywhere. Should we remove it / use it for the deployment of `external-dns` (`--interval=42s` [upstream chart](https://github.com/kubernetes-sigs/external-dns/blob/f3e04dfe8c46853966b878ec76f3eeb6b2ed1441/charts/external-dns/templates/deployment.yaml#L58))

@jkremser Looks like the param was present in external-dns template, but we lost it during introduction of coredns plugin. https://github.com/k8gb-io/k8gb/commit/cb79e1ee1df7ea31f1ce6c17e2d3c84e415c67c8#diff-c5b1eb253f3b95c48e2f9203764d56752b157636d69930e79e5e7e27a59567d0L76
Probably need to restore back

PR #889, opened 2022-05-06
Add copy to clipboard button for code snippets

jkremser:
This pr also
- fixes the Makefile target for colima environment
- don't use black text on black background for syntax highlighting (see the screenshots)

before:
Screenshot 2022-05-06 at 14 21 17

after:
Screenshot 2022-05-06 at 14 22 01

If one doesn't want to have the copy button present on the code snippet, just use either:

``````
```
some text output from a command that doesn't need to be copied
```
``````
or
``````
```text
something
```
``````

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ ifneq ($(CUSTOM_CERT_PATH),)
 	CERT_INSTALL_CMD := update-ca-certificates;
 endif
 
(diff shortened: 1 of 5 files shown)


somaritane (changes requested):
@jkremser, looks good, just one thing: can we modify the style to display the [copy] element only on hover?

jkremser:
@somaritane it should be there. It shows up only after hovering the parent element

somaritane:
> @somaritane it should be there. It shows up only after hovering the parent element

Hmm, interesting. This is what it looks like when I serve the website locally from the PR branch:
image
But on Netlify preview it looks OK, so no biggie 👍

Issue #892, opened 2022-05-25
Another example app


jkremser:
Podinfo is cool, but we can have more examples than podinfo and curl. Contour guys were using this on their presentations https://github.com/kubernetes-up-and-running/kuard

jkremser:
done in https://github.com/k8gb-io/k8gb/pull/894

PR #894, opened 2022-06-01
Add Kuar web app as another demo application

jkremser:
Related issue: #892 

I didn't want to make our makefile even more confusing so it's in a standalone script.

some screens:
Screenshot 2022-06-01 at 18 04 29
Screenshot 2022-06-01 at 18 04 16

Nice feature of the app is that it can simulate a failure from the web ui. So one can set the artificial failure for next 10 readiness probes in the web ui. So that the failover should be done and then switch back. It also has web based "dig" (hence adding the coredns servers to `/etc/resolv.conf`) + file browser where one can see the content of `/etc/resolv.conf`. So it's "clicky clicky", but might be handy for demos.


kuritka:
looks good, shouldn't it have some reference to Makefile (deploy.sh execution), or installation directly from Makefile ?

jkremser:
>  looks good, shouldn't it have some reference to Makefile (deploy.sh execution), or installation directly from Makefile ?

That's what I wanted to avoid. I think it's pretty complicated as it is. Maybe after the refactoring.

kuritka (approved):

somaritane:
@jkremser, looks good.
The related issue should be #892.
Is the plan to deploy kuard along with other test apps? In that case, `deploy-test-apps` and `clean-test-apps` makefile targets should be updated. 
If not, then I still believe there should be a target for that, something like `deploy-kuard-app` or similar.
Would be good to update the contributor docs as well.
Otherwise, there's no info for contributors and users on how to deploy and use the app.

docs/local-kuar.md
```diff
@@ -0,0 +1,43 @@
+# Local playground for testing and development
```
somaritane, line 1: Small suggestion:
```suggestion
# Local playground with Kuar web app
```

docs/local-kuar.md
```diff
@@ -0,0 +1,43 @@
...
+
+```sh
+make deploy-kuar-app
+```
+
+This task will deploy a simple web application called Kuar into both clusters and exposes it. It also patches the installed nginx controller to serve the app even if the `Host` header is not provided in the request. This way you can access the application in your browser under http://localhost:80 for cluster 1 and http://localhost:81 for cluster 2.
+
+Make sure the app on http://localhost is responding, it may take a minute for the nginx ingress controller to restart with the correct parameters.
+
+## Kuar
```
somaritane, line 23: I'd probably move brief info about the Kuar app to the very beginning.

docs/local-kuar.md
```diff
@@ -0,0 +1,43 @@
+# Local playground for testing and development
+
+Make sure you have all the tools mentioned in [this section](./local.md#environment-prerequisites) installed.
+
+First spin up two local k3s clusters:
```
somaritane, line 5:
```suggestion
First, spin up two local k3s clusters:
```

docs/local-kuar.md
```diff
@@ -0,0 +1,43 @@
...
+make deploy-kuar-app
+```
+
+This task will deploy a simple web application called Kuar into both clusters and exposes it. It also patches the installed nginx controller to serve the app even if the `Host` header is not provided in the request. This way you can access the application in your browser under http://localhost:80 for cluster 1 and http://localhost:81 for cluster 2.
+
+Make sure the app on http://localhost is responding, it may take a minute for the nginx ingress controller to restart with the correct parameters.
+
+## Kuar
+
+This application contains couple of handy features that can verify the k8gb functionality. It can resolve DNS names, simulate probe failures, can simulate a cpu intensive workload by generating RSA key pairs and also contains a simple api for queue (push & pop strings).
```
somaritane, line 25:
```suggestion
This application contains a couple of handy features that can verify the k8gb functionality. It can resolve DNS names, simulate probe failures, can simulate a CPU-intensive workload by generating RSA key pairs, and also contains a simple API for the queue (push & pop strings).
```

docs/local-kuar.md
```diff
@@ -0,0 +1,43 @@
...
+
+Together with Kuar, we also prepared the failover gslb resource for k8gb. Where the first cluster (geotag = `eu`) is the primary one.
+
+## Simulate failure
+
+Before we do that, we can visit Kuar on [cluster 2](http://localhost:81) and verify that hostname `kuar.cloud.example.com` is correctly resolved IPs corresponding to nodes in the primary cluster - which is the cluster 1. This dns resolution will work the same way on both clusters, but we will be simulating the failure on cluster 1 soon so the web api will not respond on cluster 1.
+
+![Kuar dns resolution](/docs/images/kuar1.png)
+
+Now we can open a new tab with Kuar on cluster one and simulate the readiness probe failure for next 10 checks - http://localhost/-/readiness.
```
somaritane, line 37:
```suggestion
Now we can open a new tab with Kuar on cluster one and simulate the readiness probe failure for the next 10 checks - http://localhost/-/readiness.
```

somaritane (changes requested):
@jkremser, looks awesome, just a few things to make Grammarly happy :)

jkremser:
Thanks @somaritane for your 👀! All your fixes/suggestions should be there.

somaritane (approved):
@jkremser thanks a lot, lgtm

PR #895, opened 2022-06-02
Bump the config files for k3d to v1alpha4 format

jkremser:
`v1alpha3` has been [deprecated](https://github.com/k3d-io/k3d/blob/main/CHANGELOG.md#deprecated-1) in `k3d@5.3.0` and our build produces the [warnings](https://github.com/k8gb-io/k8gb/runs/6607508138?check_suite_focus=true#step:6:22).

~Unfortunately, there was a breaking change introduced between `v1alpha3` -> `v1alpha4`. Their json schema is pretty aggressive and complains about the `name` being part of the cluster config. Furthermore, if we remove it from the yaml config and don't specify it during the k3d command, it will be set to some default value (k3d-k3s-cluster or similar) so it needs to be there for the k3d command and needs to be missing in the config files.~

~Also aligning the naming for "edgedns" cluster (`edgedns` -> `edge-dns`), because it wasn't consistent, meaning the file name with the config didn't match the cluster name.~

UPDATE: The `name` was just moved under `metadata` object.

k3d-action should be [ok](https://github.com/AbsaOSS/k3d-action/blob/main/run.sh#L71) with this change.

### DIFF
--- a/k3d/edge-dns.yaml
+++ b/k3d/edge-dns.yaml
@@ -1,6 +1,7 @@
-apiVersion: k3d.io/v1alpha3
+apiVersion: k3d.io/v1alpha4
 kind: Simple
-name: edgedns
+metadata:
+  name: edgedns
 image: docker.io/rancher/k3s:v1.22.6-k3s1
 agents: 0
 network: k3d-action-bridge-network
--- a/k3d/gslb.yaml.tmpl
+++ b/k3d/gslb.yaml.tmpl
@@ -1,6 +1,7 @@
-apiVersion: k3d.io/v1alpha3
+apiVersion: k3d.io/v1alpha4
 kind: Simple
-name: test-gslb$CLUSTER_INDEX
+metadata:
+  name: test-gslb$CLUSTER_INDEX
 image: docker.io/rancher/k3s:v1.22.6-k3s1
 agents: 1
 network: k3d-action-bridge-network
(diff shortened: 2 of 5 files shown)


k0da:
@jkremser I still see a name in v1alpha4:
`    "metadata": {
      "type": "object",
      "properties": {
        "name": {
          "description": "Name of the cluster (must be a valid hostname and will be prefixed with 'k3d-'). Example: 'mycluster'.",
          "type": "string",
          "format": "hostname"`

k0da:
It was a point to have name specified in a config but not as argument

jkremser:
> @jkremser I still see a name in v1alpha4:

hmm, weird

```bash
λ k3d version
k3d version v5.4.1
k3s version v1.22.7-k3s1 (default)

λ make deploy-full-local-setup

Creating 2 k8s clusters
/Library/Developer/CommandLineTools/usr/bin/make create-local-cluster CLUSTER_NAME=edge-dns

Create local cluster edge-dns
k3d cluster create edge-dns -c k3d/edge-dns.yaml
FATA[0000] Schema Validation failed for config file k3d/edge-dns.yaml: - (root): Additional property name is not allowed
make[1]: *** [create-local-cluster] Error 1
```

> It was a point to have name specified in a config but not as argument

Could you elaborate more? What was the point? I like the new format better actually.. having the "anonymous" specs (/profile) for the cluster.. well, it still needs to be somewaht personalised because of the nodeport shifts, but this won't affect majority of k3d users.

jkremser:
@k0da 
Ah found it, it moved under [metadata](https://k3d.io/v5.4.1/usage/configfile/#all-options-example) object. Ok I will change this as it was and only change the configs 👍

ytsarev (approved):

k0da (approved):

PR #896, opened 2022-06-02
Always run terratest_log_parser and upload the artifact

jkremser:
Running the `terratest_log_parser` and uploading the zip with the logs makes especially sense if there is an error so that one can debug it. However currently, if there is a failure the subsequent tasks are not being run.

Other option would be having `if failure()` there but why not to have the nice logs even for success?

`continue-on-error: true` would not work because it would make the pipeline green even if there is an issue.

Example of such pipeline with an error but the logs were not processed: https://github.com/k8gb-io/k8gb/runs/6706323648?check_suite_focus=true

### DIFF
--- a/.github/workflows/terratest.yaml
+++ b/.github/workflows/terratest.yaml
@@ -85,6 +85,7 @@ jobs:
           echo "::endgroup::"
 
       - name: Terratest log parser
+        if: always()
         run: |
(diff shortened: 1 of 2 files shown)


ytsarev (approved):
Nice one

kuritka (approved):

Issue #50, opened 2020-02-24
Implement weighted round robin load balancing strategy


donovanmuller:
As per the supported load balancing strategies in the [initial design](https://github.com/AbsaOSS/ohmyglb/blob/master/docs/index.md#load-balancing-strategies) a _weighted_ round robin strategy should be implemented to ensure the guarantees stated:

> **Weighted round robin** - Specialisation of the above (default round robin #45) strategy but where a percentage weighting is applied to determine which cluster's Ingress node IPs to resolve. E.g. 80% cluster X and 20% cluster Y

Scenario 1:

* Given 2 separate Kubernetes clusters, X, and Y
* Each cluster has a _healthy_ `Deployment` with a backend `Service` called `app` and that backend service exposed with a `Gslb` resource on cluster X as:

```yaml
apiVersion: ohmyglb.absa.oss/v1beta1
kind: Gslb
metadata:
  name: app-gslb
  namespace: test-gslb
spec:
  ingress:
    rules:
      - host: app.cloud.example.com
        http:
          paths:
            - backend:
                serviceName: app
                servicePort: http
              path: /
  strategy: roundRobin 
    weight: 80%
```

and a `Gslb` resource on cluster Y as:

```yaml
apiVersion: ohmyglb.absa.oss/v1beta1
kind: Gslb
metadata:
  name: app-gslb
  namespace: test-gslb
spec:
  ingress:
    rules:
      - host: app.cloud.example.com
        http:
          paths:
            - backend:
                serviceName: app
                servicePort: http
              path: /
  strategy: roundRobin 
    weight: 20%
```

* Each cluster has _one_ worker node that accepts Ingress traffic. The worker node in each cluster has the following name and IP:

```
cluster-x-worker-1: 10.0.1.10
cluster-y-worker-1: 10.1.1.11
```

When issuing the following command, `curl -v http://app.cloud.example.com`, I would expect the IP's resolved to reflect as follows (if this command was executed 6 times consecutively):

```
$ curl -v http://app.cloud.example.com # execution 1
*   Trying 10.0.1.10...
...

$ curl -v http://app.cloud.example.com # execution 2
*   Trying 10.0.1.10...
...

$ curl -v http://app.cloud.example.com # execution 3
*   Trying 10.0.1.10...
...

$ curl -v http://app.cloud.example.com # execution 4
*   Trying 10.0.1.10...
...

$ curl -v http://app.cloud.example.com # execution 5
*   Trying 10.1.1.11...
[...]

PR #905, opened 2022-06-13
Configuration weight RR (1/4)

kuritka:
Related to #50
The weight (optional field) configuration goes into the GSLB strategy. This PR only aims to load and validate the weight configuration, which is specified in percentages. The weight distribution must be available to all clusters, so the weight field contains the weights of all regions.

```yaml
  strategy:
    type: roundRobin # Use a round robin load balancing strategy, when deciding which downstream clusters to route clients too
    weight:
      us: 60%
      eu: 15%
      za: 25%
```
gslb.spec.strategy.weight is not allowed for failover and geoip strategies. If it is not specified in roundRobin,
then it is not a weight RoundRobin but classic one.


k0da (approved):
LGTM

PR #911, opened 2022-06-21
Count with LoadBalancer IP in Service status

k0da:
On AKS for example, service type LoadBalancer doesn't expose Hostname
but direct IP.

We need to look at both properties


controllers/providers/assistant/gslb.go
```diff
@@ -101,14 +101,24 @@ func (r *Gslb) CoreDNSExposedIPs() ([]string, error) {
 		err := coreerrors.New(errMessage)
 		return nil, err
 	}
-	IPs, err := utils.Dig(lbHostname, r.edgeDNSServers...)
-	if err != nil {
-		log.Warn().Err(err).
-			Str("loadBalancerHostname", lbHostname).
-			Msg("Can't dig CoreDNS service LoadBalancer FQDN")
-		return nil, err
+	return extractIPFromLB(lb, r.edgeDNSServers)
```
kuritka, line 104: Just trying to avoid [else statement](https://www.youtube.com/watch?v=gKaGpNnR07c&ab_channel=FriendlyNeighborhoodCoder) for little better readibility.  
```go
if len(coreDNSService.Status.LoadBalancer.Ingress) == 0 {
		errMessage := "no LoadBalancer ExternalIPs are found"
		log.Warn().
			Str("serviceName", coreDNSService.Name).
			Msg(errMessage)
		return nil, coreerrors.New(errMessage)
}
lb = coreDNSService.Status.LoadBalancer.Ingress[0]
return extractIPFromLB(lb, r.edgeDNSServers)
```

k0da: Nice, updated

controllers/gslb_controller_test.go
```diff
@@ -849,6 +849,78 @@ func TestCreatesDNSNSRecordsForExtDNS(t *testing.T) {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      defaultCoreDNSExtServiceName,
+			Namespace: predefinedConfig.K8gbNamespace,
+			Labels: map[string]string{
+				"app.kubernetes.io/name": "coredns",
+			},
+		},
+	}
+	serviceIPs := []corev1.LoadBalancerIngress{
+		{IP: "1.1.1.1"}, // rely on 1.1.1.1 response from Cloudflare
```
somaritane, line 890: @k0da , are we really checking resolution by CloudFlare here or are these just mock test data?
https://blog.cloudflare.com/cloudflare-outage-on-june-21-2022/

k0da: This is just a test data

ytsarev: @k0da if it's just test data, then the comment describing that it's not should be removed. Looks like a copy-paste from https://github.com/k8gb-io/k8gb/blob/master/controllers/gslb_controller_test.go#L818 where we actually relying on Cloudflare's FQDN resolution

Issue #642, opened 2021-10-05
Deploy k8gb on MS Azure


jkremser:
Replicate our steps described in [local setup](https://www.k8gb.io/docs/local.html) on MS Azure's AKS. Including the
- `make test-failover`
- `make test-round-robin`
(the concepts captured by these Makefile targets)

Deliverables:
pull-request with the results describing the:
- reproduction steps (bash, makefile or markdown)
- issues that had to be solved in order to achieve this (platform specifics)
- all the steps that were necessary to make in the Azure web ui as well as the `az aks ...` CLI invocations

NOTE: the files should land somewhere under `/docs/platforms/azure/*`

You can use the free tier offering on [AKS](https://azure.microsoft.com/en-us/free).

If you need advice, we are on [Slack](https://cloud-native.slack.com/archives/C021P656HGB).

ManishS6:
@jkremser I would like to work on this.
Please assign this task to me.

ManishS6:
@ytsarev on it.

dmeytin:
@ytsarev, @ManishS6, any news on this issue?

ytsarev:
@dmeytin nope, are you interested in k8gb on Azure?

CC: @somaritane

somaritane:
@ytsarev thanks for upvoting ;) 
@dmeytin since we're using [ExternalDNS](https://github.com/kubernetes-sigs/external-dns) to communicate with DNS providers,  k8gb "[should just work](https://github.com/kubernetes-sigs/external-dns/tree/master/provider/azure)" with Azure DNS,
but we wanted to test E2E, hence were asking for help :)
My free Azure tier trial is over, have to find a workaround...
If you have access and are willing to check with our assistance, you're more than welcome.

ytsarev:
@somaritane while conceptually you are right, I guess we still need to extend the helm templates with azure related edgedns provider configuration similar to NS1 and Route53 before we can test anythng

somaritane:
@ytsarev yes, that too

dmeytin:
In addition to integration with Azure Traffic Manager (the azure alternative for Route53) the default cross-region communication will be slow. There is a need to make advanced configuration using [Vnet service endpoints](https://docs.microsoft.com/en-us/azure/virtual-network/virtual-network-service-endpoints-overview).

sebader:
I'd also be interested to see this running on Azure :)

dmeytin:
Any concrete plans to support Azure implementation?

[...]

PR #912, opened 2022-06-21
[WIP] Support for Azure DNS on AKS

sebader:
**Work in progress - just opening for transparency already**

This PR adds support for Azure DNS when running on AKS.

For now it only supports using the kubelet identity to authenticate against the Azure DNS zone with external-dns. One could add support for service principals as well, but that requires to store the client secret which I wanted to avoid for now. Once workload identities become available on AKS (successor of pod identity), that should be added, since using kubelet identity means, that all pods running on AKS will have access to that DNS zone. So it is not ideal but it is a first working version.

Working/implemented:
- [x] Helm chart updates to deploy. This mostly is about providing the required properties to external-dns

Missing/WIP
- [ ]  NS support for Azure DNS in external-dns. I managed this with a custom build of external-dns. https://github.com/kubernetes-sigs/external-dns/pull/2835
- [ ] Documentation and samples
- [x] Requires https://github.com/k8gb-io/k8gb/pull/911 to be merged and made available in k8gb

Closes https://github.com/k8gb-io/k8gb/issues/642


ytsarev:
@sebader how does it look with this work? Do you need any help? We really want Azure support :)

sebader:
> @sebader Sebastian Bader FTE how does it look with this work? Do you need any help? We really want Azure support :)

I didnt have time in the last few weeks to keep working further on this. However, I do have a working prototype running. Biggest blocker at the moment is my open PR on external-dns which adds NS support for Azure DNS. Any help to get that moving forward is much appreciated.

Apart from that it will be mostly documentation that needs to be written - and possibly end-to-end tests. I don't have any understanding so far how your test suite for k8gb works.

ytsarev:
@sebader we can temporarily fork external-dns and run the custom build to unblock ( it's not the first time we will do it).

sebader:
Ok. Let me try to find some time soon to continue here and build out the terraform templates to get the infrastructure set up for Azure. 

I'll obviously start with the public load balancing setup. But the original reason why I actually started all this work was to get to an internal global load balancer. I have that setup working, too. So as a later step I'll try to add the required setup for that as well.

ytsarev:
Both internal and public cases are super interesting, thanks!

ytsarev:
Hi @sebader,  we just switched to our own fork of external-dns here https://github.com/k8gb-io/k8gb/pull/1134 . The fork incorporates the support of NS record implementation for Azure. Do you want to revive the work on this PR? Please let me know if you need any help. It will be great to finalize your great work here 👍  Thanks a lot!

sebader:
this is great to hear @ytsarev! I'll see if I can grab some time (and refresh my memory first :D) I would also make the switch to Workload Identity if possible, now that this is (almost) GA on Azure

ytsarev:
Implemented by https://github.com/k8gb-io/k8gb/pull/1525  and https://github.com/k8gb-io/k8gb/pull/1593 . NS record support is implemented in [external-dns fork](https://github.com/k8gb-io/external-dns/) 

@sebader thank you so much for the initial implementation! Closing this PR as effectively completed.

PR #913, opened 2022-06-22
Drop irrelevant comment in newly added test

k0da:


jkremser (approved):
it could have been some artificial ip like `10.42.42.42` because it's not being used, but lgtm

kuritka (approved):

PR #914, opened 2022-06-22
Weight Round Robin (3/4) - annotations

kuritka:
related to #50

implemented code that adds an annotation to the local dnsEndpoint.

 - covered by unit-tests
 -  the annotation is as follows (formatted):
```json
{
  "eu":{
    "weight":35,
    "targets":[
      "10.10.0.1",
      "10.10.0.2"
    ]
  },
  "us":{
    "weight":50,
    "targets":[
      "10.0.0.1",
      "10.0.0.2"
    ]
  },
  "za":{
    "weight":15,
    "targets":[
      "10.22.0.1",
      "10.22.0.2",
      "10.22.1.1"
    ]
  }
}
```


controllers/dnsupdate.go
```diff
@@ -139,11 +145,16 @@ func (r *GslbReconciler) gslbDNSEndpoint(gslb *k8gbv1beta1.Gslb) (*externaldns.D
...
+	if err != nil {
+		return nil, err
+	}
+
 	dnsEndpoint := &externaldns.DNSEndpoint{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:        gslb.Name,
 			Namespace:   gslb.Namespace,
-			Annotations: map[string]string{"k8gb.absa.oss/dnstype": "local"},
+			Annotations: externalAnnotation,
```
k0da, line 157: You can drop annotations. Those are information only.

k0da: I guess you want to put your info into a Label

k0da: endpoint label to be more precise

kuritka: Hi thanx for review,
 
when I check dnsEndpointSpec I see it does not contain an annotation:
```txt
kubectl explain --api-version=externaldns.k8s.io/v1alpha1 DNSEndpoint.spec.endpoints

DESCRIPTION:

     Endpoint is a high-level way of a connection between a service and an IP

FIELDS:
   dnsName      <string>
     The hostname of the DNS record

   labels       <map[string]string>
     Labels stores labels defined for the Endpoint

   providerSpecific     <[]Object>
     ProviderSpecific stores provider specific config

   recordTTL    <integer>
     TTL for the record

   recordType   <string>
     RecordType type of record, e.g. CNAME, A, SRV, TXT etc

   setIdentifier        <string>
     Identifier to distinguish multiple records with the same name and type
     (e.g. Route53 records with routing policies other than 'simple')

   targets      <[]string>
     The targets the DNS record points to
```

So as you mentioned, to keep information in `endpoints` field I must refactor it so that the DNSEndpoint object looks something like this:

```yaml
apiVersion: externaldns.k8s.io/v1alpha1
kind: DNSEndpoint
metadata:
  name: app_host
spec:
  endpoints:
  - dnsName: app.bar.com
    labels:
      weight-eu-0-35: 10.10.0.1
      weight-eu-1-35: 10.10.0.2
      weight-us-0-50: 10.0.0.1
      weight-us-1-50: 10.0.0.2
      weight-za-0-15: 10.22.0.1
      weight-za-1-15: 10.22.0.2
      weight-za-2-15: 10.22.1.1
    recordTTL: 180
    recordType: A
    Targets:
      - 10.10.0.1
      - 10.10.0.2
      - 10.0.0.1
      - 10.0.0.2
      - 10.22.0.1
      - 10.22.0.2
      - 10.22.1.1
```
Do it through labels will be a bit more difficult to generate and parse, but it's nothing that can't be done...

 But I still don't understand why I can't keep the state of the weights directly in the `DNSEndpoint.metadata.annotations` annotation. `coredns-crd-plugin` won't be able to read it ?   Information like `weight-eu-0-35: 15` is nothing to group or selector by, or?

k0da: It just logical to keep endpoint specific metadata right next to the endpoint itself. As overall DNSEndpoint object could carry completely irrelevant endpoints. Hope that makes sense

kuritka: Okay, I'll refator  it to the labels then.

kuritka: @k0da , refactored; Labels instead of annotations, see PR description.

k0da (approved):
LGTM 👍

PR #917, opened 2022-06-28
Liqo Integration Tutorial

aleoli:
I'd like to add a new integration tutorial similar to the one provided for Admiralty, but for [Liqo](https://docs.liqo.io). Liqo is an open-source project that enables dynamic and seamless Kubernetes multi-cluster topologies, supporting heterogeneous on-premise, cloud, and edge infrastructures.

More details can be found in the online documentation: https://docs.liqo.io
The source code repo is: https://github.com/liqotech/liqo/tree/master

The [Liqo documentation](https://docs.liqo.io/en/stable/examples/global-ingress.html) already shows how the integration works, it would be nice to have a similar tutorial also in the K8GB documentation. Let me know if that makes sense and I'm happy to make any change per request.

I really appreciate any help you can provide.


ytsarev:
@aleoli I am super excited to see this integration, thank you so much! I want to test it end-to-end, will provide feedback soon :)

jkremser (approved):
Amazing tutorial! It works like a charm, I've gone through it and everything was smooth. Except 1 little bug, I've opened [PR for it already](https://github.com/liqotech/liqo/pull/1321) on your side.

I like the idea of virtual nodes representing the whole cluster, namespace extension and offloading. Very cool 👍 

btw. taints and tolerations should also work with Liqo, right?

aleoli:
@jkremser thanks!

Yes, exactly. The liqo virtual node is precisely like the physical ones from the Kubernetes perspective, so all standard taints, tolerations, affinities, and node drains/cordons work as expected. But you are working with an entire cluster instead of a single node

ytsarev:
@somaritane I still didn't have time to test it, but it should not block the merge

PR #918, opened 2022-07-01
Terratest checking Local DNSEndpoint

kuritka:
Please review after #914 merged

I had to prepare one extra terratest to verify that #914 is creating labels correctly. The terratests that control the weight will come later. Together with the test I extended the utils with DNSEndpoint class, which returns information about DNS endpoint.


controllers/dnsupdate.go
```diff
@@ -166,3 +161,19 @@ func (r *GslbReconciler) updateRuntimeStatus(gslb *k8gbv1beta1.Gslb, isPrimary b
 		m.UpdateFailoverStatus(gslb, isPrimary, isHealthy, finalTargets)
 	}
 }
+
+// getLabels map of where key identifies region and weight, value identifies IP.
+func (r *GslbReconciler) getLabels(gslb *k8gbv1beta1.Gslb, targets assistant.Targets) (labels map[string]string) {
```
k0da, line 166: Shouldn't that be named genLabels instead? getLables quite confusing.

kuritka: Hi @k0da, thx for good comment. I will try to explain: Yes it generates labels and therefore it should be called `genLabels`.  `getLabels` is **getter function** that **always returns** a label map. 

The main reason why I would choose `get` over `gen`. now the main reason why I would choose `get` over `gen`. Overall practice is getters don't change state. A getter returns or generates a value, but it doesn't change the state (for example, if the `getLabels` function somewhere inside would somehow change the endpoint, or gslb etc....). A **getter can't change anything**, just read anything and return the value or modified or generated value. 

This reduces the cognitive load a lot, because if something is a getter, the developer doesn't have to worry about the details of how the get works. He doesn't have to study the body of the function in order to have it change some instances.

kuritka:
follow-up for #914. Putting to draft after #914 merged.

PR #921, opened 2022-07-12
Terratest checking Local DNSEndpoint

kuritka:
I had to prepare one extra terratest to verify that #914 is creating labels correctly.
The terratests that control the weight will come later. Together with the test I extended the utils
with DNSEndpoint class, which returns information about DNS endpoint.


somaritane (approved):
lgtm

PR #924, opened 2022-07-20
Add unmarshalling to Percentage

kuritka:
if we do `kubectl apply -f ...` with GSLB having:
```yaml
   # valid values in glsb roundrobin
    weight:
      eu: 50%
      us: 50
```
the JSON masrshaller doesn't know whether values are number (50) or string (50%, "50").
The value is `type Percentage string`. Unmrashal function makes conversion, so 50, 50%, "50","50%" are valid.

I made several changes:
- change comment in CRD
- Unmarshall
- Fix one of the unit-tests
- Regenerate Mocks

### DIFF
--- a/api/v1beta1/gslb_types.go
+++ b/api/v1beta1/gslb_types.go
@@ -33,7 +33,7 @@ import (
 type Strategy struct {
 	// Load balancing strategy type:(roundRobin|failover)
 	Type string `json:"type"`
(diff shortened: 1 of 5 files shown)


jkremser (changes requested):
lgtm, but you will have prolly conflict in that crd file, coz of my previous pr, sorry for that. Just remove the change in that file from the PR + rebase and you should be good.

PR #926, opened 2022-07-21
Separate helm stuff from pure (/reusable) crd definitions

### LINKED ISSUE #925: CRD + helm
We [have](https://github.com/k8gb-io/k8gb/pull/818) that `k8gb.deployCrds` flag that controls whether crds are deployed or not, but it comes with a cost. Currently, one can't do something like 

```bash
k apply -f https://raw.githubusercontent.com/k8gb-io/k8gb/master/chart/k8gb/templates/crds/k8gb.absa.oss_gslbs.yaml
```
.. and see how it looks like

It would be nice to have the CRDs in their pure form w/o the helm curly brackets and then have another helm file that will conditionally include them. This will also simplify the makefile where we add those `{{ if foobar }} stuff {{end}}`

jkremser:
Fixes #925 

Using helm's `$files.Get` to include those two CRDs based on the `k8gb.deployCrds` flag. This way we can have pure CRDs that are also deployable w/o helm (or remotely referencing the github url)

side note: there is also a line with `$$(( $(CLUSTERS_NUMBER) + 1 ))`, because when doing full local setup, we are actually deploying n+1 clusters (+1 for the edge dns cluster)

### DIFF
--- a/chart/k8gb/crd/dns-endpoint-crd-manifest.yaml
+++ b/chart/k8gb/crd/dns-endpoint-crd-manifest.yaml
@@ -1,4 +1,3 @@
-{{- if .Values.k8gb.deployCrds }}
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
@@ -92,4 +91,3 @@ status:
     plural: ""
   conditions: []
   storedVersions: []
-{{- end }}
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ DNS_ZONE ?= cloud.example.com
 DEMO_URL ?= http://failover.cloud.example.com
 DEMO_DEBUG ?=0
 DEMO_DELAY ?=5
-GSLB_CRD_YAML ?= chart/k8gb/templates/crds/k8gb.absa.oss_gslbs.yaml
+GSLB_CRD_YAML ?= chart/k8gb/crd/k8gb.absa.oss_gslbs.yaml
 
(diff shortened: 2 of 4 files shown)


chart/k8gb/crd/dns-endpoint-crd-manifest.yaml
```diff
@@ -1,4 +1,3 @@
-{{- if .Values.k8gb.deployCrds }}
 ---
```
kuritka, line 1: does `---` needs to be here ? When brackets were used my controller-gen move it unintentionally.

kuritka (changes requested):
Awesome, thanks 👍 ! 

Please check my review comment

chart/k8gb/crd/dns-endpoint-crd-manifest.yaml
jkremser: I had a mistake in my previous version of the PR.. it was appending the crds for k8gb to that file, now it's only `>` instead of `>>`.

`---` at the beginning is what `controller-gen crd:crdVersions=v1 paths="./..." output:crd:stdout` actually produces, I am not sure what you mean by "controllergen can move it" (perhaps it was the issue with append instead of rewrite logic?)

kuritka: @jkremser , thx for clarification.

kuritka (approved):

PR #931, opened 2022-08-02
Bump Infoblox go client to v2 (2.1.2-0.20220727224704-88a5ba602dbf)

### LINKED ISSUE #920: Infoblox provider: upgrade infoblox client to recent version
Currently, we're using v1.1.1 version of the official Infoblox go client for the Infoblox provider.
whereas there has been a major `v2.0.0` release in the client
https://github.com/infobloxopen/infoblox-go-client/releases/tag/v2.0.0
with `v2.1.0` being the latest version 
https://github.com/infobloxopen/infoblox-go-client/releases/tag/v2.1.0

Makes sense to analyze the possible improvements/features, and upgrade to new version if that makes sense

jkremser:
Fix #920 

The latest released version of this go client doesn't have the full
support for CRUDING the TXT records (get is missing) so we need to depend on a git sha.

It was a major version upgrade so some API was changed and some method signatures were modified.

Namely:
 - inicialization of the client is different, it now requires the auth config to be passed as a separate struct
 - creating TXT record requires a view argument
 - `uint -> uint32` for ttl in method signatures
 - `*ibcl.ObjectManager -> ibcl.IBObjectManager`


jkremser:
I've also opened https://github.com/infobloxopen/infoblox-go-client/issues/177 on the client's repo side so that we can later switch to properly released version.

somaritane:
@jkremser It would be good to test the change with Infoblox instance.

ytsarev (approved):
LGTM assuming tested end-to-end

somaritane:
> LGTM assuming tested end-to-end

@ytsarev it's not, also it pins non officially released version.
Taking this and https://github.com/kubernetes-sigs/external-dns/issues/2945#issuecomment-1235732386 into account,
I suggest we park the PR until we sort all the points out

jkremser:
agreed on parking


> @jkremser It would be good to test the change with Infoblox instance.

https://github.com/k8gb-io/k8gb/issues/920#issuecomment-1238219332

Yep, for changes like this the mock tests that we currently have are pretty much useless. They changed the contract of some methods and I had to change the mocking, because they were making assumptions about internal implementations of the API.. like how many times some internal method is being called etc. It would be much more useful to have a way to spawn infoblox service and test it against it. However, if I get it right, there is no easy way to do that. I wasn't able to find any public container image for that.. In the official docs, they suggest to download the tar from their web ui and "docker load" it - wow. I am not a lawyer, but I think this altered image won't be freely usable in the public CI on GitHub. I can also check if the client library has some real tests against Infoblox.

ytsarev:
This one looks very stale, and I will close it for now.

PR #948, opened 2022-09-05
Refactoring SetupManager

### LINKED ISSUE #932: [BUG] GSLB is not updated when Ingress has change

kuritka:
related to #932

The setup manager is quite complex. We are currently solving a bug related to the fact that changes in ingress also change the GSLB and can cause cycles #932. I am doing the necessary refactoring (code is better structured, but the functionality remains the same), to make it easier to fix #932.

 - split SetupWithManager and Reconciliation into two files
 - encapsulate createGslbFromIngress
 - simplify ingressHandler
 - extracting parse strategy func


controllers/gslb_controller_setup.go
```diff
@@ -0,0 +1,197 @@
...
+	if err != nil {
+		log.Err(err).
+			Str("ingress", ingressToReuse.Name).
+			Str("gslb", gslb.Name).
+			Msg("Cannot set the Ingress as the owner of the Gslb")
+	}
+
+	log.Info().
+		Str("gslb", gslb.Name).
+		Msg("Creating new Gslb out of Ingress annotation")
```
jkremser, line 155: nit: I know there was no change in this msg, but can we put here:
"Creating a new Gslb out of Ingress with 'k8gb.io/strategy' annotation"  (we don't care about ingresses that don't have this annotation)

(`k8gb.io/strategy` == `strategy`)

controllers/gslb_controller_setup.go
```diff
@@ -0,0 +1,197 @@
...
+				return result, err
+			}
+		case primaryGeoTagAnnotation:
+			result.PrimaryGeoTag = annotationValue
+		}
+	}
+
+	if strategy == depresolver.FailoverStrategy {
+		if len(result.PrimaryGeoTag) == 0 {
+			return result, fmt.Errorf("%s strategy requires annotation", depresolver.FailoverStrategy)
```
jkremser, line 192: This will print:
"failover strategy requires annotation" which is not very helpful imho

can we put here

"failover strategy requires annotation k8gb.io/primary-geotag to be also present on the Ingress resource"

(`primaryGeoTagAnnotation` == `k8gb.io/primary-geotag`)

jkremser (changes requested):
lgtm otherwise, only the msg wording

controllers/gslb_controller_setup.go
jkremser, line 155:
```suggestion
		Msg(fmt.Sprintf("Creating a new Gslb out of Ingress with '%s' annotation", strategyAnnotation))
```

jkremser (approved):
👍 thanks

PR #949, opened 2022-09-05
Introduce new composite action that can be reused from other workflows to print the debug info

### LINKED ISSUE #946: Provide k8gb controller metrics on GHA
Sometimes it is very handy to have an idea what happens with k8gb during terratest run on GHA

jkremser:
Fix #946 

By default it runs with verbose being set to false, however user can re-run the action from the
github's web UI with the enable debug checkbox (see the img) being checked. Then more info is printed.

Screenshot 2022-09-05 at 14 26 41

example of usage from other workflows:
```yaml
...
  - name: Print debug info
    if: always()
    uses: ./.github/actions/print-debug
...
```
It also adds terratests results to `GITHUB_STEP_SUMMARY` file, which is then visible on the workflow "landing page".

It prints:
- events
- evn vars
- gslbs
- dnsendpoints
- metrics
- logs
- /etc/hosts

example runs:
- default - non-verbose: https://github.com/jkremser/k8gb/actions/runs/2993177731/attempts/1
- verbose: https://github.com/jkremser/k8gb/actions/runs/2993177731

### DIFF
--- a/.github/workflows/terratest-more-clusters.yaml
+++ b/.github/workflows/terratest-more-clusters.yaml
@@ -85,7 +85,12 @@ jobs:
           kubectl get no -owide --context=k3d-test-gslb3
 
       - name: Run Terratest
-        run: make terratest
+        run: |
+          mkdir -p ${{ github.workspace }}/tmp/terratest
+          set -o pipefail
+          echo "::group::Terratest logs"
+          make terratest | tee ${{ github.workspace }}/tmp/terratest/all.log
+          echo "::endgroup::"
 
       - name: Print debug info
         if: always()


kuritka (dismissed):
good idea 👍

jkremser:
@kuritka you've already approved this, but the CI was [failing](https://github.com/k8gb-io/k8gb/actions/runs/2993326056/jobs/4800739129), because the composite action wanted to show the parsed terratest logs, but they were not parsed when called from the context of `.github/workflows/terratest-more-clusters.yaml`. That's why I had to add this commit https://github.com/k8gb-io/k8gb/pull/949/commits/9126b180724df5c4b7966de17829176e56e556ed It's a separate commit so that you can verify what has been changed since the last review and it will be squashed in the end.

kuritka (approved):

jkremser:
thanks

PR #950, opened 2022-09-09
Helm values schema check

jkremser:
if there is a change to `values.yaml` or `values.schema.json`, this workflow will verify that `values.yaml` file is still valid against the corresponding JSON schema.

I had to remove those `required: []` lines because the python tool was failing on them. Semantically it's the same, these lines were actually redundant.

### DIFF
--- a/.github/workflows/helm_check-values-schema.yaml
+++ b/.github/workflows/helm_check-values-schema.yaml
@@ -6,8 +6,8 @@ on:
     branches:
       - master
     paths:
-      - 'chart/{repo}/values.yaml'
-      - 'chart/{repo}/values.schema.json'
+      - 'chart/k8gb/values.yaml'
+      - 'chart/k8gb/values.schema.json'
 jobs:
   check:
     name: 'Check values.yaml and its schema in PR'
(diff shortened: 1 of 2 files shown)


.github/workflows/helm_check-values-schema.yaml
```diff
@@ -0,0 +1,27 @@
+# thanks to https://github.com/giantswarm/devctl/blob/v5.9.0/pkg/gen/input/workflows/internal/file/check_values_schema.yaml.template for the idea
```
kuritka, line 1:
```suggestion
see: https://github.com/giantswarm/devctl/blob/v5.9.0/pkg/gen/input/workflows/internal/file/check_values_schema.yaml.template
```

kuritka (changes requested):

jkremser:
meh, applying your suggestion broke it... the comment is missing. Will fix that on Monday + I was actually happier with the original form since it's only based on this idea, however the implementaion is totally different

kuritka (approved):
At first glance, `thanks comment` looked more like you forgot it was there. If that's what you want, it's okay.

Issue #923, opened 2022-07-20
REFACTORING: Use DepResover mock in constroller tests


kuritka:
Use DepResover mock instead of `settings := provideSettings(t, predefinedConfig)` in constroller tests.

Using the settings object and calling reconcile is no longer sufficient. For example, the function works with `k8gb.absa.oss_v1beta1_gslb_cr.yaml` from which it generates a gslb for the tests. Any change to this file will 
cause the test to crash. We now have a depressolver mock (and a number of other mocks) that we can use for testing.

somaritane:
@kuritka what's the status of this one, can we close it?

kuritka:
Hi, not fully.  Depresolver mock for tests is ready (#951), but controller tests would be refactored, to use this mock  instead of direct instanptiating. Coverage requirement is still met but it's more about unify controller tests with mocks.

ytsarev:
Stale issue, feel free to reopen if needed

PR #951, opened 2022-09-09
Putting all mocks into one directory

kuritka:
related to #923,#932

I'm going to introduce new mocks for Manager and Client. I'm putting all mocks into one directory.


ytsarev (approved):
lgtm

PR #952, opened 2022-09-13
Skip the print debug action if the action is run redundantly

jkremser:
It looks like I broke the build with my last merged PR. The reason for that is that I was opening the pull request from my forked repo, so the action was run only once for the PR. However, we have that logic in our action that handles redundant action runs (one for push to feature branch on the repo, one for the PR, squashed commit vs the pr, etc.)

So this time I am opening the PR from the feature branch on this repo just to be sure it will be ok.


jkremser:
closing, because I was wrong about the root cause here.. the master was failing because of something else

PR #953, opened 2022-09-15
[WRR] remove percentage

### LINKED ISSUE #930: Remove percentage from GSLB weight
#50 has discussion of the deletion of the percentage type for weights. Let's replace it with portions defined as integer. 

- Also see route53 weight definition: https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/routing-policy-weighted.html

kuritka:
closes #930

remove percentage, using portions instead. see #930

Screen Shot 2022-09-15 at 10 58 51


jkremser (approved):
lgtm

PR #961, opened 2022-10-05
Bump k3s to v1.25.2

jkremser:
Lets see if terratests will fly on `k8s@1.25`

`--no-deploy=x,y,z` had to be changed to `--disable=..`, otherwise it fails (it's not only warning, it really fails to start) on:

`fatal log from node k3d-edgedns-server-0 (retrying 9/10): ctime="2022-10-05T13:05:05Z" level=fatal msg="no-deploy flag is deprecated. Use --disable instead."`


kuritka (approved):

ytsarev (approved):
lgtm, i renamed the header to k3s for clarity

PR #964, opened 2022-10-12
Expose NGINX Ingress values Path

kuritka:
NGINX ingress defines the [digest](https://github.com/kubernetes/ingress-nginx/blob/main/charts/ingress-nginx/values.yaml#L27) value. Not all environments work correctly with this value.

I am exposing the path to `nginx-ingress-values.yaml` from our Makefile so that I can easily override these values if needed. If `NGINX_INGRESS_VALUES_PATH` is not explicitly specified, the standard file path is used, as we use now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,6 +79,8 @@ STABLE_VERSION := "stable"
 # default bundle image tag
 BUNDLE_IMG ?= controller-bundle:$(VERSION)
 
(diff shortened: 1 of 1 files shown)


ytsarev (approved):
lgtm

jkremser (approved):

PR #967, opened 2022-10-13
remove digest completelly from NGINX settings

kuritka:
Reverting back NGINX values file path and remove digest completelly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,8 +79,6 @@ STABLE_VERSION := "stable"
 # default bundle image tag
 BUNDLE_IMG ?= controller-bundle:$(VERSION)
 
-NGINX_INGRESS_VALUES_PATH ?= deploy/ingress/nginx-ingress-values.yaml
-
 # options for 'bundle-build'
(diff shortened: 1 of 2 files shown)


jkremser (approved):

ytsarev (commented):
Now it is slightly cryptic. What is the reason for this change? what happens with the digest?

kuritka:
Hi, @ytsarev, thanks for question. The digests don't behave very well in the environment behind corporate proxies. The original idea was to expose the path to the NGINX values yaml file. In case you need to override a property (for any reason), you have choice to use the path to custom `nginx-values.yaml` instead having the one defined in the project.  This is how the previous commit was conceived. 

Because this seemed to be too general a solution, we had a short discussion about whether to overwrite the digest directly in `nginx values.yaml`.

I thought I'd throw it up on github and see.

kuritka:
okay, the more I think about it, I can leave it as it is... digest is defined by NGINX and custom environments would adapt.

Issue #965, opened 2022-10-12
node@12 has been deprecated warnings


jkremser:
The GH actions that use node12 produce these warnings in our pipeline:

```
Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: fkirc/skip-duplicate-actions
```

Also similarly `set-output` was deprecated and this should be used instead: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable

So once/if those actions are updated to 16 or later we should also update those versions to get rid of those warnings.
Namely it's:
- `actions/checkout` (3.1.0 should be ok)
- `fkirc/skip-duplicate-actions` (master is on [16](https://github.com/fkirc/skip-duplicate-actions/blob/master/package.json#L27) so probably 5.2.0 should be ok)
- `actions/setup-go` (https://github.com/actions/setup-go/blob/main/package.json#L37 anything 3.x.x should be ok)
- `docker/login-action` (2.1.0 should be ok)

kuritka:
Also found that issue in [coredns-crd-plugin](https://github.com/k8gb-io/coredns-crd-plugin/actions/runs/3234403056).

PR #969, opened 2022-10-14
Get rid of set-output commands in the workflows

jkremser:
Fixes https://github.com/k8gb-io/k8gb/issues/965

as suggested [here](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter) and [here](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/), the `echo "::set-output name=foo::${bar}"` should be replaced with `echo foo=bar >> $GITHUB_ENV` and then read as normal env var in all subsequent steps in the same job.

### DIFF
--- a/.github/actions/print-debug/action.yaml
+++ b/.github/actions/print-debug/action.yaml
@@ -10,7 +10,6 @@ runs:
   using: "composite"
   steps:
     - name: Check verbose
-      id: check
       shell: bash
       run: |
         if [[ "x${{ runner.debug }}" == "x1" ]]; then
@@ -26,7 +25,7 @@ runs:
         env | sort
         echo ---------------------------------------
         echo inputs.kubectl-context=${{ inputs.kubectl-context }}
-        echo verbose: ${{ steps.check.outputs.verbose }}
+        echo verbose: ${{ env.verbose }}
         echo "::endgroup::"
 
     - name: Terratest Logs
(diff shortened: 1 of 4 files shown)


kuritka:
Hi, can't this generate some kind of scope issues ?
 
ENV exists once for whole workflow run while step.<something>.<something>  points to property of concrete instance.

Anyway, using env certainly simplifies the usage.

jkremser:
@kuritka what do you mean by scope issues? I agree that the semantics is different here and previously it was better isolated in the step itself. I haven't tested the behavior, but I can imagine that it's appending key=value entries to some env file which is being sourced at the beginning of each subsequent step (it's not set for the step itself in which we do the thing).

So I think the latter occurrences will override the previous ones, but I didn't test it. Anyway we do not do that nor are relying on such behavior AFAICT. One can also set the env vars explicitly for whole job but also for each step, in this case I have no idea about the precedence rules, it's not well documented either. For sharing the date between the jobs this will not work at all, it's designed to work under one job and for that use-case one has to use job outputs.

kuritka (approved):
lgtm

somaritane:
@kuritka we're basically following GH standard recommendations here for deprecated output commands

somaritane (approved):
lgtm

kuritka:
@somaritane - thx I see. I read it in links @jkremser mentioned in description. I'm convinced that the change is right.

ytsarev (approved):
LGTM

jkremser:
sorry, I forgot to change the way how verbose flag was read in the print-debug action so I am adding a new commit: https://github.com/k8gb-io/k8gb/pull/969/commits/1a9c02c8612ae919b6a0eac01977eb830772c9e6 to [fix](https://github.com/k8gb-io/k8gb/actions/runs/3265343219/jobs/5367883615#step:10:4664) it.

The rest is the same as before, the review should be easier this way, in the end I will squash the changes.

somaritane (approved):

PR #1002, opened 2022-10-18
Permissions setting for the rest of the actions

ytsarev:
Follow up to #990

For some reason https://app.stepsecurity.io/ didn't manage to handle these actions automatically.

Signed-off-by: Yury Tsarev <yury@upbound.io>

### DIFF
--- a/.github/workflows/gh-pages.yaml
+++ b/.github/workflows/gh-pages.yaml
@@ -7,6 +7,8 @@ on:
     paths:
       - '**.md'
       - 'docs/**'
+permissions:
+  contents: read
 
 jobs:
   publish:
--- a/.github/workflows/gh-pages.yaml
+++ b/.github/workflows/gh-pages.yaml
@@ -12,6 +12,8 @@ permissions:
 
 jobs:
   publish:
+    permissions:
+      contents: write
     runs-on: ubuntu-20.04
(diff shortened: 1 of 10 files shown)


.github/workflows/gh-pages.yaml
```diff
@@ -7,6 +7,8 @@ on:
     paths:
       - '**.md'
       - 'docs/**'
+permissions:
+  contents: read
```
jkremser, line 11: i think we need write here, the implicit GITHUB_TOKEN is being used by EndBug/add-and-commit

.github/workflows/release.yaml
```diff
@@ -6,11 +6,13 @@ on:
       - "v*"
 
 permissions:
-  contents: write
+  contents: read
```
jkremser, line 9: this will break the next release, the contents: write is required by two other jobs in this workflow

jkremser (changes requested):

.github/workflows/gh-pages.yaml
ytsarev: we are setting 
```
    permissions:
      contents: write
```
on a job level

jkremser (approved):
thanks

PR #1003, opened 2022-10-19
Add Linux Foundation trademarks

somaritane:
Add k8gb.io copyright and Linux Foundation trademarks 
See https://clomonitor.io/docs/topics/checks/#trademark-disclaimer

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -59,6 +59,13 @@
           </a>
           sandbox project.
         </div>
+        <div>
+          <br>
+          Copyright © 2022 k8gb.io Project Authors. All rights reserved.
+          <br>
+          <br>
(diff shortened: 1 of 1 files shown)


kuritka (approved):

_layouts/default.html
```diff
@@ -59,6 +59,13 @@ <h2 class="project-tagline">{{ page.description | default: site.description | de
           </a>
           sandbox project.
         </div>
+        <div>
+          <br>
+          Copyright © 2022 k8gb.io Project Authors. All rights reserved.
```
ytsarev, line 64: we have `The k8gb Contributors.` in https://github.com/k8gb-io/k8gb/blob/master/.golic.yaml#L19 

Probably better stick with that version for consistency.

ytsarev (changes requested):

_layouts/default.html
somaritane: thank you @ytsarev, amended

jkremser (approved):

Issue #979, opened 2022-10-18
github-changelog-generator-action doing too many requests to git hub api


jkremser:
even though we use the [sinceTag](https://github.com/k8gb-io/k8gb/actions/runs/3274553640/jobs/5388234055#step:5:5) param for this action, 

It [performs](https://github.com/k8gb-io/k8gb/actions/runs/3274553640/jobs/5388234055#step:5:877) too many requests (1 per each PR so far -> 832). However, there is 1k requests per gh token limit so we will hit the limit with this action.

Workaround, use different gh token for this.

jkremser:
This has been done when fixing the failed release - https://github.com/k8gb-io/k8gb/commit/efbb9d2ee2c65a0168b5d7df7a552283c9fdaa1c

PR #1010, opened 2022-10-21
Change gh token to avoid 1k limit

AugustasV:
#979 Github token would solve 1k limit issue, because it's unique and created per job, after it's destroyed. More info: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -35,7 +35,7 @@ jobs:
           echo "previous_tag=${previous_tag}" >> $GITHUB_ENV
       - uses: heinrichreimer/github-changelog-generator-action@854576bb5274851427527d8199fba0061552dff5
         with:
-          token: ${{ secrets.CHANGELOG_GH_TOKEN }}
+          token: ${{ secrets.GITHUB_TOKEN }}
           project: k8gb
           sinceTag: ${{ env.previous_tag }}
           output: changes


jkremser:
following runs had it set to `GITHUB_TOKEN` and the release was failing because of it:
- https://github.com/k8gb-io/k8gb/actions/runs/3274553640/jobs/5388234055
- https://github.com/k8gb-io/k8gb/actions/runs/3273350012/jobs/5385449908

It's caused mainly by the `heinrichreimer/action-github-changelog-generator` action that pulls each PR and issue and what not: https://github.com/k8gb-io/k8gb/actions/runs/3273350012/jobs/5385449908#step:5:875

so it spends like 850 requests only by puling the existing prs. Even though the token's request budget is unique for each action run, we are not able to make it with 1 token for the release because of this action. Perhaps we should get rid of it and use the goreleaser for the changelog. It's not that detailed but, this is painful :D

AugustasV:
@jkremser or just use different kind of action? Maybe it could work better, without doing that much requests

jkremser:
@AugustasV yep, I am not happy with this action, it simply doesn't scale with the number of PRs.. once we have 1k PRs it can't work even with a dedicated own token. Goreleaser itself can produce changelog, but it can't aggregate fixed issues, merged PRs together.

It can however group things together based on the regexp in the commit msg - https://goreleaser.com/customization/changelog/#changelog so it's more a convention-based grouping.

jkremser:
The issue #979  is actually fixed for now, however this will fail again once we have enough PRs on the repo so if you (@AugustasV ) want to solve it properly by removing this action and using the goreleaser for it, feel free to create a issue for it and go ahead, but I am closing this PR

Issue #1018, opened 2022-11-03
Deprecate GSLB CRD


ytsarev:
The proposal is to deprecate core GSLB CRD and rely purely on standard Ingress annotations.

Motivation:
* Currently we already have Ingress -> Gslb creation logic here https://www.k8gb.io/docs/ingress_annotations.html
* It already led to the issues around associated reconciliation, annotation propagation and keeping Ingress and Gslb in sync.
* It is not clear what should be source of truth - Gslb or Ingress
* Gslb is just embedding/wrapping Ingress type within itself so relying purely on Ingress+Annotations should not lead to dramatic changes within the k8gb controller logic

ytsarev:
After some discussions with the maintainers team:

* We can actually keep Gslb CRD but make it read-only status exposing resource to keep contextual GSLB information that does not fit into the standard Ingress event stream
* We need to come up with reliable migration plan to not to break stuff for existing k8gb users

ytsarev:
We agreed to keep Gslb CRD and enable flexible referencing of various networking resources instead of full switch to Ingress.
First case of resource reference implemented in https://github.com/k8gb-io/k8gb/pull/1557

PR #1019, opened 2022-11-07
Fix 932 - ingress annotations accept external changes

### LINKED ISSUE #932: [BUG] GSLB is not updated when Ingress has change

kuritka:
closes #932
related to #1018

 - Fixed implementation. In some cases, when the ingress was changed, reconciliation was triggered, which changed the annotations back to the values calculated with GSLB. I modified the merge function to filter which values should be synced and which should not. In fact, it is only `k8gb.io/strategy` and `k8gb.io/primary-geotag` that are changed on initialization.
 - utit tests update
 - terratests are focused on repeated patching of annotations in ingress and examining ingress when patching annotations in GSLB


terratest/utils/extensions.go
```diff
@@ -582,6 +621,7 @@ func (i *Instance) logIfError(err error, message string, args ...interface{}) {
 	}
 }
 
+// kubectl -n k8gb patch deployment k8gb -p '{".spec.containers[0].image":"docker.io/kuritka/k8gb:temp005"}'
```
k0da, line 624: ????

PR #1020, opened 2022-11-08
Update GO.MOD Dependencies + Upgrade to GO1.19

### LINKED ISSUE #956: GO 1.19 migration - terratests
I succesfully migrated terratest in [coredns-crd-plugin](https://github.com/k8gb-io/coredns-crd-plugin/pull/43).

We can gradually migrate the first terratests, then k8gb.
 - upgarde go.mod
 - upgare github workflows 
 - consider refactor terratest utils into generic types

kuritka:
closes #956

I'm upgrading the dependency version because of dependabot. Unfortunately the change is huge and all linters started to report different types of errors. That's why I'm upgrading the linters, which in turn requires upgrading the GO version. With all these changes I had to make changes in the code and linters because of deprecated libraries and linters. I also had to make changes to our build pipelines.


ytsarev (commented):
Thanks for the tedious work.

What led to this nice ratio? Why `go.sum` became much shorter? Is it some go1.19 optimization?
image

kuritka:
@ytsarev , yes, many dependencies throughout the dependency tree are away.  With the  generics, I guess they also switched to new versions of libraries that don't have so many internal dependencies. It also plays a role that I'm changing both go.sum files, not just one as usually.

ytsarev (dismissed):
cool, thanks for the clarification and the brave PR :)

somaritane:
> @somaritane what do you mean by 'many potential changes' ? If terratest is green then we are covered or we do not trust our test suite?

@ytsarev  I'm thinking more about potential rebase issues for other open PRs we have, but if @kuritka is OK to manage them, I'm OK too.

somaritane:
@kuritka  we might need to update the docs for 1.19
https://github.com/k8gb-io/k8gb/blob/master/docs/local.md?plain=1#L25

kuritka:
@somaritane ,  documentation change amended. There is a risk that dependabot will discover some CVE issue, because I radically cut the `exclude` and `replace` sections in the `go.mod` files. I've checked the versions and it should be fine, but it's possible I might have missed something or something new will come up. In that case, a fix PR on CVE will be done.

somaritane:
> @somaritane , documentation change amended. There is a risk that dependabot will discover some CVE issue, because I radically cut the `exclude` and `replace` sections in the `go.mod` files. I've checked the versions and it should be fine, but it's possible I might have missed something or something new will come up. In that case, a fix PR on CVE will be done.

no worries @kuritka , by that time we might be running [renovate](https://github.com/k8gb-io/k8gb/pull/1021) instead :)

somaritane (approved):
lgtm 👍

PR #1021, opened 2022-11-08
Add renovate; remove dependabot

jkremser:
It also requires to go and click at https://github.com/marketplace/renovate because of the rights and PAT

Signed-off-by: Jiri Kremser <jiri.kremser@gmail.com>


somaritane:
> we need to run `clomonitor` linter to be sure we do not break our fancy rating :)
@ytsarev, is this something that runs automatically?

somaritane (approved):

PR #1112, opened 2023-04-03
Give external-dns assume role capabilities with route53 provider (helm)

eliasbokreta:
Hello,

I would like to add the capability for `external-dns` to assume role, using `route53` provider when the hosted zones are handled in a different AWS account by adding `--aws-assume-role` argument (see [this external-dns PR](https://github.com/kubernetes-sigs/external-dns/pull/524#issue-181256561)).
The change was successfully deployed on our clusters and is now, able to access the given hosted zone.

Besides this change, I've also fixed `route53.irsaRole` when disabled. The json schema validation regex was not allowing any other value than a valid ARN string.

If anything is unclear/missing please let me know 🙂


ytsarev (approved):
The changes look great. Thank you so much for this contribution!

PR #1123, opened 2023-04-08
Adjust concurrency for a terratest-based jobs

### LINKED ISSUE #1083: Mike pipelines lightweight by default
Currently we run very heavy terratest pipelines by default that are also brittle and frequently provide false positives, especially when running in parallel.

* Move heavy pipes behind PR comments
* Leave only essential ones to be executed by default

ytsarev:
* Terratest-based pipelines are frequently failing when running in parallel
* Make terratest pipelines running sequentially
* Make heavy terratest for n clusters pipeline triggered only when a special 'heavy-tests'  label set
* Fixes #1083

### DIFF
--- a/.github/workflows/terratest-more-clusters.yaml
+++ b/.github/workflows/terratest-more-clusters.yaml
@@ -18,6 +18,8 @@ on:
 permissions:
   contents: read
 
+concurrency: terratest-${{ github.head_ref || github.run_id }}
+
 jobs:
   skip-check:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


jkremser (approved):
lgtm, so if that `heavy-tests` label is added to PR then `terratest-n-clusters` will be run.. But they will not be run at the same time as upgrade testing or terratest, because they are in the same concurrency group. Nice!

It's still weird, those environments should be isolated from each other so not sure if concurrency being "sequential" can fix something here, but it can't hurt :D

k0da (approved):

PR #1134, opened 2023-04-11
Switch to own external-dns fork(again)

ytsarev:
* Unfortunately upstream external-dns reviews take too long and we have to switch to own fork again(we did it before)
* Main motivation at the moment is to unblock Azure NS support
* Fork incorporates https://github.com/kubernetes-sigs/external-dns/pull/2835

### DIFF
--- a/chart/k8gb/values.yaml
+++ b/chart/k8gb/values.yaml
@@ -47,6 +47,10 @@ k8gb:
 
 externaldns:
   # -- external-dns image repo:tag
+  # It is important to use the image from k8gb external-dns fork to get the full
+  # functionality. See links below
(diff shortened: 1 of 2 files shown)


kuritka (approved):
lgtm

chart/k8gb/values.yaml
```diff
@@ -47,7 +47,7 @@ k8gb:
 
 externaldns:
   # -- external-dns image repo:tag
```
k0da, line 49: We should put a comment that this is important. Cause if you override an image to a private registry source could be an upstream one.

ytsarev: Thanks, PR amended with extended comment.

PR #1141, opened 2023-04-18
Consume coredns crd plugin with WRR functionality

ytsarev:


k0da (approved):
LGTM

kuritka (approved):

kuritka (changes requested):
seems terratests fails, requires investigation.

ytsarev:
@kuritka can be related to WRR in action https://github.com/k8gb-io/k8gb/blob/master/terratest/examples/roundrobin_weight1.yaml#L39-L44 ? I assume that before this example was never properly evaluated by terratest with older k8s_crd plugin

kuritka:
@ytsarev , hi, probably. need to check terratests locally.

chart/k8gb/templates/coredns-cm.yaml
```diff
@@ -18,6 +18,7 @@ data:
             resources DNSEndpoint
             filter k8gb.absa.oss/dnstype=local
             negttl {{ .Values.k8gb.dnsZoneNegTTL }}
+            loadbalance weight
```
k0da, line 21: isn't it hardcoding particular method?

ytsarev: Would be great to get confirmation from @kuritka , but if I understand correctly it does standard round-robin if no WRR in Gslb spec, see https://github.com/k8gb-io/coredns-crd-plugin/pull/45

chart/k8gb/templates/coredns-cm.yaml
```diff
@@ -11,13 +11,12 @@ data:
         errors
         health
         ready
-        loadbalance round_robin
         prometheus 0.0.0.0:9153
         forward . /etc/resolv.conf
         k8s_crd {
-            resources DNSEndpoint
```
ytsarev, line 18: @kuritka it's not not immediately clear from https://github.com/k8gb-io/coredns-crd-plugin/pull/45 - why do we need to remove this config line?

chart/k8gb/templates/coredns-cm.yaml
kuritka: Hi @ytsarev, RR is only a subset of WRR.  If weights are not defined , the RR is automatically applied.  If weights are defined in CR, the WRR is used. If an error occurs for any reason, then the RR is used automatically. So there should stay `loadbalance weight `. 

The reason why I defines this field is for extendability. I was planning that in the future there will be a `loadbalance consistent` etc.

chart/k8gb/templates/coredns-cm.yaml
kuritka: `resources DNSEndpoint ` field was meaningless, we didn't work with it in any way.

kuritka (approved):
lgtm

chart/k8gb/templates/coredns-cm.yaml
ytsarev: @kuritka cool, I was mostly wondering why it actually breaking the terratest, but cool to get rid of it any case, thanks!

chart/k8gb/templates/coredns-cm.yaml
ytsarev: Perfect, thank you!

PR #1142, opened 2023-04-18
Update multi-cluster interoperability diagram

ytsarev:
* Correct the deviations from reality
* Document special `localtargets-$app` service FQDN
* Emphasize on cluster coredns-based cross-polling
* Avoid confusion of too much reliance on edgeDNS - we keep only NS records + Glue A records for Coredns host resolution there and not the application IPs
* Clarify that k8s_crd plugin == extended build of CoreDNS
* Remove `.drawio` derivative, diagrams.net can handle `.svg` directly


kuritka (approved):
lokks good, thx!

PR #1156, opened 2023-04-26
Don't let renovate bot to update the versions on the release pipeline

jkremser:
because the during the latest release there was an issue in it. Also using the same versions that were used for the release `0.10.0` which went fine.


ytsarev (approved):
LGTM

PR #1247, opened 2023-08-30
Add optionally servicemonitor service and object

kuritka:
Build on the top of @k0da https://github.com/k8gb-io/k8gb/pull/1245
..

<details>
---

  ### Heavy tests

  ### Debug tests

</details>

### DIFF
--- a/chart/k8gb/templates/metrics-service.yaml
+++ b/chart/k8gb/templates/metrics-service.yaml
@@ -1,0 +1,16 @@
+{{- if or .Values.k8gb.exposeMetrics .Values.k8gb.serviceMonitor.enabled -}}
+apiVersion: v1
+kind: Service
+metadata:
+  name: k8gb-metrics
+  labels:
+    {{- include "chart.labels" . | nindent 4 }}
+spec:
+  type: ClusterIP
+  ports:
+    - port: {{ include "k8gb.metrics_port" . }}
+      protocol: TCP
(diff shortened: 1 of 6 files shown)


chart/k8gb/templates/metrics-service.yaml
```diff
@@ -0,0 +1,16 @@
+{{- if .Values.k8gb.serviceMonitor.enabled -}}
```
ytsarev, line 1: we might want to provider the ability to expose metrics with service without enabling ServiceMonitor below(not everybody are using prometheus operator, but still might want to scrape the metrics )

ytsarev (commented):
looks great overall, just one design comment

chart/k8gb/templates/metrics-service.yaml
kuritka: @ytsarev , thx for comment. Something like this would make sense than: 

```yaml
`{{- if or .Values.k8gb.exposeMetrics .Values.k8gb.serviceMonitor.enabled -}}`
```

kuritka: @ytsarev @k0da  - I added `exposeMetrics` property.

k0da: 👍

ytsarev (approved):
thx!

PR #1255, opened 2023-09-04
Rancher Fleet support

kuritka:
Rancher fleet installs applications from the existing Helm Chart. During installation it injects at least properties into `global.fleet`. Because k8gb schema is too strict, `helm install` does not work because fleet injects its properties We solved this by deleting the schema and rebuilding Chart.

This PR whitelists the fleet object in globals, so k8gb can be installed directly.

### DIFF
--- a/docs/rancher.md
+++ b/docs/rancher.md
@@ -1,0 +1,33 @@
+# Integration with Rancher Fleet
+
+The K8gb has been modified to be easily deployed using [Rancher Fleet](https://fleet.rancher.io/). All you need to supply is a 
+[fleet.yaml](https://fleet.rancher.io/ref-fleet-yaml) file  and possibly expose the labels on your cluster.
+
+## Deploy k8gb to Target clusters
+The following shows the rancher application that will be installed on the target cluster.  The values `k8gb-dnsZone`, 
+`k8gb-clusterGeoTag`, `k8gb-extGslbClustersGeoTags` will be taken from the labels that are set on the cluster.
+
+```yaml
+# fleet.yaml
+defaultNamespace: k8gb
+kustomize:
+  dir: overlays/kustomization
+labels:
+  bundle: k8gb
+helm:
+  repo: https://www.k8gb.io
+  chart: k8gb
+  version: v0.11.4
+  releaseName: k8gb
+  values:
+    k8gb:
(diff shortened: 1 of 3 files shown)


k0da (approved):
LGTM

k0da:
@ytsarev @jkremser whats your take on this?

ytsarev (commented):
allowing totally unrelated value into the schema does not feel right. How exactly does it work with Fleet? Does it mutate the chart?

kuritka:
Hi @ytsarev, good question:  currently we deal with clusterLabels map injected by fleet, which injects its own values into values.yaml before installing the helmchart.  

```yaml
# values.yaml
global:
  fleet:
    clusterLabels:
      k8gb: "true"
      k8gb-ClusterGeoTag: "us"
      k8gb-ExtClusterGeoTags: "eu"
    ... 
k8gb:
  image: ...
  tag: ...
...
```
In fact, I don't know exactly what else fleet is able to inject in different edge-cases, so I provided it with an object. Outside of this bounded `fleet: {}` object, fleet can't do anything.

ytsarev (commented):
ok, thanks for the clarification. Can we extend the documentation mentioning the Rancher Fleet support, please?  It would be great to make people immediately aware of the integration.

k0da:
@ytsarev this is not something user controlls. This is internal usage of those fields. A structure evaluated prior helm call and processes helm values based on this. Take it as internal passthrough storage for particular cluster.

without it. Helm just fails to validate values, as schema didn't allow extra values to be passed in.

ytsarev:
@k0da the statement that k8gb was tested and helm chart adapted for Rancher fleet setup will be anyway useful for users and for the project

k0da:
👍

k0da (approved):

kuritka:
@k0da , @ytsarev -  Hi, I extended with simple doc

k0da (approved):

docs/rancher.md
ytsarev, line 1: this is great. Can we link this manual in the main README, probably extending https://www.k8gb.io/#installation-and-configuration-tutorials ?

kuritka: @ytsarev , fixed, thx.

ytsarev (approved):
looks great, thanks for amending the docs 👍

k0da (approved):

PR #1257, opened 2023-09-06
Removing @somaritane from CODEOWNERS

somaritane:
Let's make this official


ytsarev (approved):
@somaritane, thanks for all the contributions!

PR #1266, opened 2023-09-11
Change Extra plugins from object to string

kuritka:
I need to change the schema from object to string.

fir the `values.yaml`configuration:

before:

```yaml
  coredns:
    extra_plugins: {log}
```

```yaml
cloud.example.com:5353 {
    errors
    health
    log: {}
    prometheus 0.0.0.0:9153
    forward . /etc/resolv.conf
    k8s_crd {
        negttl 300
        loadbalance weight
    }
```

after:

```yaml
  coredns: 
    extra_plugins: |
       log
```

```yaml
cloud.example.com:5353 {
    errors
    health
    log 
    prometheus 0.0.0.0:9153
    forward . /etc/resolv.conf
    k8s_crd {
        negttl 300
        loadbalance weight
    }
```

### DIFF
--- a/chart/k8gb/values.schema.json
+++ b/chart/k8gb/values.schema.json
@@ -337,7 +337,7 @@
             "additionalProperties": false,
             "properties": {
                 "extra_plugins": {
-                    "type": "object"
+                    "type": "string"
                 }
             },
             "title": "k8gbCoredns"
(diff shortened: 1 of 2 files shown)


k0da (approved):

ytsarev (approved):
good catch

Issue #147, opened 2020-06-27
Support Service of type LoadBalancer to enable global load balancing on L4 


ytsarev:
Currently k8gb fully relies on underlying Ingress that it controls. It has limitations of exclusive http/https support so services on another ports and another protocols are not covered by Gslb operations.
We can change that by enabling k8gb to control underlying Service of type LoadBalancer. It will create mechanism to expose any tcp/udp service on any port and protocol. We definitely should keep current Ingress operations for L7 http/https .

Service enabled spec might look like:
```
apiVersion: k8gb.absa.oss/v1beta1
kind: Gslb
metadata:
  name: test-gslb
  namespace: test-gslb
spec:
  loadBalancer:
    host: app1.cloud.example.com
    serviceName: app # Service with type: LoadBalancer
  strategy:
    type: roundRobin
```

abaguas:
During the contribfest today an additional usecase was pointed out for Services of type ClusterIP. In their setup the service IP addresses are routable in the WAN.

When implementing this feature we should support both services of type LoadBalancer and ClusterIP

13567436138:
```
cat << EOF | kubectl --kubeconfig ~/.kube/minio-01 apply -f - 
apiVersion: k8gb.absa.oss/v1beta1
kind: Gslb
metadata:
  name: minio-svc-gslb
  namespace: minio
spec:
  ingress:
    backend:
      service:
        name: minio
        port:
          number: 9000
  strategy:
    type: failover 
    primaryGeoTag: cn-hz
EOF

cat << EOF | kubectl --kubeconfig ~/.kube/minio-02 apply -f - 
apiVersion: k8gb.absa.oss/v1beta1
kind: Gslb
metadata:
  name: minio-svc-gslb
  namespace: minio
spec:
  ingress:
    backend:
      service:
        name: minio
        port:
          number: 9000

  strategy:
    type: failover 
    primaryGeoTag: cn-hz
EOF
```


when will this support ，I do not know how to loadbalancer tcp traffic

abaguas:
This feature was implemented in https://github.com/k8gb-io/k8gb/pull/2029 and will be released in v0.16.0

PR #1363, opened 2023-12-20
add support for svc backend

tanujd11:
Adds LoadBalancer service as part of the backend for K8GB. Relate Issue is: https://github.com/k8gb-io/k8gb/issues/147 

<details>
---

  ### Heavy tests

  ### Debug tests

</details>


ytsarev (commented):
It is exciting, thanks a lot for your contribution!

Can you please add

* Reconciliation test covering new capability to https://github.com/k8gb-io/k8gb/blob/master/controllers/gslb_controller_reconciliation_test.go
* e2e tests to https://github.com/k8gb-io/k8gb/tree/master/terratest
* document the new feature in https://github.com/k8gb-io/k8gb/tree/master/docs

Thanks!

controllers/dnsupdate.go
```diff
@@ -43,9 +43,19 @@ func (r *GslbReconciler) gslbDNSEndpoint(gslb *k8gbv1beta1.Gslb) (*externaldns.D
...
+	localTargets := []string{}
+	if gslb.Spec.Ingress != nil {
+		localTargets, err = r.DNSProvider.GslbIngressExposedIPs(gslb)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if gslb.Spec.LoadBalancer != nil {
+		localTargets, err = r.DNSProvider.GslbServiceExposedIPs(gslb)
```
jkremser, line 55: this way, if both `gslb.spec.loadBalancer` and `gslb.spec.ingress` are specified in the CR, then the `localTargets` will be obtained using the LB section. I am not sure how it should be designed, to be honest. Does it actually make sense that gslb CR is backed by both at once? If not we could implement the [`x-kubernetes-validations`](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/2876-crd-validation-expression-language/README.md) to catch this during resource admission, otherwise I'd go with simple `if - else if` here for now.

jkremser (commented):
👍

controllers/dnsupdate.go
tanujd11: yeah, I think adding a validation for it is better so that both can't be specified. It is a cleaner and better approach then making some blunders while missing some case in if - else approach. It is better to be clear at the start of the implementation and state that both can't be used at once. WDYT?

jkremser: makes sense to me. Would you mind exploring those [x-kubernetes-validations](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/2876-crd-validation-expression-language/README.md)? There should be an annotation for kubebuilder for that ([link](https://github.com/kubernetes-sigs/kubebuilder/issues/2798#issuecomment-1310275396))

something like 

```
// +kubebuilder:validation:XValidation:rule=( has(self.loadBalancer) && !has(self. ingress) ) || ( !has(self.loadBalancer) && has(self. ingress) )
```

ytsarev:
Hi @tanujd11 ! Do you plan to finalize this PR?

ytsarev:
Hi @tanujd11 , do you want to update this PR after we got #1557 implemented? Thanks a lot!

ytsarev:
I will close this PR as the codebases progressed heavily since then. Thanks a lot for the contribution attempt, I really appreciate it!

PR #1370, opened 2023-12-22
renovate: fix new line + don't group the pr with library bumps

jkremser:
using own `groupName` so that pr is not combined with the `update all non-major dependencies` one + fixing the new line issue - https://github.com/k8gb-io/k8gb/pull/1052/files#diff-cb650ef911579f6ff1e25b1b7ab7e70a96c1a7e609cf5598a0635bca1b6984acR76

this has been tested on my fork and the resulting pr from renovate bot should look like this: https://github.com/jkremser/k8gb/pull/83/files

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -28,6 +28,12 @@
       "groupSlug": "all-minor-patch"
     }, {
         "matchPackageNames": [
+        "k8gb-io/k8gb",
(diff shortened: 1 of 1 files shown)


ytsarev (approved):
Thanks!

PR #1308, opened 2023-10-18
add k8gb to community-operators-prod

jkremser:
This should now open two PRs: 
- to `k8s-operatorhub/community-operators`
- to `redhat-openshift-ecosystem/community-operators-prod`

it also fixes the "helm smoke tests" - context: https://cloud-native.slack.com/archives/C021P656HGB/p1697629319760079

<details>
  <summary>HOW TO RUN CI</summary>
---

  By default, all the checks will be run automatically. Furthermore, when changing website-related stuff, the preview will be generated by the netlify bot.

  ### Heavy tests
  Add the [`heavy-tests`](/k8gb-io/k8gb/issues?q=is%3A*+label%3Aheavy-tests) label on this PR if you want full-blown tests that include more than 2-cluster scenarios.

  ### Debug tests
  If the test suite is failing for you, you may want to try triggering `Re-run all jobs` (top right) with [debug logging](https://docs.github.com/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging) enabled. It will also make the [print debug](/k8gb-io/k8gb/blob/master/.github/actions/print-debug/action.yaml) action more verbose.

</details>

PR #1371, opened 2023-12-22
Don't let OLM job invocations fail the helm pipeline job + fix the second invocation

jkremser:
name of the workflow to be called should be "`OLM bundle and PR`" (the `workflow: foobar` part)
mea culpa, i missed that in https://github.com/k8gb-io/k8gb/pull/1308


ytsarev (approved):
Thanks for the fix!

PR #1422, opened 2024-01-28
Add support of secret based AuthN/Z for Route53

jkremser:
The preferred mechanism of AuthN,Z for aws comm from k8s cluster is using the service account -> role mapping, however in non-managed k8s (no EKS) environments where people just use pure AWS's EC2 instances with kubeadm or using cluster-api (non EKS mode), there the irsa controller (or how is it called today) is missing.

So it would be nice to have a way to make k8gb work also w/o setting this up. Another use-case for this would be a multi-cluster setup where all clusters need to talk to route53, we can't assume the irsa controller in GCP or in Azure.

it's one of the documented ways for externaldns - [docs](https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/aws.md#static-credentials)

<details>
---

  ### Heavy tests

  ### Debug tests

</details>


ytsarev (approved):
LGTM, @jkremser  I assume you tested it e2e :)

k0da:
@jkremser just playing with IAM Roles Anywhere. I had to use similar approach across various projects. It was done by extraVolumes and extraVolumeMounts. Would it be better to just allow user to directly specify volume and volumeMounts?

PR #1428, opened 2024-01-30
Allow to change dnsPolicy of ExternalDNS deployment

jkremser:
why?

When deploying k8gb to vanilla k8s clusters that were created using cluster API and kubeadm (no EKS, no managed control planes).

The ExternalDNS deployment has an issue with resolving `route53.amazonaws.com`. By default the dnsPolicy if not specified is `ClusterFirst` so this PR is a no-op in terms of backward compatibility, but if changed to `Default` this fixed the previous situation for me. I guess, this can be solved also by tweaking the dns server for this k8s cluster or by providing own corefile for coredns, but it can't hurt either to allow this easy fix. 

[docs](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy)
`Default` ... use host for resolving the dns names
`ClusterFirst` ... Any DNS query that does not match the configured cluster domain suffix, such as "www.kubernetes.io", is forwarded to an upstream nameserver by the DNS server. Cluster administrators may have extra stub-domain and upstream DNS servers configured.

### DIFF
--- a/chart/k8gb/templates/external-dns/external-dns.yaml
+++ b/chart/k8gb/templates/external-dns/external-dns.yaml
@@ -50,6 +50,7 @@ spec:
           - mountPath: /etc/krb5.conf
             name: kerberos-config-volume
             subPath: krb5.conf
+      dnsPolicy: {{ .Values.externaldns.dnsPolicy }}
       volumes:
         - name: kerberos-config-volume
           configMap:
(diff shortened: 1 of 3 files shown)


kuritka (approved):

ytsarev (approved):
Good find

PR #1525, opened 2024-04-02
Documented Azure DNS deployment

infbase:
Updated Helm chart, docs and examples for connecting to Azure Public DNS service.

Users should now be able to connect external-dns to an existing Azure Public DNS zone by setting the correct helm values, as demonstrated in the example deployment added to the docs in this PR.

Related to issue #642


chart/k8gb/templates/_helpers.tpl
```diff
@@ -156,3 +162,24 @@ k8gb-{{ .Values.k8gb.dnsZone }}-{{ .Values.k8gb.clusterGeoTag }}
...
+{{- if .Values.azuredns.enabled -}}
+{
+  "tenantId": "{{ .Values.azuredns.tenantId }}",
+  "subscriptionId": "{{ .Values.azuredns.subscriptionId }}",
+  "resourceGroup": "{{ .Values.azuredns.resourceGroup }}",
+  {{- if .Values.azuredns.aadClientId -}}
+  "aadClientId": "{{ .Values.azuredns.aadClientId }}",
+  {{- end -}}
+  {{- if .Values.azuredns.aadClientSecret -}}
+  "aadClientSecret": "{{ .Values.azuredns.aadClientSecret }}",
```
ytsarev, line 176: Does it mean we will have the client secret as plain text in the `values.yaml`? We probably should keep it in the `Secret`

chart/k8gb/templates/external-dns/external-dns-azure-auth.yaml
```diff
@@ -0,0 +1,9 @@
+{{- if and .Values.azuredns.enabled .Values.azuredns.createAuthSecret }}
```
ytsarev, line 1: we probably should create this Secret with Credentials outside of the helm chart and point to it from the `values.yaml` for k8gb to pick up

chart/k8gb/values.schema.json
```diff
@@ -600,6 +603,58 @@
...
+                "resourceGroup": {
+                    "type": "string",
+                    "minLength": 1
+                },
+                "aadClientId": {
+                    "type": "string"
+                },
+                "aadClientSecret": {
+                    "type": "string"
+                },
```
ytsarev, line 638: As pointed out above, I have concerns about propagating `aadClientId` and `aadClientSecret` as plaintext values. Can we keep them in some input Secret? I appreciate that it might be tricky as we construct the external-dns Secret data in `external-dns.azure-credentials` template function.

ytsarev (commented):
@infbase thank you so much for the contribution! Overall, it looks great, my main concern is that the Azure `clientSecret` is getting exposed as plain text. Can we find a way to avoid it? Happy to collaborate on this topic and try to find a solution together.

ytsarev:
@infbase as discussed in community , let's provide ability to reference existing secret for more secure scenario

ytsarev (dismissed):
To not to bloat this PR, i created separate  https://github.com/k8gb-io/k8gb/issues/1585 to track the existing secret reference case.

@infbase thanks a lot for your contribution!

PR #1549, opened 2024-05-01
Queue reconciliation of all GSLBs that reference the same endpoint

abaguas:
The controller watches Endpoint resources to quickly update the health status of GSLBs. When an Endpoint changes, the controller fetches all GSLB resources and checks if they reference this Endpoint. However, there is a bug where the loop only queues for reconciliation the last GSLB that references the Endpoint.
In this PR the bug described above is fixed so that the endpoint handler is capable of queuing multiple GSLB resources for reconciliation.

---

  ### Manual Test

Best done after setting a long `reconcileRequeueSeconds` (e.g. 600 seconds) to prevent GSLBs from being reconciled more often.

* Create multiple GSLBs with different domain names using the same backend service.
* Scale to 0 the sts/deploy behind the backend service to trigger an update of the endpoints.
* Verify the DNS records are removed for all GSLB domains.


ytsarev (approved):
Wow, great catch!

PR #1555, opened 2024-05-13
Rancher changed the suffix from k3s2 back to k3s1

jkremser:
https://hub.docker.com/r/rancher/k3s/tags

I guess the k3s2 suffix was some temporary thing, but we can support both just in case it's gonna be there again


ytsarev (approved):
Thanks!

PR #1598, opened 2024-06-11
Bump ControllerGen, CRD 

kuritka:
I'm upgrading the kube-linter to the latest version. I need to do this before I bump the depenencies in the go.mod files.  This change modifies the CRD! 

..

<details>
---

  ### Heavy tests

  ### Debug tests

</details>


ytsarev:
@k0da CRD change, I would appreciate extra pair of eyes in review here, thanks :)

k0da:
@kuritka do we really need to change CRD?

kuritka:
@k0da , good question. Not IMO. 
Curretly  CRD is ochanged by controller-gen (in our case automatically when `make check` but can be disabled).

chart/k8gb/crd/k8gb.absa.oss_gslbs.yaml
```diff
@@ -77,26 +83,31 @@ spec:
                         - kind
                         - name
                         type: object
+                        x-kubernetes-map-type: atomic
```
ytsarev, line 86: SSA https://kubernetes.io/docs/reference/using-api/server-side-apply/

chart/k8gb/crd/k8gb.absa.oss_gslbs.yaml
```diff
@@ -366,9 +372,3 @@ spec:
     storage: true
     subresources:
       status: {}
-status:
-  acceptedNames:
-    kind: ""
-    plural: ""
-  conditions: []
-  storedVersions: []
```
ytsarev, line 374: looks like a fix for https://github.com/kubernetes-sigs/controller-tools/issues/456 , good to proceed 👍

chart/k8gb/crd/k8gb.absa.oss_gslbs.yaml
ytsarev:
```
Applicable to maps. atomic means that the map can only be entirely replaced by a single manager. granular means that the map supports separate managers updating individual fields.
```

Looks matching our case

ytsarev (approved):
I've double-checked the CRD changes, see inline comments. Looks safe to upgrade to me.

PR #1615, opened 2024-06-26
Increase number of renovate bot PRs

abaguas:
Since #1610 was merged the number of PRs created by renovate increased, as bumps of libraries on version 0.x are handled in a dedicated PR. The maximum number of concurrent PRs is 5 which results in rate-limiting: https://github.com/k8gb-io/k8gb/issues/1048. The proposal is to increase it to 15 so that we can see and tackle the PRs containing potential breaking changes.

In addition, the `update all non-major dependencies` group should very rarely contain breaking changes (#1052) so I propose to turn on automerge (if the pipeline is green ofc). This should save the maintainers some precious minutes.

PR #1625, opened 2024-06-30
remove non-working auto merge

abaguas:
In #1615 auto merge was enabled for [all non-major dependencies](https://github.com/k8gb-io/k8gb/pull/1052#event-13317496471). However, the PR cannot be merged without an approving review. The missing piece would be the following app: https://github.com/apps/renovate-approve. It would approve all PRs created by the renovate bot that have auto merge enabled. Unfortunately, it cannot do it for projects that have CODEOWNERS setup, which is our case:
> Important note: Due to a GitHub limitation, it is not possible to assign any app like this one as a CODEOWNER, so unfortunately this bot won't work that way if you have CODEOWNERS set up.

Since auto merge does not work in our setup this PR removes its configuration to avoid any confusion.

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -27,7 +27,6 @@
       "matchCurrentVersion": "!/^v?0/",
       "groupName": "all non-major dependencies",
       "groupSlug": "all-minor-patch",
-      "automerge": true,
     }, {
         "matchPackageNames": [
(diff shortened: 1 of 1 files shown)


ytsarev (approved):
Thanks for this investigation 👍

PR #1652, opened 2024-07-10
Fix ingress Annotations replacement loop

### LINKED ISSUE #932: [BUG] GSLB is not updated when Ingress has change

kuritka:
Annotation implementation change, terratests covered, test change.

closes #932
closes #1019
related to #1018

 - When I create GSLB - not k8gb annotations are not copied to ingress. Annotations that are directly related to k8gb are now created in ingress.

 - When I modify an annotatiosn in ingress, the annotations are not copied to the GSLB

 - If I change an annotation that is related to k8gb in ingress, the GSLB as source of truth returns the annotation to it's state

 - When I change an annotation in ingress that is not related to k8gb, reconciliation is not triggered and the annotation remains in ingress and is ignored by k8gb. 

 - When I create a GSLB from an ingress, the annotations are not copied to the GSLB. 

- I simplify WeightRoundRobin terratests from waiting cycles, the waiting cycles are already implemented in WaitForApp.


controllers/gslb_controller_setup.go
```diff
@@ -123,9 +138,9 @@ func (r *GslbReconciler) createGSLBFromIngress(c client.Client, a client.Object,
 	}
 	gslb := &k8gbv1beta1.Gslb{
 		ObjectMeta: metav1.ObjectMeta{
-			Namespace:   a.GetNamespace(),
-			Name:        a.GetName(),
-			Annotations: a.GetAnnotations(),
+			Namespace: a.GetNamespace(),
+			Name:      a.GetName(),
+			//			Annotations: a.GetAnnotations(),
```
k0da, line 143: Is this intended to be commented?

ytsarev (approved):
This is great, I performed the e2e testing validating annotation propagation behaviour especially focusing on

> If I change an annotation that is related to k8gb in ingress, the GSLB as source of truth returns the annotation to it's state

> When I change an annotation in ingress that is not related to k8gb, reconciliation is not triggered and the annotation remains in ingress and is ignored by k8gb.

```yaml
k -n test-gslb get ing test-gslb-failover -o yaml
kind: Ingress
metadata:
  annotations:
    k8gb.io/primary-geotag: eu # this is getting reverted according to Gslb spec if changed manually on Ingress
    k8gb.io/strategy: failover
    annotation: test # this is staying intact
```

PR #1672, opened 2024-07-29
Integrate istio ingress

abaguas:
Fixes [#552](https://github.com/k8gb-io/k8gb/issues/552) The K8GB controller can now speak istio 🎉

### User config

In #1557 the groundwork was done to be able to reference any ingress type.
The only implementation so far was `Kubernetes Ingress resources`. With this PR we bring to life the second implementation without touching any of the core logic: [istio virtual services](https://istio.io/latest/docs/reference/config/networking/virtual-service/). The configuration looks as follows:
```
spec:
  resourceRef:
    kind: VirtualService
    matchLabels:
      app: istio
```
### CRD changes

The previous API required CRD changes on every new integration because it was depending on a key with its name:
```
spec:
  resourceRef:
     istio:
        ...
     ingress:
        ...
```
To keep the CRD stable the structure is changed to:
```
spec:
  resourceRef:
    apiVersion: foo
    kind: bar
```

### How istio ingress works

Istio uses [istio ingress gateways](https://istio.io/latest/docs/tasks/traffic-management/ingress/ingress-control/) for ingress. The users send traffic to the gateway, and then the gateway forward traffic to the application.
The ingress gateway is configured as follows:
* [Gateway resources](https://istio.io/latest/docs/reference/config/networking/gateway/) configure the ports open on the gateway. This resource contains also a list of hosts (Host header or SNI) that it expects on that port.
* [Virtual Service resources](https://istio.io/latest/docs/reference/config/networking/virtual-service/) configure to which kubernetes service the traffic should be routed, after it has been accepted by the ingress gateway. It contains a list of hosts (Host header or SNI), the service it should route the traffic to, and a reference to a `Gateway` resource.

### How the k8gb controller integrates with istio

The k8gb controller is interested in three pieces of information:  the domain of the endpoint and the IP address of the cluster ingress to create DNS entries; the kubernetes service in front of the application to monitor the health of the `endpoints`.
From the `VirtualService` resource all this information can be fetched, that is why this is the resource referenced by the GSLB. The domain and service are immediately available on the `VirtualService` resource. The IP address needs to be fetched from the istio ingress controller's LoadBalancer service which can reached by following -> `VirtualService.spec.gateways` -> `Gateway.spec.selector`.

### Playground setup

To showcase this new integration the playground was extended with an istio ingress controller running side by side with the nginx ingress controller. The nginx ingress controller is still exposed on ports 80 and 81 (for clusters `k3d-test-gslb1` and `k3d-test-gslb2` respectively), and the istio ingress controller is exposed on ports 8080 and 8081.
In addition, since the podinfo pod behind the istio ingress requires a sidecar, a new namespace `test-gslb-ingress` was created to host this application and the respective GSLB resources.

### Helm chart

The helm chart is adapted with a feature flag to enable the necessary RBAC read permissions.

### Testing

These changes are backed by unit and e2e tests. The framework needed to be slightly adapted to support applying also `VirtualService` and `Gateway` resources but the core logic of both the frameworks and the tests remains.
In addition, a workflow was created to run the `terratest` suit for the `v1beta1` apiVersion of istio, since we plan to support both `v1beta1` and `v1`.
The PR looks big, but most of the changes are new manifests for the e2e testing setup.


api/v1beta1/gslb_types.go
```diff
@@ -45,6 +45,8 @@ type Strategy struct {
 type ResourceRef struct {
 	// Ingress selects a kubernetes.networking.k8s.io/v1.Ingress resource
 	Ingress metav1.LabelSelector `json:"ingress,omitempty"`
+	// IstioVirtualService selects a networking.istio.io/v1.VirtualService resource
+	IstioVirtualService metav1.LabelSelector `json:"istioVirtualService,omitempty"`
```
ytsarev, line 49: Now I am slightly concerned that we have to change/extend the API everytime we add new integration.

Is it possible to make resourceRef more dynamic and clear, something like

```
  resourceRef:
      kind: Ingress
      matchLabels:
        app: roundrobin-test
```

```
  resourceRef:
      kind: VirtualService
      apiVersion: networking.istio.io/v1beta1
      matchLabels:
        app: roundrobin-test
```

It will make obvious what kind of exact gvk we are selecting and will keep the main Gslb CRD stable.

controllers/refresolver/istiovirtualservice/istiovirtualservice.go
```diff
@@ -0,0 +1,219 @@
...
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"strings"
+
+	k8gbv1beta1 "github.com/k8gb-io/k8gb/api/v1beta1"
+	"github.com/k8gb-io/k8gb/controllers/logging"
+	"github.com/k8gb-io/k8gb/controllers/utils"
+	istio "istio.io/client-go/pkg/apis/networking/v1beta1"
```
ytsarev, line 30: Related to the proposal of the resourceRef spec modification - how we handle the future upgrade situation of `v1beta1` to `v1` in this case? and maybe supporting both versions in parallel for the time being?

ytsarev (commented):
Made a first round of review with some discussion points. @abaguas please let me know what you think!

api/v1beta1/gslb_types.go
abaguas: Good suggestion. We can definitely do that and keep the CRD stable

controllers/refresolver/istiovirtualservice/istiovirtualservice.go
abaguas: I didn't realize there was a more recent version of the API 😅
Yes, supporting multiple versions would be necessary to support upgrades. We can do that.

In that case would you "duplicate" also the tests? It will be mostly copy paste and it will be make our pipelines slower. But it is probably the only way we have to guarantee both versions work.

Could we

controllers/refresolver/ingress/testdata/ingress_multiple_servers.yaml
```diff
@@ -43,4 +43,3 @@ status:
   loadBalancer:
     ingress:
       - ip: 10.0.0.1
-      - ip: 10.0.0.2
```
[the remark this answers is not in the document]
abaguas: I created a `multiple_ips` file that covers the case where there are multiple load balancer IP addresses.

Multiple servers refers to multiple ingresses and backends: `h1.cloud.example.com` -> `s1` and `h2.cloud.example.com` -> `ss1/ss2`

controllers/refresolver/istiovirtualservice/istiovirtualservice.go
abaguas: The `v1` golang library is compatible with `v1beta1` and `v1` CRDs, at least for the fields that we are interested it.
Which makes sense, because the istio controller should be able to handle `apiVersion` upgrades. That means we don't have to have an implementation of the reference resolver for each `apiVersion`, we can rely on `v1` for both.

That might not be the case in the future though, at some point apiVersions will be deprecated. For that I created a workflow that replaces `networking.k8s.io/v1` with `networking.k8s.io/v1beta1` and is only triggered when the `istio` tag is added to the PR.

ytsarev (dismissed):
I've updated the default-full-local-setup exaples to use new resourceRef style in e825c56247ed1e41acc4795fe47bcfeb756c263b

Also resolved the conflict in GHA workflow to be able to merge with `master`

```
k -n test-gslb get gslb failover-ingress -o yaml
kind: Gslb
metadata:
  annotations:
      {"apiVersion":"k8gb.absa.oss/v1beta1","kind":"Gslb","metadata":{"annotations":{},"name":"failover-ingress","namespace":"test-gslb"},"spec":{"resourceRef":{"apiVersion":"networking.k8s.io/v1","kind":"Ingress","matchLabels":{"app":"test-gslb-failover"}},"strategy":{"primaryGeoTag":"eu","type":"failover"}}}
  creationTimestamp: "2024-09-15T21:48:05Z"
  finalizers:
  - k8gb.absa.oss/finalizer
  generation: 2
  name: failover-ingress
  namespace: test-gslb
  resourceVersion: "1038"
  uid: abf12c92-0490-453d-bf1f-75194962a0aa
spec:
  ingress: {}
  resourceRef:
    kind: Ingress
    matchLabels:
      app: test-gslb-failover
  strategy:
    dnsTtlSeconds: 30
    primaryGeoTag: eu
    type: failover
status:
  geoTag: us
  healthyRecords:
    failover.cloud.example.com:
    - 172.20.0.6
    - 172.20.0.7
  hosts: failover.cloud.example.com
  loadBalancer:
    exposedIps:
    - 172.20.0.10
    - 172.20.0.11
  servers:
  - host: failover.cloud.example.com
    services:
    - name: frontend-podinfo
      namespace: test-gslb
  serviceHealth:
    failover.cloud.example.com: Healthy
```

Local test with new style resourceRef looks great.

I've noticed that if we reference non-existing apiGroup or Kind , the Gslb will be silent and will not give the user any hints that it can't resolve resource and how to fix the situation

```
cat test-broken.yaml
kind: Gslb
metadata:
  name: roundrobin-ingress-broken
  namespace: test-gslb
spec:
  resourceRef:
    kind: Degress
    matchLabels:
      app: roundrobin-test-gslb
  strategy:
    type: roundRobin # Use a round robin load balancing strategy, when deciding which downstream clusters to route clients to
    splitBrainThresholdSeconds: 300 # Threshold after which external cluster is filtered out from delegated zone when it doesn't look alive
    dnsTtlSeconds: 30 # TTL value for automatically created DNS records
```

The Event stream will be empty
```
k describe -f test-broken.yaml
Name:         roundrobin-ingress-broken
Namespace:    test-gslb
Labels:       <none>
Annotations:  <none>
API Version:  k8gb.absa.oss/v1beta1
Kind:         Gslb
Metadata:
  Creation Timestamp:  2024-09-15T22:04:22Z
  Finalizers:
    k8gb.absa.oss/finalizer
  Generation:  2
  Managed Fields:
    API Version:  k8gb.absa.oss/v1beta1
    Fields Type:  FieldsV1
    fieldsV1:
      f:metadata:
        f:finalizers:
          .:
          v:"k8gb.absa.oss/finalizer":
      f:spec:
        f:ingress:
    Manager:      k8gb
    Operation:    Update
    Time:         2024-09-15T22:04:22Z
    API Version:  k8gb.absa.oss/v1beta1
    Fields Type:  FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          .:
          f:kubectl.kubernetes.io/last-applied-configuration:
      f:spec:
        .:
        f:resourceRef:
        f:strategy:
          .:
          f:dnsTtlSeconds:
          f:splitBrainThresholdSeconds:
          f:type:
    Manager:         kubectl-client-side-apply
    Operation:       Update
    Time:            2024-09-15T22:04:22Z
  Resource Version:  2147
  UID:               e14fd3fb-0374-4f92-9fcf-73b1641e6d1a
Spec:
  Ingress:
  Resource Ref:
    API Version:  networking.k8s.io/v1
    Kind:         Degress
    Match Labels:
      App:  roundrobin-test-gslb
  Strategy:
    Dns Ttl Seconds:                30
    Split Brain Threshold Seconds:  300
    Type:                           roundRobin
Events:                             <none>
```

I suggest to create separate issue to track the verbosity enhancement and not bloat this PR

Approved, @abaguas thanks for the incredible work 🚀

ytsarev (approved):

PR #1674, opened 2024-07-30
Fix reconciliation on `Endpoints`'s subset changes

abaguas:
The idea was to still trigger reconciliation cycles if resources changes by checking if there was a `generation` update. https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#metadata
> generation: a sequence number representing a specific generation of the desired state. Set by the system and monotonically increasing, per-resource. May be compared, such as for RAW and WAW consistency.

However, the change caused a degradation watching `Endpoints` resources. These resources don't have a `generation` because they don't have a `status`. As a result, a reconciliation was only triggered on creation and deletion, but not on subset update. This caused the controller to become too slow reacting to `Endpoints` changes, thus leading to slow failovers and failing e2e tests if the GSLB reconcile timer didn't fall during the expected window.

The solution proposed in this PR is to always trigger the `endpointMapHandler` function when an `Endpoint` resource changes. This is the same behavior as before #1652.

---
In addition, the bind image `internetsystemsconsortium/bind9:9.16` pulled from the `edgedns` cluster was no longer available, so it is updated to version `9.18`:
```
➜  ~ kubectl get pods --context k3d-edgedns
NAME                    READY   STATUS             RESTARTS   AGE
edge-5df69889cf-8vths   0/1     ImagePullBackOff   0          4m37s
```


terratest/test/k8gb_weight_test.go
```diff
@@ -27,8 +27,6 @@ import (
 )
 
 func TestWeightsExistsInLocalDNSEndpoint(t *testing.T) {
-	t.Parallel()
```
abaguas, line 30: There were some random failures in this test, I had to remove parallelism to have it stable. One cluster was taking too long discovering the DNS entries of the other cluster, but I am not sure why.
Before:
```
   --- FAIL: TestWeightsExistsInLocalDNSEndpoint (337.85s)
      --- FAIL: TestWeightsExistsInLocalDNSEndpoint/with_gslb_only (245.38s)
      --- PASS: TestWeightsExistsInLocalDNSEndpoint/with_ref (92.48s)
```

After:
```
   --- PASS: TestWeightsExistsInLocalDNSEndpoint (195.55s)
      --- PASS: TestWeightsExistsInLocalDNSEndpoint/with_gslb_only (128.23s)
      --- PASS: TestWeightsExistsInLocalDNSEndpoint/with_ref (67.32s)
```

kuritka (approved):
Hi, thanks, I'm glad you found the issue. I'm a bit worried that in the future maybe the tests will crash again at a different point and parallelism will be the cause.  Anyway, thanks and happy to see  the correction.

abaguas:
> Hi, thanks, I'm glad you found the issue. I'm a bit worried that in the future maybe the tests will crash again at a different point and parallelism will be the cause. Anyway, thanks and happy to see the correction.

Hopefully when we try to revamp the tests we can migrate them one by one to the new framework and can get to the bottom of these parallelism issues

ytsarev (approved):
Great catch

PR #1677, opened 2024-07-30
fix: externaldns.interval not being respect

### LINKED ISSUE #1676: externaldns.interval not being respect
We have noticed an instance of incorrect helming in one of your Kubernetes configuration files. Hardcoded or default configuration values in the template are considered anti-pattern and configurable values in the values.yaml file are not being respect, resulting in user-provided configurations never being applied. Additionally, we provide anecdotal evidence from [trivy-operator#729](https://github.com/aquasecurity/trivy-operator/issues/729) regarding the incorrect helming defect.

[externaldns.interval](https://github.com/k8gb-io/k8gb/blob/bd9f599ae2e1aa57e97ac721836e024c54a25b44/chart/k8gb/values.yaml#L72) does not take affect when change it to other values. This is because missing implementation for this value in external-dns/external-dns.yaml file:

https://github.com/k8gb-io/k8gb/blob/bd9f599ae2e1aa57e97ac721836e024c54a25b44/chart/k8gb/templates/external-dns/external-dns.yaml#L23

## Expected Behavior
Change externaldns.interval to other value can take effect

## Actual Behavior
Only default value '1m' take effect

zyue110026:
fix #1676 

respect externaldns.interval

### DIFF
--- a/chart/k8gb/templates/external-dns/external-dns.yaml
+++ b/chart/k8gb/templates/external-dns/external-dns.yaml
@@ -28,7 +28,7 @@ spec:
         - --managed-record-types=A
         - --managed-record-types=CNAME
         - --managed-record-types=NS
-        - --interval={{ .Values.externaldns.interval | default "1m" }}
+        - --interval={{ .Values.externaldns.interval }}
         - --annotation-filter=k8gb.absa.oss/dnstype=extdns # filter out only relevant DNSEntrypoints
         - --txt-owner-id={{ include "k8gb.extdnsOwnerID" . }}
         - --txt-prefix=k8gb-{{ .Values.k8gb.clusterGeoTag }}- # add custom prefix to TXT records, critical for Cloudflare NS record creation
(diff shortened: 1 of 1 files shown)


ytsarev:
@zyue110026 thanks a lot for the catch! Could you please rebase the PR with the `master` branch?  we had a regression in our test pipelines that was fixed after the merge of https://github.com/k8gb-io/k8gb/pull/1674

ytsarev:
Made short investigation, this regression happened a while ago in https://github.com/k8gb-io/k8gb/pull/292/files#diff-c5b1eb253f3b95c48e2f9203764d56752b157636d69930e79e5e7e27a59567d0L76 . Thanks again for bringing it up 👍

zyue110026:
@ytsarev I rebased the pr with the master branch.

ytsarev (approved):
Thanks a lot for the catch and the contribution!

PR #1684, opened 2024-08-02
Fix flaky e2e test `TestFailoverPlayground/*stop_podinfo_on_eu_cluster`

abaguas:
The [FailoverPlaygroundTest](https://github.com/k8gb-io/k8gb/blob/master/terratest/test/k8gb_failover_playground_test.go#L110) where `podinfo` is stopped on one of the clusters has been flaky. Example [run 1](https://github.com/k8gb-io/k8gb/actions/runs/10212107371/job/28254769667#step:10:9968) and [run2](https://github.com/k8gb-io/k8gb/actions/runs/10204827942/job/28234222500#step:10:9857).

The test deploys a GSLB with failover strategy on two clusters. Then, the app is stopped on the EU cluster and the test sometimes fails because the following line returns an error: `err = instanceEU.WaitForAppIsStopped()`. This function calls a chain of functions that ends up on the [WaitForApp function](https://github.com/k8gb-io/k8gb/blob/master/terratest/utils/extensions.go#L443) expecting the app to have 0 replicas and 0 `DNSEndpoint` targets. In some situations this is the case. However, since the app is running on the US cluster, in other situations the failover happens very quickly (good feature of K8GB) and the US targets are already part of the DNSEndpoints targets: [example 1](https://github.com/k8gb-io/k8gb/actions/runs/10212107371/job/28254769667#step:10:8462), [example 2](https://github.com/k8gb-io/k8gb/actions/runs/10204827942/job/28234222500#step:10:8351).
The `DNSEndpoint` taken from the links above looks as follows, where the targets are the IP addresses of the US cluster:
```
{
     "apiVersion": "externaldns.k8s.io/v1alpha1",
     "kind": "DNSEndpoint",
     ...
     "spec": {
         "endpoints": [
             {
                 "dnsName": "playground-failover.cloud.example.com",
                 "labels": {
                     "strategy": "failover"
                 },
                 "recordTTL": 5,
                 "recordType": "A",
                 "targets": [
                     "172.18.0.7",
                     "172.18.0.8"
                 ]
             }
         ]
     }
 }
```
This is an expected output, and should be accepted. To better understand the proposed solution here is how the `DNSEndpoint` resource looked like before the failover:
```
 {
     "apiVersion": "externaldns.k8s.io/v1alpha1",
     "kind": "DNSEndpoint",
     ...
     "spec": {
         "endpoints": [
             {
                 "dnsName": "localtargets-playground-failover.cloud.example.com",
                 "recordTTL": 5,
                 "recordType": "A",
                 "targets": [
                     "172.18.0.4",
                     "172.18.0.5"
                 ]
             },
             {
                 "dnsName": "playground-failover.cloud.example.com",
                 "labels": {
                     "strategy": "failover"
                 },
                 "recordTTL": 5,
                 "recordType": "A",
                 "targets": [
                     "172.18.0.4",
                     "172.18.0.5"
                 ]
             }
         ]
     }
 }
```
Above we can see that in addition to the main `playground-failover.cloud.example.com` domain there is also a `localtargets-playground-failover.cloud.example.com`. This `localtargets-*` domain disappears once the app is stopped, which indicates that the controller learnt that the app was scaled to 0 replicas.
The proposed fix is therefore to check for the targets of the `localtargets-*` domain. This shows that the K8GB controller behaved as expected and does not depend on the synchronization of records between clusters.

### DIFF
--- a/terratest/utils/extensions.go
+++ b/terratest/utils/extensions.go
@@ -423,12 +423,12 @@ func (i *Instance) waitForApp(predicate func(instances int) bool, stop bool) (er
 	i.w.t.Logf("Wait for ExternalDNSEndpoint %s.%s to be filled by targets %s", i.w.state.gslb.name, i.w.namespace, i.w.state.gslb.host)
 	// second conditions
 	for n := 0; n < maxRetries/2; n++ {
-		ep, err := i.Resources().GetExternalDNSEndpointByName(i.w.state.gslb.name, i.w.namespace).GetEndpointByName(i.w.state.gslb.host)
+		ep, err := i.Resources().GetExternalDNSEndpointByName(i.w.state.gslb.name, i.w.namespace).GetEndpointByName(fmt.Sprintf("localtargets-%s", i.w.state.gslb.host))
 		if err != nil {
 			// app is already stopped and cant be found
 			if stop && err.Error() == notFoundError {
 				i.w.t.Logf("App is stopped %s", i.w.state.testApp.name)
-				return nil
+				break
 			}
 			i.w.t.Logf("Error waiting for the app %s. %s", i.w.state.testApp.name, err)
 			require.NoError(i.w.t, err)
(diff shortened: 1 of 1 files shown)


terratest/utils/extensions.go
```diff
@@ -449,7 +449,7 @@ func (i *Instance) waitForApp(predicate func(instances int) bool, stop bool) (er
 	i.w.t.Logf("Wait for coreDNS to be filled by local targets %s", i.w.state.gslb.host)
 	for n := 0; n < maxRetries/2; n++ {
 		localTargets := i.GetLocalTargets()
-		if len(localTargets) == 0 {
+		if (!stop && len(localTargets) == 0) || (stop && len(localTargets) != 0) {
```
abaguas, line 452: This is a follow up of https://github.com/k8gb-io/k8gb/pull/1682 since the number of `localTargets` should be 0 if the action is stop. It helps keeping the code more understandable but it didn't cause any issues since in that case the code would return already on line 431.

ytsarev (approved):
Another great find 👍 Thank you so much!

Issue #1698, opened 2024-08-11
Decommission embedded ingress


abaguas:
Since #1557 (released on v0.13.0) k8gb can reference ingress resources via labels. This functionality provides the same features as embedded ingress configuration and is easier to maintain since the GSLB CRD does not depend on external CRDs. It also makes it easier to adopt k8gb since no ingress changes are required.

By decommissioning embedded ingresses the code base will be reduced, as well as the runtime of the tests, without any loss of functionality.

Plan of action:
* Update documentation to use referenced instead of embedded ingress resources
* Update k8gb-lite (annotation based) capability to create a GSLB with a referenced ingress instead of a GSLB with an embedded ingress
* Remove embedded ingress from the CRD and codebase

PR #1700, opened 2024-08-11
Create referenced ingress if GSLB is configured with Ingress annotations

abaguas:
Tackles one of the prerequisites outlined in https://github.com/k8gb-io/k8gb/issues/1698 to decommission embedded ingresses.

By default a GSLB with a referenced Ingress will be created if a user adds `k8gb` annotations to an Ingress resource.
To smoothen the user migration:
* a warning is shown if an Ingress resource with k8gb annotations does not have any labels
* it is still possible to create a GSLB with an embedded ingress by toggling the helm value `annotationCreatesGSLBWithEmbeddedIngress`

In v1.1 the option will be deleted and the code will be cleaned up to support only Referenced Ingresses.


ytsarev:
thanks a lot for the contribution. As we are moving away from the ingress annotation, I believe we can close this one.

PR #1704, opened 2024-08-12
Allow adding custom annotations and labels to the k8gb controller

abaguas:
I faced a usecase where I needed to add an istio exception to the k8gb controller via an annotation. However, the chart did not allow it.

This PR also allows adding custom labels since this feature may be desired for other usecases.

### Testing
Modify values.yaml by adding/removing labels/annotations and run:
```
helm package . && helm template k8gb k8gb-v0.13.0.tgz > manifests.yaml
```


ytsarev (approved):
LGTM

PR #1709, opened 2024-08-16
Allow configuration of extra coredns server blocks

abaguas:
The goal of this PR is to allow configuration of the glue records of the other clusters in k8gb's coreDNS. I know this defies DNS best practices because k8gb is not authoritative for the parent zone. However, it allows k8gb to do cluster discovery without depending on any infrastructure outside of the cluster.

The configuration is quite generic, giving the users freedom to configure what they desire, therefore we could add it to the chart.

Example configuration:
```
    extraServerBlocks: |
      example.com:5353 {
        errors
        health
        ready
        prometheus 0.0.0.0:9153
        forward . /etc/resolv.conf
        k8s_crd {
            filter k8gb.absa.oss/dnstype=glue
            negttl {{ .Values.k8gb.dnsZoneNegTTL }}
            loadbalance weight
        }
      }
```

Testing:
```
helm package . && helm template k8gb k8gb-v0.13.0.tgz > manifests.yaml
```

### DIFF
--- a/chart/k8gb/README.md
+++ b/chart/k8gb/README.md
@@ -87,7 +87,8 @@ For Kubernetes `< 1.19` use this chart and k8gb in version `0.8.8` or lower.
 | infoblox.wapiPort | int | `443` | WAPI port |
 | infoblox.wapiVersion | string | `"2.3.1"` | WAPI version |
 | k8gb.clusterGeoTag | string | `"eu"` | used for places where we need to distinguish between different Gslb instances |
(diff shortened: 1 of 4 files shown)


ytsarev (commented):
> it allows k8gb to do cluster discovery without depending on any infrastructure outside of the cluster.

Can we document this use case?

ytsarev (approved):
This is useful for flexible configuration. Thanks for the contribution!

abaguas:
Thanks. I will revise my usecase and document it if I still find the setup useful

PR #1718, opened 2024-09-06
Update coredns helm chart repo url

abaguas:
The coredns helm chart fork was moved from https://github.com/AbsaOSS/coredns-helm/ to https://github.com/k8gb-io/coredns-helm


ytsarev (approved):
Thanks!

PR #1743, opened 2024-10-01
use upstream coredns chart instead of fork

### LINKED ISSUE #1741: CoreDNS AWS NLB health check not getting healthy
**Describe the issue**
When k8gb is deployed on AWS, the k8gb-coredns service creates an AWS Network Load Balancer with TCP health check, but no TCP port is available, so the target group never gets healthy. 

Generated service example:
```YAML
apiVersion: v1
kind: Service
metadata:
  annotations:
    ...
    service.beta.kubernetes.io/aws-load-balancer-type: nlb
  labels:
    ...
  name: k8gb-coredns
  namespace: k8gb
spec:
  ...
  ports:
  - name: udp-5353
    nodePort: 31412
    port: 53
    protocol: UDP
    targetPort: 5353
  selector:
    app.kubernetes.io/instance: k8gb
    app.kubernetes.io/name: coredns
  type: LoadBalancer
```

**To Reproduce**
Deploy k8gb following: https://www.k8gb.io/docs/deploy_route53.html

**Expected behavior**
The k8gb-coredns should expose a TCP port along with the UDP, so the target group gets healthy, as UDP health checks are not supported by AWS NLB.

**Workaround**
In case you are facing the same issue, you can overcome the error if you add the following annotation to your values.yaml within coredns:

`service.beta.kubernetes.io/aws-load-balancer-healthcheck-port: "<nginx_service_port>"`

The port must be filled with the Nginx service nodePort. As Nginx has TCP ports exposed, the health check will work. 

**Screenshots**
image

**Additional context**
Me and @ytsarev tried to check if it was possible to add extra ports through CoreDNS helm chart, but the service ports are imported using a helper, and not .Values.xx. References:

- https://github.com/k8gb-io/coredns-helm/blob/master/stable/coredns/templates/service.yaml#L43-L45
- https://github.com/k8gb-io/coredns-helm/blob/master/stable/coredns/templates/_helpers.tpl#L60-L63

abaguas:
CoreDNS is a core component of the K8GB application. Until now we were installing it using a fork of the official helm chart with the following [diff](https://github.com/coredns/helm/compare/master...k8gb-io:coredns-helm:master). The fork was necessary because Kubernetes did not support services of type load balancer with both udp and tcp ports, and to reduce the attack surface by running K8GB on a non-privileged port.

Since [version 1.26](https://kubernetes.io/docs/concepts/services-networking/service/#load-balancers-with-mixed-protocol-types) Kubernetes supports load balancers with mixed ports.
The coredns helm chart also evolved over the years, and with [this contribution](https://github.com/coredns/helm/pull/175) we are able to use coredns without opening privileged ports.

### Generated configuration

The generated configuration by the base chart has only the following differences:
* Ports exposed by the coredns service (fixes #1741)
```
<   - {port: 53, targetPort: 5353, protocol: UDP, name: udp-5353}
---
>   - {"name":"udp-5353","port":53,"protocol":"UDP","targetPort":5353}
>   - {"name":"tcp-5353","port":53,"protocol":"TCP","targetPort":5353}
```
* Ports exposed by the coredns container (allows prometheus metrics scraping)
```
<         - {containerPort: 5353, protocol: UDP, name: udp-5353}
<         - {containerPort: 5353, protocol: TCP, name: tcp-5353}
---
>         - {"containerPort":5353,"name":"udp-5353","protocol":"UDP"}
>         - {"containerPort":5353,"name":"tcp-5353","protocol":"TCP"}
>         - {"containerPort":9153,"name":"tcp-9153","protocol":"TCP"}
```
* coredns container security context (to disable the `NET_BIND_SERVICE` capability)
```
>         securityContext:
>           capabilities:
>             add: []
```

## Abusing the `servers` block

Since the service ports and the prometheus port configuration are taken from the `servers` block of the configuration we need to configure it as follows:
```
  servers:
  - port: 5353
    servicePort: 53
    plugins:
    - name: prometheus
      parameters: 0.0.0.0:9153
```
This opens the ports that we need on the coredns pod and service, nothing more.
In the coredns chart this configuration would be used to create the configmap that configures the served zones. But we are creating that config ourselves (by setting `coredns.deployment.skipConfig=true`) to reuse helm values.

### User configuration
The are no changes to the user configuration

### Conclusion

With this change we are now able to use the upstream chart instead of a fork. Unfortunately there is a overhead on the configuration necessary to the user, but we gain a lot. We can profit from all the features the coredns community offers, expose TCP port 53 publicly (fixes #1741) and expose the metrics port on the container which allows Prometheus metrics scraping.

---

### k3d services

By enabling k3d's loadbalancer we enable routing to Kubernetes services. This means we can use services of type LoadBalancer or ClusterIP. We no longer need to rely on the service type NodePort, or on making sure a pod is running on the node exposing the port. CoreDNS can be queried exactly the same way as it was before:
```
➜  ~ dig -p 5054 @localhost localtargets-roundrobin.cloud.example.com +short
172.19.0.11
172.19.0.12
➜  ~ dig -p 5054 @localhost localtargets-roundrobin.cloud.example.com +tcp +short
172.19.0.11
172.19.0.12
```
In a follow up PR I will try to extend this setup to expose the istio ingress gateway and the nginx ingress controller.

### upgrade testing

`kubectl -n k8gb delete svc k8gb-coredns --ignore-not-found`

### creating namespaces with kubectl apply

While reading the upgrade testing logs I saw errors for the creation of namespaces with `kubectl create` because the namespaces already existed. By using kubectl apply these errors are no longer thrown.

### istio version pinning

While updating the Makefile I noticed the istio version of the ingress controller was not pinned, so I added it.


k0da:
@abaguas This is cool. It just an upstream chart never worked with ServiceType=LoadBalancer. It has a conflict due to [this](https://kubernetes.io/docs/concepts/services-networking/service/#load-balancers-with-mixed-protocol-types) wasn't there. I never tested this FeatureGate

abaguas:
> @abaguas This is cool. It just an upstream chart never worked with ServiceType=LoadBalancer. It has a conflict due to [this](https://kubernetes.io/docs/concepts/services-networking/service/#load-balancers-with-mixed-protocol-types) wasn't there. I never tested this FeatureGate

Thank you @k0da! I just tested it in my cluster and I can have a LoadBalancer service with both UDP and TCP

abaguas:
This PR was getting slightly out of control with many little changes. I decided to split it in two PRs, one for the chart change and another one for the e2e testing changes.

abaguas:
The final part of this PR was merged in https://github.com/k8gb-io/k8gb/pull/1828.
This draft can now be closed

PR #1753, opened 2024-10-15
Add Open Systems to the adopters list

abaguas:
After the introduction of Istio support we are happy users of K8GB :)


ytsarev (approved):
@abaguas amazing news, thanks a lot! 🚀

Issue #1660, opened 2024-07-20
PoC: chainsaw as alternative to terratest


ytsarev:
**Problem statement**

[Terratest](https://terratest.gruntwork.io/) is great and we have very powerful e2e test suite at https://github.com/k8gb-io/k8gb/tree/master/terratest

Unfortunately we face multiple issues with imperative test logic, race conditions and false negatives in the test suite.

**Possible solution**

Create PoC of https://github.com/kyverno/chainsaw integration.

We can try to creat a declarative k8gb test suite with chainsaw 

Chainsaw is actively maintained by Kyverno and is successor of https://kuttl.dev/

Some high-level material on chainsaw: https://www.youtube.com/watch?v=hQJWGzogIiI

PR #1758, opened 2024-10-21
Introduce chainsaw as e2e testing framework

abaguas:
As stated in #1660, the terratest setup is quite complex. When developing new features it is often necessary to tweak the e2e testing setup, and at the moment it is quite time consuming to get it right. In addition, e2e tests are slow, they take about 25 minutes to run.

In this PR I experiment with Chainsaw Kyverno, the most popular testing framework for operators. I implemented the failover playground test as a proof of concept and I believe we can move forward with more tests. Let me know your thoughts.

### Analysis

Advantages:
* declarative testing allows developers to focus only on the expected outcomes, not the testing logic
* open-source framework frees up capacity since we no longer have to maintain a testing framework
* stepTemplate CRD allows reuse of steps across multiple tests
* tests run in parallel. We have a new opportunity to get parallelism right and reduce testing time
* contains all the features we need to run a test

Disadvantages:
* no loops. Running the same test for different ingress types will involve test duplication, but at least steps can be reused. Hopefully this feature will come soon.

### Test logic

The test's logic is as follows:
* Create the podinfo application
* Wait until initialization is complete: the DNSEndpoint is populated on both clusters
* Fetch the expected DNS targets from the DNSEndpoint resource and compare it with the dig's results
* Delete podinfo
* Fetch the expected DNS targets from the DNSEndpoint resource and compare it with the dig's results
* Recreate podinfo
* Fetch the expected DNS targets from the DNSEndpoint resource and compare it with the dig's results


ytsarev (changes requested):
This is great! I believe we can merge it and run terratest and chainsaw in parallel for a while. Just small step name bung in chainsaw github action to fix as suggested

abaguas:
> This is great! I believe we can merge it and run terratest and chainsaw in parallel for a while. Just small step name bung in chainsaw github action to fix as suggested

Yes, let's run them in parallel while we add chainsaw tests for all the scenarios we have terratest.
Then, if things are stable we can remove terratest.

ytsarev (approved):

PR #1769, opened 2024-10-31
Record event if GSLB references cannot be resolved

abaguas:
In the [PR where the istio ingress integration was introduced](https://github.com/k8gb-io/k8gb/pull/1672#pullrequestreview-2305623370), it was suggested to record events if GSLB references cannot be resolved. The controller was already generating logs, but an event would help the users debugging using `kubectl describe`. This is a new feature that was not yet present in K8GB. Therefore, to implement it we need to fetch an EventRecorder and grant the controller the appropriate permissions. The error messages were also adapted and currently look like:
```
[k3d-test-gslb2|test-gslb] ➜ k describe gslb roundrobin-ingress-broken
...
Events:
  Type     Reason          Age                From             Message
  ----     ------          ----               ----             -------
  Warning  ReconcileError  2s (x12 over 12s)  gslb-controller  error resolving references (APIVersion:networking.k8s.io/v1, Kind:Degress not supported)
```
```
[k3d-test-gslb2|test-gslb] ➜ k get events
LAST SEEN   TYPE      REASON           OBJECT                           MESSAGE
5s          Warning   ReconcileError   gslb/roundrobin-ingress-broken   error resolving references (APIVersion:networking.k8s.io/v1, Kind:Degress not supported)
```

The controller also logs a message (this was already the case before this PR):
```
2024-10-31T16:47:29Z ERR github.com/k8gb-io/k8gb/controllers/logging/logr.go:72 > events: Reconciler error {"Gslb":"test-gslb/roundrobin-ingress-broken","controller":"gslb","controllerGroup":"k8gb.absa.oss","controllerKind":"Gslb","name":"roundrobin-ingress-broken","namespace":"test-gslb","reconcileID":"2a6a4ad5-96f6-4ee6-8c04-28de0016fc57"} error="error resolving references (APIVersion:networking.k8s.io/v1, Kind:Degress not supported)"
```

There are many other situations where we can generate events, but they won't be included in this PR to avoid bloating it.

[Documentation about events using kubebuilder](https://book-v1.book.kubebuilder.io/beyond_basics/creating_events)

---

This PR also upgrades the golinter since the old version was not compatible with my local go version. Upon upgrading I also resolved the following warning:
```
WARN The linter 'exportloopref' is deprecated (since v1.60.2) due to: Since Go1.22 (loopvar) this linter is no longer relevant. Replaced by copyloopvar.
```


controllers/refresolver/ingress/ingress.go
```diff
@@ -52,7 +52,7 @@ func NewReferenceResolver(gslb *k8gbv1beta1.Gslb, k8sClient client.Client) (*Ref
 	}
 
 	if len(ingressList) != 1 {
-		return nil, fmt.Errorf("exactly one Ingress resource expected but %d were found", len(ingressList))
```
abaguas, line 55: This is unrelated, but since I was touching error messages I make it more consistent by using an integer both for expected and found ingresses

kuritka (approved):

Issue #1772, opened 2024-11-02
Create test setup for upstream DNS providers


abaguas:
During the community meeting on the 30.11.2024 we discussed how we can easily test K8GB's integrations with upstream DNS providers.
The K8GB maintainers don't have have in their company environment to all the supported integrations, therefore there is the need to introduce guides to perform such testing.
We discussed that connecting a local kubernetes cluster to the cloud infrastructure would be the most cost-efficient solution since the DNS infrastructure is often quite cheap.

nXtCyberNet:
hi @ytsarev i was seing this  issue and based on my research the aws route 53 was already completed for the integration setup , but the gcp and cloudflared was not setup till now , can i help you with that part if this issue is in your plan.

ytsarev:
@nXtCyberNet we need some form of automated e2e coverage for major/all integrations, definitely a thing to enhance

nXtCyberNet:
hi @ytsarev i have created the first e2e ci for route53 could you please review #2498 after you agree with my setup , i will create for gcp and cloudflare too

PR #1762, opened 2024-10-26
Switch extenral-dns back to official upstream

ytsarev:
Fixes #1744

This one needs extensive testing with various edge DNS environments. 

The initial local test with rfc2136 is successful.

..

<details>
  <summary>HOW TO RUN CI</summary>
---

  By default, all the checks will be run automatically. Furthermore, when changing website-related stuff, the preview will be generated by the netlify bot.

  ### Heavy tests
  Add the [`heavy-tests`](/k8gb-io/k8gb/issues?q=is%3A*+label%3Aheavy-tests) label on this PR if you want full-blown tests that include more than 2-cluster scenarios.

  ### Debug tests
  If the test suite is failing for you, you may want to try triggering `Re-run all jobs` (top right) with [debug logging](https://docs.github.com/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging) enabled. It will also make the [print debug](/k8gb-io/k8gb/blob/master/.github/actions/print-debug/action.yaml) action more verbose.

</details>

PR #1773, opened 2024-11-02
Create test setup for Azure DNS integration

abaguas:
As stated in #1772, we would like to have test setup with the main DNS providers. This PR creates a complete guide documenting how to test the Azure integration, including terraform and helm configuration.

According to the [docs](https://azure.microsoft.com/en-us/pricing/details/dns/) the infrastructure should cost $0.90 per month, but it can also be destroyed and re-provisioned anytime:
```
First 25 hosted DNS zones	$0.50 per zone per month
First billion DNS queries/month	$0.40 per million
```


abaguas:
If you have a suggestion of another folder where to place this content let me know

ytsarev:
@abaguas that's a great setup, thank you so much for your work on it. However, I am hesitant to merge any more terraform code to the repo after [BUSL](https://www.hashicorp.com/license-faq) changes. I tested the alternative setup with Crossplane please see https://github.com/ytsarev/k8gb/tree/azure-crossplane/dns-providers/azure/crossplane .
The rest of the guide was almost intact and I got to a successful external-dns test in https://github.com/k8gb-io/k8gb/pull/1762 

If you are ok with this route, I can create Crossplane-based abstraction(s) that will handle the test setup.

In the long run, I also plan to refactor to Crossplane and deprecate the old https://github.com/k8gb-io/k8gb/tree/master/docs/examples/route53/terraform-aws-eks-k8gb reference architecture to fully eliminate the Terraform footprint in this repository.

PR #1774, opened 2024-11-05
Add support for multiple zones

donovanmuller:
This PR adds the ability for K8gb to handle multiple edge and DNS zones.

We have a use case where we need to support multiple zones. There were some ideas on how to achieve this. Namely running multiple K8gb deployments in different namespaces, each with its own `edgeDNSZone`/`dnsZone`, but that ultimately turned out to be quite difficult to do given the Helm Chart, CRD and other issues. So we decided to add support for handling multiple `edgeZone`/`zone` pairs, taking inspiration from the `edgeDNSServers` implementation was done.

The Helm Chart has been updated to support the following values as an example of supporting two zones:

```yaml
...
k8gb:
...
  dnsZones:
    - edgeZone: "example.com"
      zone: "cloud.example.com"
      dnsZoneNegTTL: 300
    - edgeZone: "example.org"
      zone: "cloud.example.org"
      dnsZoneNegTTL: 300
...
```

<details>
---

  ### Heavy tests

  ### Debug tests

</details>


controllers/gslb_controller_reconciliation.go
```diff
@@ -176,7 +141,74 @@ func (r *GslbReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
+					Str("server", server.Host).
+					Str("zone", z.Zone).
+					Msg("Matched zone to server host")
+				zone = z
+				matchedServers = append(matchedServers, server)
+			}
+		}
+	}
+
+	if zone == (utils.DNSZone{}) {
```
abaguas, line 166: Is this logic to trigger a requeue if no server matched necessary?
Let's suppose the configured dnsZone is "example.com". There is a GSLB resource with a server that has the hostname "cloud.example.com". Then, the users swaps the hostname to "cloud.other.com". With this implementation the GSLB will never stop advertising "cloud.example.com".

controllers/gslb_controller_reconciliation.go
```diff
@@ -103,42 +104,6 @@ func (r *GslbReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		Str("namespace", gslb.Namespace).
 		Interface("strategy", gslb.Spec.Strategy).
 		Msg("Resolved strategy")
-	// == Finalizer business ==
```
abaguas, line 106: Why did you move the "Finalizer business" down?

Afaiu we try to keep it as up as possible to terminate the reconciliation immediately if there GSLB is going to be deleted

controllers/gslb_controller_reconciliation.go
```diff
@@ -190,7 +222,7 @@ func (r *GslbReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		Msg("Resolved LoadBalancer and Server configuration referenced by Ingress")
 
 	// == external-dns dnsendpoints CRs ==
-	dnsEndpoint, err := r.gslbDNSEndpoint(gslb)
+	dnsEndpoint, err := r.gslbDNSEndpoint(gslb, zone)
```
abaguas, line 225: A GSLB may have servers with different zones.
What about storing the zone in the status of the GSLB, perhaps by extending Adapting the `Server` datastructure?

abaguas (commented):
Thank you for the PR, this is indeed a nice feature.

Is the implementation complete?
External DNS would need an updated `domain-filter`:
```
- --domain-filter={{ .Values.k8gb.edgeDNSZone }}
```

Btw, shameless plug, if the e2e framework is slowing down your development then chainsaw would be a great alternative: https://github.com/k8gb-io/k8gb/pull/1758

donovanmuller:
> Is the implementation complete? External DNS would need an updated `domain-filter`:
> 
> ```
> - --domain-filter={{ .Values.k8gb.edgeDNSZone }}
> ```

Good catch! I wanted to raise the PR to get eyes on it, in case there were discussions around the overall design. I haven't had time to work through the integration tests.

I am following the Chainsaw work closely 👍

controllers/gslb_controller_reconciliation.go
donovanmuller: I will test this out 👍

controllers/gslb_controller_reconciliation.go
donovanmuller: Agreed, I didn't want to, but the finalizer needs to [pass](https://github.com/k8gb-io/k8gb/pull/1774/files#diff-adb1e6d23c5a4c850b9b760b27a3d557b341c5c438526f607068e4d1f9d00352R190) the `zone`, which is only available after the zone has been matched.

controllers/gslb_controller_reconciliation.go
donovanmuller: Good idea, will add.

controllers/gslb_controller_reconciliation.go
abaguas: I see. Actually, if we store the zone in the status of the GSLB then the finalizer can use the status of the last reconciliation, which is already present on the GSLB resource.

abaguas:
One more idea. Do we need to check in the K8GB controller to which zone a domain belongs to?
We are already configuring coredns, which will ignore `DNSEndpoint` resources for zones it cannot control.
The only thing we really need to do is to configure zone delegation for all zones.

This would dramatically simplify the implementation of this PR.

kuritka:
#1845

PR #1776, opened 2024-11-11
use upstream coreDNS chart instead of fork

### LINKED ISSUE #1741: CoreDNS AWS NLB health check not getting healthy

abaguas:
CoreDNS is a core component of the K8GB application. Until now we were installing it using a fork of the official helm chart with the following [diff](https://github.com/coredns/helm/compare/master...k8gb-io:coredns-helm:master). The fork was necessary because Kubernetes did not support services of type load balancer with both udp and tcp ports, and to reduce the attack surface by running K8GB on a non-privileged port.

Since [version 1.26](https://kubernetes.io/docs/concepts/services-networking/service/#load-balancers-with-mixed-protocol-types) Kubernetes supports load balancers with mixed ports.
The coredns helm chart also evolved over the years, and with [this contribution](https://github.com/coredns/helm/pull/175) we are able to use coredns without opening privileged ports.

### Generated configuration

The generated configuration by the base chart has only the following differences:
* Ports exposed by the coredns service (fixes #1741)
```
<   - {port: 53, targetPort: 5353, protocol: UDP, name: udp-5353}
---
>   - {"name":"udp-5353","port":53,"protocol":"UDP","targetPort":5353}
>   - {"name":"tcp-5353","port":53,"protocol":"TCP","targetPort":5353}
```
* Ports exposed by the coredns container (allows prometheus metrics scraping)
```
<         - {containerPort: 5353, protocol: UDP, name: udp-5353}
<         - {containerPort: 5353, protocol: TCP, name: tcp-5353}
---
>         - {"containerPort":5353,"name":"udp-5353","protocol":"UDP"}
>         - {"containerPort":5353,"name":"tcp-5353","protocol":"TCP"}
>         - {"containerPort":9153,"name":"tcp-9153","protocol":"TCP"}
```
* coredns container security context (to disable the `NET_BIND_SERVICE` capability)
```
>         securityContext:
>           capabilities:
>             add: []
```

### Abusing the `servers` block

Since the service ports and the prometheus port configuration are taken from the `servers` block of the configuration we need to configure it as follows:
```
  servers:
  - port: 5353
    servicePort: 53
    plugins:
    - name: prometheus
      parameters: 0.0.0.0:9153
```
This opens the ports that we need on the coredns pod and service, nothing more.
In the coredns chart this configuration would be used to create the configmap that configures the served zones. But we are creating that config ourselves (by setting `coredns.deployment.skipConfig=true`) to reuse helm values.

### User configuration
The are no changes to the user configuration

### Conclusion

With this change we are now able to use the upstream chart instead of a fork. We can profit from all the features the coredns community offers, expose TCP port 53 publicly (fixes #1741) and expose the metrics port on the container which allows Prometheus metrics scraping.

---

#### Creating namespaces in "Upgrade Testing"

In the "Upgrade Testing" pipeline we install k8gb twice, first the stable and then the test version. While reading the upgrade testing logs I noticed errors on the second namespace creation with `kubectl create`. By using `kubectl apply` these errors are no longer thrown if the namespace already exists.

#### Istio version pinning

While updating the Makefile I noticed the istio version of the ingress controller was not pinned, so I pinned it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,7 @@ LOG_LEVEL ?= debug
 CONTROLLER_GEN_VERSION  ?= v0.16.5
 GOLIC_VERSION  ?= v0.7.2
 GOLANGCI_VERSION ?= v1.60.3
+ISTIO_VERSION ?= v1.23.3
 POD_NAMESPACE ?= k8gb
 CLUSTER_GEO_TAG ?= eu
 EXT_GSLB_CLUSTERS_GEO_TAGS ?= us
@@ -179,7 +180,7 @@ deploy-local-cluster:
 	kubectl config use-context k3d-$(CLUSTER_NAME)$(CLUSTER_ID)
 
 	@echo -e "\n$(YELLOW)Create namespace $(NC)"
-	kubectl apply -f deploy/namespace.yaml
+	kubectl create namespace k8gb --dry-run=client -o yaml | kubectl apply -f -
 
 	@echo -e "\n$(YELLOW)Deploy GSLB operator from $(VERSION) $(NC)"
 	$(MAKE) deploy-k8gb-with-helm
@@ -191,18 +192,18 @@ deploy-local-cluster:
 		--version 4.0.15 -f $(NGINX_INGRESS_VALUES_PATH)
 
 	@echo -e "\n$(YELLOW)Install Istio CRDs $(NC)"
-	kubectl create namespace istio-system
+	kubectl create namespace istio-system --dry-run=client -o yaml | kubectl apply -f -
 	helm repo add --force-update istio https://istio-release.storage.googleapis.com/charts
 	helm repo update
-	helm upgrade -i istio-base istio/base -n istio-system
+	helm upgrade -i istio-base istio/base -n istio-system --version "$(ISTIO_VERSION)"
 
 	@echo -e "\n$(YELLOW)Install Istiod $(NC)"
-	helm upgrade -i istiod istio/istiod -n istio-system --wait
+	helm upgrade -i istiod istio/istiod -n istio-system --version "$(ISTIO_VERSION)" --wait
 
 	@echo -e "\n$(YELLOW)Install Istio Ingress Gateway $(NC)"
-	kubectl create namespace istio-ingress
+	kubectl create namespace istio-ingress --dry-run=client -o yaml | kubectl apply -f -
 	helm upgrade -i istio-ingressgateway istio/gateway -n istio-ingress \
-		-f $(ISTIO_INGRESS_VALUES_PATH)
+		--version "$(ISTIO_VERSION)" -f $(ISTIO_INGRESS_VALUES_PATH)
 
 	@if [ "$(DEPLOY_APPS)" = true ]; then $(MAKE) deploy-test-apps ; fi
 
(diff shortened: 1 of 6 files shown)


ytsarev (commented):
@abaguas did we test this change for the case of LoadBalancer service type exposure?
Important bits:
* https://github.com/k8gb-io/k8gb/blob/master/chart/k8gb/templates/deployment.yaml#L113-L116 special handling
* Example usage https://github.com/k8gb-io/k8gb/blob/master/docs/examples/route53/k8gb/k8gb-cluster-eu-west-1.yaml#L14

The upstream chart seems compatible  https://github.com/coredns/helm/blob/master/charts/coredns/values.yaml#L37
but I want to explicitly discuss this important code path

abaguas:
> @abaguas did we test this change for the case of LoadBalancer service type exposure? Important bits:
> 
> * https://github.com/k8gb-io/k8gb/blob/master/chart/k8gb/templates/deployment.yaml#L113-L116 special handling
> * Example usage https://github.com/k8gb-io/k8gb/blob/master/docs/examples/route53/k8gb/k8gb-cluster-eu-west-1.yaml#L14
> 
> The upstream chart seems compatible https://github.com/coredns/helm/blob/master/charts/coredns/values.yaml#L37 but I want to explicitly discuss this important code path

Yes yes, nothing changes there. The LoadBalancer service exposure is the same as it was before.
In https://github.com/k8gb-io/k8gb/pull/1743/files I changed k3d to expose coredns using a service type of service load balancer and it was successful. I then decided to split the changes into multiple PRs because it was getting quite hard to review. Once this is merged I will update the k3d setup, so that we can start testing automatically what is likely the most frequent method of exposing coredns.

ytsarev (approved):
Thanks for the clarification 👍

PR #1788, opened 2024-12-10
Add flag to support ClusterIP exposed CoreDNS

bewing:
Some environments (Calico bare-metal, etc) may allow direct client reachability to the Service CIDR, bypassing the need to assign and use LoadBalancerIPs.  A new config flag and environment variable allows selection/return of the ClusterIPs in the assistant rather than externally assigned LoadBalancer IPs.

### DIFF
--- a/controllers/providers/assistant/gslb.go
+++ b/controllers/providers/assistant/gslb.go
@@ -58,8 +58,7 @@ func NewGslbAssistant(client client.Client, k8gbNamespace string, edgeDNSServers
 	}
 }
 
-// CoreDNSExposedIPs retrieves list of IP's exposed by CoreDNS
-func (r *Gslb) CoreDNSExposedIPs() ([]string, error) {
+func (r *Gslb) GetCoreDNSService() (*corev1.Service, error) {
 	serviceList := &corev1.ServiceList{}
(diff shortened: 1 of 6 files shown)


controllers/providers/assistant/gslb.go
```diff
@@ -58,8 +58,7 @@ func NewGslbAssistant(client client.Client, k8gbNamespace string, edgeDNSServers
 	}
 }
 
-// CoreDNSExposedIPs retrieves list of IP's exposed by CoreDNS
-func (r *Gslb) CoreDNSExposedIPs() ([]string, error) {
+func (r *Gslb) GetCoreDNSService() (*corev1.Service, error) {
```
abaguas, line 61: I like this separation of responsibilities. One function to fetch the service and another one to fetch the IP address

abaguas (commented):
We can simplify the configuration to the users by not adding a new flag. Setting the `CoreDNSExposed` flag to true is enough.
A service only has one type so we dynamically check if the service is of type LoadBalancer or ClusterIP. Then we can extend logic of the `CoreDNSExposedIPs` function to fetch the IP addresses depending on the type.

abaguas (approved):

PR #1802, opened 2025-01-15
[maintenance] bump golangci-lint to latest

kuritka:
bumping golangci-lint to latest


controllers/providers/dns/infoblox.go
```diff
@@ -227,6 +227,7 @@ func (p *InfobloxProvider) saveHeartbeatTXTRecord(objMgr *ibcl.ObjectManager, gs
 		log.Info().
 			Str("HeartbeatTXTName", heartbeatTXTName).
 			Msg("Creating split brain TXT record")
+		//nolint:gosec
```
ytsarev, line 230: @kuritka, can we document the skipping of this linter? Without additional comment, it looks slightly insecure :D

kuritka: Hi @ytsarev , sure, amended

PR #1803, opened 2025-01-15
[maintenance] bump module dependencies 2025-01-15

kuritka:


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/k8gb-io/k8gb
 
-go 1.23
+go 1.23.3
 
 toolchain go1.23.4
```
ytsarev, line 5: @kuritka is it expected to have different patch versions between `go` and `toolchain`?

kuritka: nice catch @ytsarev, while toolchain is same as go version, I can remove toolchain completelly and make live easier.

ytsarev (approved):
Thanks a lot!

PR #1828, opened 2025-02-09
[draft] expose coredns as loadbalancer service in local setup

abaguas:


k0da:
@abaguas any blockers with this?

PR #1829, opened 2025-02-11
Remove SplitBrain implementation

kuritka:
- from CRD
- Terratests
- CR, Spec definition
- From Depresolver (including Heartbeat functionality etc.).
- HelmChart
- Remove from Infoblox provider

Removing SplitBrain functionality from GLSB spec, terratests, CRD

for reviewers: Basically the core of change is `infoblox.go` and `gslb.go`, `gslb_types.go` + helmChart. the rest are tests, depresolver, mocks, data.


k0da (dismissed):

ytsarev (approved):

ytsarev:
It was working code but isolated to in-tree Infoblox provider code only. Externaldns-based providers are not affected by this change 👍

kuritka:
@ytsarev , @k0da - please review one more time. In previous implementation was breaking change as I changed CRD, which causes potentially  issues on already existing GSLBs as they could have defined `splitBrainThresholdSeconds`.

```yaml
strategy:
    type: roundRobin
    dnsTtlSeconds: 30
```

ytsarev:
@kuritka good catch. We need to track the `splitBrainThresholdSeconds:` field deprecation in consequent releases then. Could you create an Issue?

ytsarev (approved):

k0da (approved):

PR #1849, opened 2025-03-03
Create test setup for AWS integration

abaguas:
Similar to the Azure setup done in #1773


abaguas:
Tackled in https://github.com/k8gb-io/k8gb/pull/1897

PR #1853, opened 2025-03-08
Ignore 'mesh' gateway in Istio Virtual Service

### LINKED ISSUE #1833: Ignore 'mesh' gateway when counting referenced gateways in VirtualService
### Description:
Currently, when counting the referenced gateways in the VirtualService, the gateway named "mesh" is included in the count, which can cause issues when trying to combine internal and external VirtualServices.

### Proposed Change:
- Ignore the "mesh" gateway when counting gateways in the VirtualService.
- Allow for the combined use of internal and external VirtualServices, as described in the Istio documentation.

The reserved word "mesh" is used to imply all the sidecars in the mesh. 
For more details: [Istio VirtualService Documentation](https://istio.io/latest/docs/reference/config/networking/virtual-service/#VirtualService-gateways)

### Example Code Change:
The following code in `k8gb/controllers/refresolver/istiovirtualservice/istiovirtualservice.go:105` is proposed to ignore the "mesh" gateway:

```go
var validGateways []string
for _, gateway := range virtualService.Spec.Gateways {
    if gateway != "mesh" {
        validGateways = append(validGateways, gateway)
    }
}

if len(validGateways) != 1 {
    return nil, fmt.Errorf("expected exactly 1 Gateway to be referenced by the VirtualService but %d were found", len(virtualService.Spec.Gateways))
}
gatewayRef := strings.Split(validGateways[0], "/")
```

### Example usage in Istio:
The "mesh" gateway can be combined with external gateways, allowing the same VirtualService rule to apply both internally and externally:

```yaml
kind: VirtualService
metadata:
  name: myapp
spec:
  hosts:
  - myapp.com
  - helloworld.default.svc.cluster.local
  gateways:
  - mesh
  - myapp-gateway
  http:
  - match:
    - uri:
        prefix: /hello
      gateways:
      - myapp-gateway
    route:
    - destination:
        host: helloworld.default.svc.cluster.local
        subset: v1
  - match:
    - gateways:
      - mesh
    route:
    - destination:
        host: helloworld.default.svc.cluster.local
        subset: v1

abaguas:
The 'mesh' gateway is not used for ingress, therefore it should be ignored by K8GB. This allows K8GB users to expose the virtual service to both via ingress and to the rest of the mesh.

Closes #1833

Thank you @jvlerner for the great description and solution of the issue!

### DIFF
--- a/controllers/refresolver/istiovirtualservice/istiovirtualservice.go
+++ b/controllers/refresolver/istiovirtualservice/istiovirtualservice.go
@@ -102,10 +102,18 @@ func getGslbVirtualServiceRef(gslb *k8gbv1beta1.Gslb, k8sClient client.Client) (
 
 // getGateway retrieves the istio gateway referenced by the istio virtual service
 func getGateway(virtualService *istio.VirtualService, k8sClient client.Client) (*istio.Gateway, error) {
-	if len(virtualService.Spec.Gateways) != 1 {
-		return nil, fmt.Errorf("expected exactly 1 Gateway to be referenced by the VirtualService but %d were found", len(virtualService.Spec.Gateways))
+	var ingressGateways []string
+	for _, gateway := range virtualService.Spec.Gateways {
+		// count only dedicated ingress gateways
+		if gateway != "mesh" {
+			ingressGateways = append(ingressGateways, gateway)
+		}
 	}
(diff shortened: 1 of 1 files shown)


ytsarev (approved):
Thanks a lot for the fix 👍

PR #1855, opened 2025-03-09
Improve external dns performance

abaguas:
External dns currently filters DNSEndpoint on the client side since it looks for the value of `k8gb.absa.oss/dnstype` in the annotations. Performance can be improved by using `--label-filter` since filtering will happen server-side: https://github.com/kubernetes-sigs/external-dns/blob/master/docs/faq.md?plain=1#L271-L274

This also paves the way to using the upstream external-dns chart.


kuritka (approved):

k0da (approved):
Makes sense. Since external-dns does cluster scope fetch. So DNSEndpoints fetchs are mixed application and delegation types.

PR #1860, opened 2025-03-12
DNSZones setup on bootstrap

kuritka:
## Motivation
If we talk about enterprise environments, it is easy to encounter a situation where we have thousands of GSLBs and a single DNS server. This server can be easily overwhelmed. Currently, each reconciliation (which occurs every 30 seconds by default) results in a read hit on the DNS server. With such a large number of GSLBs, this becomes a problem.

This PR moves the DNS zone configuration to the bootstrap process, eliminating the need to query the DNS on every reconciliation. Additionally, it fixes finalizers, preventing the recreation of a DNS zone each time a GSLB is deleted.

The solution works by introducing a second reconciliation process that only triggers on startup or when changes occur in the k8gb-coredns service or init-ingress.

If k8gb-coredns service is of type LoadBalancer, we simply read the IP addresses and create the DNS zones on the DNS server (including external-DNS endpoints in the k8gb namespace). If LoadBalancer is not used, the k8gb-coredns service must contain the annotation `k8gb.io/coredns-ingress-ref=<init-ingress-namespace/init-ingress-name>`, from which k8gb will retrieve the IP addresses. Lastly, the assistant has been refactored, and external DNS endpoints have been separated into an independent implementation (k8gbendpoint.DNSEndpoint interface, with implementations: **delegationZoneDNSEndpoint** and **applicationDNSEndpoint**). This improves code maintainability.

## DNSEndpoints refactored
- dnsendpoint refactored to standalone package. We have two implementations
   - **applicationDNSEndpoint**
   - **delegationZoneDNSEndpoint**
logic is same but implemented as strategies.

## Create Zone delegation
Creating a delegation zone during bootstrap - the necessary IP addresses are obtained from coreDNSService (if the type is LoadBalancer). If it is not this service must include an annotation (k8gb.io/coredns-ingress-ref) that contains the ingress, which includes these addresses. The format is <namespace>/ingress. Locally, we use init-ingress in the k8gb namespace.

- We added secondary reconciliation (CoreDNSReconciler) creates dns zones
    - on bootstrap
    - when k8gb-coredns service has changed
    - when the ingress specified in the annotation (k8gb.io/coredns-ingress-ref) in the k8gb-coredns service changes.
    - otherwise reconciliation waits.

## Chainsaw tests
- Added chainsaw test to check whether correct endpoint exists

## Finalizers
I removed finalizers.From the GSLB perspective, application DNS endpoints are automatically removed thanks to the owner reference. The previous implementation of the GSLB finalizer was incorrect. After removing the GSLB, the delegationZone DNS Endpoint was also removed—and was subsequently created by another GSLB during reconciliation.

Now, finalization is no longer necessary; however, the interface still accounts for finalizers and within implementation is ready for them.
- finalizators are disabled and fixed. BEFORE: Finalizer of GSLBE removed zone delegation which was recreated back by another GSLB

## Assistant refactored
I refactored assistant - now it returns only resource and IPs. We have implementations for k8gb-coredns and ingress mentioned above
- Create zone delegation refactored

## TODO:
Change affected plenty of unit tests, I will fix them in followup PR(s)
- Replace Unit tests

external-dns issue - no multizones without upgrade
- For multizones externalDNS `v0.12.0` or higher must be provided

Add one chainsaw test hitting multiple apps via multiple domains
- Chainsaw with multizones


controllers/gslb_controller_reconciliation.go
```diff
@@ -197,29 +154,37 @@ func (r *GslbReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
 	if err != nil {
-		log.Err(err).Msg("Unable to create zone delegation")
 		m.IncrementError(gslb)
-		return result.Requeue()
+		return result.RequeueError(err)
 	}
-	szd.End()
+	s.End()
+
+	// == handle delegated zone in Edge DNS
```
k0da, line 179: Please remove commented code

k0da (approved):

controllers/providers/k8gbendpoint/helper.go
```diff
@@ -0,0 +1,161 @@
...
+		},
+	}
+	// if hostToUse == "172.18.0.7" || hostToUse == "172.18.0.6" {
+	//	nameServerToUse[0].Host = "127.0.0.1"
+	//	nameServerToUse[0].Port = 5053
+	// }
+	// if hostToUse == "172.18.0.10" || hostToUse == "172.18.0.11" {
+	//	nameServerToUse[0].Host = "127.0.0.1"
+	//	nameServerToUse[0].Port = 5054
+	// }
```
ytsarev, line 142: some leftover from testing? should be removed?

controllers/gslb_controller_reconciliation.go
```diff
@@ -197,29 +154,27 @@ func (r *GslbReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 	gslb.Status.ServiceHealth = serviceHealth
 
 	// == external-dns dnsendpoints CRs ==
-	dnsEndpoint, err := r.gslbDNSEndpoint(gslb)
+	_, s := r.Tracer.Start(context.Background(), "gslbDNSEndpoint")
+	epProvider := k8gbendpoint.NewApplicationDNSEndpoint(context.TODO(), r.Client, r.Config, gslb, log, r.updateRuntimeStatus)
```
ytsarev, line 158: Is it intentional usage of TODO context? Something to consider https://dev.to/zakariachahboun/are-you-confused-about-contexttodo-vs-contextbackground-heres-what-you-need-to-know-2h46

controllers/gslb_controller_weight_test.go
```diff
@@ -35,161 +18,162 @@ limitations under the License.
...
+//			config := predefinedConfig
+//			settings := provideSettings(t, config)
+//			settings.reconciler.DNSProvider = m
+//			settings.reconciler.DepResolver = r
+//
+//			// act, assert
+//			_, _ = settings.reconciler.Reconcile(context.TODO(), settings.request)
+//		})
+//	}
+//}
```
ytsarev, line 179: do we still need this commented block?

abaguas (approved):
Really like the structure of the refactoring.
As you said, tests and minor fixes can be taken care of in a follow up PR

controllers/gslb_controller_reconciliation.go
kuritka: thanks @ytsarev , I’ll delete it in the following PR.

Just for info, here’s why it’s there: I deliberately insert this block of code so I can debug k8gb from my own IDE against local test clusters.

Example: k8gb is running on the EU cluster. The k8gb instance for the US cluster is running in the debugger, and I’m running Terratests in a second debugger instance. Very useful.

controllers/providers/k8gbendpoint/helper.go
kuritka: Hi, thanks, I’ll delete it in the next PR.

Just for info, here’s why it’s there: I deliberately insert this block of code so I can debug k8gb from my own IDE against local test clusters.

Example: k8gb is running on the EU cluster. The k8gb instance for the US cluster is running in the debugger, and I’m running Terratests in a second debugger instance.

I always insert it there during debugging.

controllers/gslb_controller_weight_test.go
kuritka: @ytsarev  I still need to refactor some unit tests (weight, factory), and I’ve already rewritten some from scratch (infoblox and externaldns – those are done, and I’ll prepare a PR for them ASAP). Let’s please keep it there for now — I want to refactor them later this week.

controllers/gslb_controller_reconciliation.go
kuritka: That’s a good question. I had previously thought about whether there should rather be a single context within one reconciler, so that in case of a hang, the loop would get cancelled.

This was meant to be just a context within a single trace call.

ytsarev (approved):
@kuritka thanks for the massive work and clarifications 👍

PR #1861, opened 2025-03-17
Fix status field requirements

k0da:
When migrating from 0.12 I've found that updates to any gslb object was failing due to validation errors.

This happens when resourceRef is not set yet, object is saved with .spec.resourceRef but status.loadbalancer nor status.servers is not yet set or calculated.

Lets relax a requirement to allow smoother transition to updated status structure

..

<details>
---

  ### Heavy tests

  ### Debug tests

</details>


ytsarev (approved):
Thanks, LGTM 👍

kuritka (approved):

PR #1864, opened 2025-03-24
Bootstrap in standalone package

kuritka:
In the current implementation, we retrieve IP addresses during bootstrap—before the Client instance is created. This brings one issue—caching. `client.Client` works with an internal cache, and when I performed a read outside of that cache, internal warnings occurred. This PR ensures that we load the IP addresses before the controller manager, safely outside the controller, as one of the very first steps during startup. Then, when the coreDNS Service or ingress changes, we can retrieve the IP address using the cached client. 

Bootstrap logic lives in standalone package now.

Already tested Chainsaw + unittests + live systems


abaguas (commented):
I understand what you are trying to do: watching a service or an ingress depending on how coredns is exposed. However, this looks quite complicated.

The coredns helm chart has the following value: [coredns.serviceType](https://github.com/coredns/helm/blob/master/charts/coredns/values.yaml#L37C1-L37C12). Could we pass it as an environment variable to the k8gb controller and use it for our logic?
- if the type is load balancer then watch the service with label `kubernetes.io/name: "CoreDNS"`
- if the type is ClusterIP then watch the ingress with label `k8gb.io/coredns=true` (or another label of your choice)

Here I am also suggesting to label the ingress directly, since the we can look it up natively using watchesOptions as the example bellow shows:
```
coreDNSPredicate := predicate.Funcs{
	CreateFunc: func(e event.CreateEvent) bool {
		return e.Object.GetLabels()["k8gb.io/coredns"] == "true"
	},
	UpdateFunc: func(e event.UpdateEvent) bool {
		return e.ObjectNew.GetLabels()["k8gb.io/coredns"] == "true"
	},
	DeleteFunc: func(e event.DeleteEvent) bool {
		return e.Object.GetLabels()["k8gb.io/coredns"] == "true"
	},
}

return ctrl.NewControllerManagedBy(mgr).
	Named(r.Bootstrap.CoreDNSService.Name).
	Watches(
		&source.Kind{Type: &corev1.Service{}},
		coreDNSWatcher,
		builder.WithPredicates(coreDNSPredicate), // Apply the label filter here
	).
	Watches(&source.Kind{Type: &netv1.Ingress{}}, ingressWatcher).
	Complete(r)
``` 

What do you think?

abaguas:
> Hi @abaguas , thx for mention. Basically, it's just about determining the type of CoreDNS Service—either dynamically or from the configuration by adding an extra environment variable. I’ll definitely consider it; it could be refactored. Externally, the behavior will remain the same, so there's no risk of a breaking change.
> 
> 

Not entirely. In your setup are you able to label the ingress resource instead of annotating the coredns service?
That is necessary for the code simplification proposed above

abaguas:
> > Hi @abaguas , thx for mention. Basically, it's just about determining the type of CoreDNS Service—either dynamically or from the configuration by adding an extra environment variable. I’ll definitely consider it; it could be refactored. Externally, the behavior will remain the same, so there's no risk of a breaking change.
> 
> Not entirely. In your setup are you able to label the ingress resource instead of annotating the coredns service? That is necessary for the code simplification proposed above

@kuritka? If that is the case I can propose a PR

kuritka:
@abaguas  Alright, let’s take it that way

Issue #1863, opened 2025-03-18
Infoblox provider creates glueA with ttl 1 hour


k0da:
infoblox client v1 creates DelegatedZones with no TTL on glueA record which results into default ttl 3600:
```
k8gb-677b5f8658-7pg82 k8gb 2025-03-18T16:49:50Z INF github.com/k8gb-io/k8gb/controllers/providers/assistant/gslb.go:260 > Resolved glue A record for NS edgeDNSServers=[{"Host":"1.1.1.1","Port":53}] glueARecord=[{"A":"2.2.2.2","Hdr":{"Class":1,"Name":"gslb-ns-270-nonprod-example.com.","Rdlength":4,"Rrtype":1,"Ttl":3595}}] nameserver=gslb-ns-270-nonprod.example.com
```

infoblox-client-go v2 supports setting TTL

ytsarev:
Fixed by https://github.com/k8gb-io/k8gb/pull/1866

PR #1866, opened 2025-03-24
Infoblox v2

kuritka:
# Infoblox v2

Thanks to the implementation of Infoblox v2 and multizone support, we should now be able to configure TTL values for individual domains (TTL per multizone is still pending completion, but the implementation will not be complex — everything is already prepared). Infoblox v2 also replaces objectmanager with IObjectmanager, which allows us to easily mock this object and create better unit tests, which are also part of this PR. Essentially, the transition to Infoblox v2 provides a springboard for more customized interaction with IBX.

The provider has already been tested on a live system.


k0da (approved):

k0da:
Fixes https://github.com/k8gb-io/k8gb/issues/1863

PR #1874, opened 2025-03-31
GO v1.24.1, golangci-lint v2.0.2, bump references

kuritka:
bumping references 

external-dns module requires (Go >= 1.24.0) which is not supported by current golangci-lint so I need to upgrade golangci-lint  to `v2.0.2` which requires [config migration](https://golangci-lint.run/product/migration-guide/). 

```
# locally you must:
brew upgrade go
```


.github/workflows/chainsaw.yaml
```diff
@@ -58,7 +58,7 @@ jobs:
         run: chainsaw version
 
       - name: Setup golang
-        uses: actions/setup-go@5a083d0e9a84784eb32078397cf5459adecb4c40
+        uses: actions/setup-go@0aaccfd150d50ccaeb58ebd88d36e91967a5f35b
```
abaguas, line 61:
```suggestion
        uses: actions/setup-go@f111f3307d8850f501ac008e886eec1fd1932a34 # v5.3.0
```
Suggesting to pin it to the same version as the rest of the usages of this action. Then renovate will bump them all

abaguas (approved):

PR #1876, opened 2025-04-02
[breaking] deprecate dnsZone and edgeDnsZone chart values

### LINKED ISSUE #1858: dnsZones block unintentionally introduced a breaking change
#1845 introduced the possibility to configure multiple DNS zones.
The helm chart was extended with the key `dnsZones:`, and the old `dnsZone`/`edgeDNSZone` were kept for backwards compatibility. However, the old configuration no longer works since `dnsZones` is used in the default `values.yaml` and `values.schema.json` makes sure only one of them is set.

### How to reproduce
Create an additional values file, `values-abaguas.yaml`:
```
k8gb:
  dnsZone: "cloud.abaguas.com"
  edgeDNSZone: "abaguas.com"
```

Generate the manifests using:
```
hem package -u . > /dev/null && helm template k8gb k8gb-v0.14.0.tgz --include-crds -f values.yaml -f values-abaguas.yaml > manifests.yaml
```
This results in an error:
```
Error: values don't meet the specifications of the schema(s) in the following chart(s):
k8gb:
- k8gb: Must validate one and only one schema (oneOf)
```

### Fix suggestions
k8gb is on major version `0`, so it is okay to introduce a breaking change and remove completely the old option, thus forcing all users to migrate.

Alternatively, a simple fix is to enable only the old configuration options in the default `values.yaml`

abaguas:
PR #1845 introduced the possibility to configure multiple DNS zones. The intention was to be backwards compatible. However it forces uses to change the new helm values, as explained in #1858.

The naming of zones was also reviewed, since edgeZone was not very clear. So the change is as follows:
* k8gb.dnsZoneNegTTL -> k8gb.dnsZones[0].dnsZoneNegTTL
* k8gb.edgeDnsZone -> k8gb.dnsZones[0].parentZone
* k8gb.dnsZone -> k8gb.dnsZones[0].loadBalancedZone

During the community meeting of 02.04.2025 it was agreed to maintain the breaking change and clean up the deprecated values configuration.

Fixes #1858


abaguas:
@kuritka can we rediscuss the new naming of the options?
* edgeDnsZone -> zone (example.com)
* dnsZone -> domain (cloud.example.com)

What is for you the different between zone and domain?
If I consider domain as an fqdn then "cloud.example.com" is not an fqdn. The fqdn will be "app.cloud.example.com".
If I consider domain as a portion of an fqdn then it is the same as zone.

Can you explain your reasoning for the change? What was confusing with the previous naming?

kuritka:
Hey, totally valid question – both naming schemes actually make sense.

I renamed `edgeDnsZone` -> `zone` and `dnsZone` -> `domain` mainly to make the hierarchy clearer and more intuitive for users configuring it:

`zone` is the top-level (e.g. `example.com`)

`domain` is a `sub-part` of it (e.g. `cloud.example.com`)

You're right that technically both are `zones`, and using `edgeDnsZone` / `dnsZone` is more DNS-accurate.
But the old naming was a bit hard to read without knowing the internal distinction between "edge" and "non-edge".

Happy to revisit if we find better names that keep both clarity and accuracy – e.g. rootZone / subZone could be a compromise.

abaguas:
I agree edge zone is not very clear.
> Why do you feel such a need to change it?
I feel calling domain to one of the zones will also lead to confusion. So, since we changed it we can get it 100% accurate before the upcoming release.

Root zone is the zone controlled by the root servers. But I like the idea of hierarchy it brings. What do you think about "zone"/"k8gbZone"/"delegatedZone" and "parentZone"?

kuritka:
> Root zone is the zone controlled by the root servers. But I like the idea of hierarchy it brings. What do you think about "zone"/"k8gbZone"/"delegatedZone" and "parentZone"?

apart `k8gbZone` (which, anyone can interpret differently) im fine with all of them.  thx!

ytsarev (approved):
LGTM. We will have to work on next release notes explicitly mentioning that change

chart/k8gb/values.yaml
```diff
@@ -16,18 +16,11 @@ k8gb:
...
-  # dnsZone: "cloud.example.com"
-  # -- edgeDNSZone: deprecated
-  # edgeDNSZone: "example.com"
-  # -- array of dns zones controlled by gslb§
+  # -- array of dns zones controlled by gslb
   dnsZones:
-    - zone: "example.com" # -- main zone which would contain gslb zone to delegate (same meaning as to edgeDNSZone)
-      domain: "cloud.example.com" # -- domain controlled by gslb (same meaning as to dnsZone)
+    - parentZone: "example.com" # -- parent zone which would contain gslb zone to delegate (same meaning as to edgeDNSZone)
+      zone: "cloud.example.com" # -- zone controlled by gslb (same meaning as to dnsZone)
```
ytsarev, line 22: Maybe loadBalancedZone to be explicit?

chart/k8gb/values.schema.json
```diff
@@ -340,24 +328,11 @@
...
-                        "clusterGeoTag",
-                        "extGslbClustersGeoTags",
-                        "edgeDNSServers",
-                        "dnsZones"
-                    ]
-                }
+            "required": [
+                "clusterGeoTag",
+                "extGslbClustersGeoTags",
+                "edgeDNSServers",
```
ytsarev, line 334: We need to change this terminology too to be consistent

kuritka (approved):
thx @abaguas

chart/k8gb/values.schema.json
abaguas: Thanks, good catch
https://github.com/k8gb-io/k8gb/pull/1876/commits/36382b760f64a64d8437c82a55bf488d850f514f

chart/k8gb/values.yaml
abaguas: Sounds good to me. @kuritka?

abaguas: Changed done here: https://github.com/k8gb-io/k8gb/pull/1876/commits/e573eb9604b78ee2a553a94dd892d0cb4145029a

PR #1878, opened 2025-04-04
Bootstrap refactored

kuritka:
Just like before, when coreDNS `serviceType=LoadBalancer`, the IP addresses from this service will be used for `delegationDNSEndpoint`. The change is that if the serviceType is not LoadBalancer, the IP addresses of the first found ingress with the label `k8gb.io/ip-source="true"` will be used instead. Additionally, I refactored the CoreDNS controller setup and the assistant.

Service type  is determined by our Helm `values.yaml` file, specifically by the coredns.serviceType value.


taskfile.yaml
```diff
@@ -0,0 +1,180 @@
+# https://taskfile.dev/usage/
```
abaguas, line 1: Is this part of your local setup?

abaguas (dismissed):
Looks good, thanks for the effort

taskfile.yaml
kuritka: thx for info, removed

abaguas (approved):

PR #1882, opened 2025-04-17
Temporarly switch from https://k8gb.io to https://k8gb.rocks

kuritka:
Due to the temporary outage of the k8gb.io domain, I'm switching to [https://k8gb.rocks](https://k8gb.rocks).

Fixes pipelines and local environment. 

I didn't get RELEASE pipeline into account.


ytsarev:
https://www.k8gb.io/ is up

kuritka:
oh, perfect 👌 @ytsarev. At least we know it works.

PR #1888, opened 2025-04-22
Add support for ExternalName service health checks

### LINKED ISSUE #1887: Support ExternalName Service type
Trying to use K8gb with the KEDA [HTTP Add-on](https://github.com/kedacore/http-add-on/tree/main) is currently not possible when KEDA is installed in cluster-global mode.

In an example where an `Ingress` host should be autoscaled, the add-on requires that an `Ingress` target the interceptor proxy Service, and not the usual workload Service. This allows the interceptor to scale based on the HTTP requests being proxied and the `requestRate` trigger configuration etc.

As an example:

```yaml
kind: Ingress
metadata:
  name: keda-http-example
  namespace: example
  labels:
    app: keda-http-example
spec:
  rules:
  - host: keda-http-example.app.test.com
    http:
      paths:
      - backend:
          service:
            name: keda-http-proxy
            port:
              name: proxy
        path: /
        pathType: Prefix
---
kind: Service
apiVersion: v1
metadata:
  name: keda-http-proxy
  namespace: keda
spec:
  type: ExternalName
  # Where the KEDA HTTP Add-on interceptor proxy Service (type: ClusterIP) host is as below
  externalName: keda-add-ons-http-interceptor-proxy.keda.svc.cluster.local
```

These resources should work as expected. However, if we add a `Gslb` resource to this. We would see the `status` show a `NotFound` for the proxy Service, and therefore not resolve the host as expected:

```yaml
kind: Gslb
metadata:
  name: keda-http-example
  namespace: example
spec:
  resourceRef:
    kind: Ingress
    matchLabels:
      app: keda-http-example
  strategy:
    dnsTtlSeconds: 30
    primaryGeoTag: zone-1
    type: roundRobin
status:
  geoTag: zone-1
  healthyRecords:
    keda-http-example.app.test.com:
    - 10.x.x.x
  hosts: keda-http-example.app.test.com
  loadBalancer:
    exposedIps:
    - 10.x.x.x
  servers:
  - host: keda-http-example.app.test.com
    services:
    - name: keda-http-proxy
      namespace: example
  serviceHealth:
    keda-http-example.app.test.com: NotFound
```

This makes sense because there is no `Service` called `keda-http-proxy` in the `example` namespace. Even if there was, it would be a `type: ExternalName` service, which does not have any `Endpoint`s to check.

To enable this use case, it would be nice if the backend `Service` from the referenced `Ingress` could be of type `ExternalName`, and have it properly checked for the `serviceHealth`.

donovanmuller:
This PR adds health status checking for [ExternalName](https://kubernetes.io/docs/concepts/services-networking/service/#externalname) services. The implementation:

- Handles `ExternalName` services pointing to internal Kubernetes services by checking the target service's endpoints
- Automatically marks `ExternalName` services pointing to external DNS as healthy
- Maintains backward compatibility with existing ClusterIP service health checks

This enhancement allows a `Gslb` resource to properly handle `type: ExternalName` services.

Fixes #1887


donovanmuller:
Closing due to no longer requiring this functionality.

PR #1889, opened 2025-04-23
ExtendResourceRef, Change ingress lifecycle

kuritka:
- Extend ResourceRef for name, namespace, kind, so we can do both: 
```yaml
kind: Gslb
metadata:
  name: playground-failover
spec:
  resourceRef:
    kind: Ingress
    matchLabels:
      app: playground-failover
```
```yaml
kind: Gslb
metadata:
  name: playground-failover
spec:
  resourceRef:
    kind: Ingress
    name: playground-failover-ingress
    namespace: playground
``` 
 - refactor ingress handler into standalone testable class, change update in lifecycle
 - fix terratest
 - unit tests
 - Chainsaw tests


k0da (approved):

ytsarev (approved):

ytsarev:
@k0da @kuritka 

I have some afterthoughts after approval of this one.

```
kind: Gslb
metadata:
  name: playground-failover
spec:
  resourceRef:
    kind: Ingress
    name: playground-failover-ingress
    namespace: playground
```

`namespace: playground` why do we need namespace here? Don't we expect Gslb to reside in the same namespace as the target Ingress?

k0da:
with `corev1.ObjectReference` you can refer exact namespace scoped resource.

ytsarev:
@k0da Is there a case when GSLB and Ingress reside in different namespaces?

k0da:
@ytsarev no, rather Ingress without label or label is not unique

ytsarev:
@k0da so only `name` reference makes sense here? `namespace` should be implied from Gslb namespace?

k0da:
@ytsarev need to check, but yes, it should be

ytsarev:
CC @kuritka it looks like we need follow up to this one

k0da:
@ytsarev it [reads](https://github.com/k8gb-io/k8gb/pull/1889/files#diff-1293ee233fafd30c2004b9c61a4f3f2209a179f30b39244ed84cb82b586a7dd0R191) gslb for namespace:

ytsarev:
Why do we need this [embedded namespace](https://github.com/k8gb-io/k8gb/pull/1889/files#diff-d14e80a23f0ff8bd02e072b90944a8d98d36055c28fb638a3cc4c6e1a3b99092R12) in the reference then?

k0da:
We [don't](https://github.com/k8gb-io/k8gb/pull/1889/files#diff-3b7b160d86aa9890de27b7095494896b96f3f31951273426e83c616946348635R43-R49) need it.

ytsarev:
then we need to fix docs/tests. What we have now is misleading.

k0da:
@ytsarev https://github.com/k8gb-io/k8gb/pull/1903 should address your concerns

ytsarev:
@k0da perfect, thanks a ton!

PR #1890, opened 2025-04-25
Remove GSLB-derived annotations from generated Ingress and add management label

kuritka:
When we apply GSLB, an Ingress is created automatically. This new Ingress had annotations generated from the GSLB values (strategy, dnsTTLSeconds, primary region). However, since GSLB is used as the single source of truth, these annotations have no place there. Additionally, I'm adding the label `app.k8gb.io/managed-by: gslb` to this generated Ingress to make it easy to select these Ingresses.

```yaml
kind: Ingress
metadata:
  name: dependent-ingress
  annotations:
   # they will not be maintained in dependent ingresses
    k8gb.io/strategy: failover
    k8gb.io/primary-geotag: "eu"
    k8gb.io/dns-ttl-seconds: "5"
  ownerReferences:
  - gslb reference ...
```

Second case - when we create a GSLB from an Ingress, the annotations of course remain in the Ingress. The test coverage is in the `TestAnnotations` Terratest.

### DIFF
--- a/controllers/gslb_controller_reconciliation.go
+++ b/controllers/gslb_controller_reconciliation.go
@@ -106,13 +106,13 @@ func (r *GslbReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 
 	// == Ingress ==========
 	if reflect.DeepEqual(gslb.Spec.ResourceRef, k8gbv1beta1.ResourceRef{}) {
-		ingress, err := r.gslbIngress(gslb)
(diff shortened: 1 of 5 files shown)


ytsarev (commented):
Now that we've moved to a non-intrusive referencing style, should we consider deprecating and removing the Ingress creation pattern?

k0da (approved):

ytsarev (approved):
I like this change 👍  Let's discuss the idea of the full ingress creation deprecation at the next community meeting.

PR #1897, opened 2025-05-05
Create automated test setup for AWS route53

abaguas:
Automates testing of k8gb's integration with AWS route53. Previously this required manual steps (#1849), now fully automated.

To test, simply install opentofu, create an AWS account, retrieve credentials and run `./test.sh`.


kuritka (dismissed):

dns-provider-test/route53/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+Install opentofu to provision the infrastructure
+```
+brew install opentofu
+```
+
+### AWS account
+
+First you will need an AWS account, if you don't have one already you can get started with a [free tier account](https://aws.amazon.com/free).
+Afterwards, create access keys in the portal `IAM Dashboard -> My security credentials -> Access keys` and use then to login using `aws configure`. This will make the credentials available to terraform.
```
k0da, line 17: Can we prefer IRSA setup or Pod identity(not sure external-dns supports that)?

abaguas: This part is only for opentofu and aws cli commands, not for access from the cluster. We can have a more secure method for sure.
The opentofu creates a user and then we get a key `SECRET_ACCESS_KEY=$(aws iam create-access-key --user-name "externaldns")`. Here again, we have multiple methods available: https://github.com/kubernetes-sigs/external-dns/blob/8d8f81369f02beeba479944e26f55dfbcb53195d/docs/tutorials/aws.md, IRSA being one of them

Let me improve in the next iteration

abaguas: Well, I am some merge conflicts. So it is a good chance to implement IRSA

abaguas: I just picked this up again. IRSA is not possible for clusters outside of AWS, as it is the case here where we try to connect the local cluster to route53. So I kept the setup where we generate a static access key for the `externaldns` user. It is temporary anyways, the scripts deletes it right after it finishes.

PR #1898, opened 2025-05-06
Waiting for ingress IP addresses in the local environment, fix logger…

kuritka:
If the bootstrap address does not exist, k8gb fails and the pod restarts. Unfortunately, the logger was used incorrectly — the Interface() function was used, which performs unmarshalling and caused panics. The logger has now been fixed.

Additionally, in the local environment, it is necessary to wait until the ingress is bound; otherwise, the k8gb pod keeps restarting during tests until the ingress is installed.


controllers/providers/k8gbendpoint/helper.go
```diff
@@ -39,7 +40,7 @@ func saveDNSEndpoint(ctx context.Context, client client.Client, namespace string
 
 		// Create the DNSEndpoint
 		logger.Info().
-			Interface("DNSEndpoint", i).
+			Str("DNSEndpoint", fmt.Sprintf("%s.%s", i.Namespace, i.Name)).
```
k0da, line 43: pls use namespace/name pattern

k0da (dismissed):
LGTM

k0da (approved):

PR #1902, opened 2025-05-10
improve default values of cluster geo tags

abaguas:
FIXME gives the impression that there is still work to be done.
    We decided to give it meaningful defaults and to improve the documentation.

PR #1913, opened 2025-05-15
Update also tags in chart's default values

abaguas:
In #1902 I updated the helm docs but forgot to update the default values.
This PR fixes that misalignment


kuritka (approved):
I will probably need to expand these texts due to #1914

ytsarev (approved):

k0da (approved):

PR #1914, opened 2025-05-15
Dynamic GeoTags

kuritka:
Dynamic geotags are loaded from NS records when `EXT_GSLB_CLUSTER_GEO_TAGS` in the configuration is empty. Otherwise, the static geotags from `EXT_CLUSTER_GEOTAGS` are used, as before.

Everything translates into two additional dig commands per GSLB:

The first dig randomly selects an edgeDNS and retrieves the IP address of the DNS where the parent zone resides. The subsequent dig retrieves NS records for the loadBalancer zone, from which geotags are extracted.

*⚠️ WARNING:* Although these are only two additional dig commands, with a large number of GSLBs and short reconciliation intervals, your DNS server could become overwhelmed! If this occurs:
- You can increase the number of DNS servers in `EDGE_DNS_SERVERS`. During reconciliation, we always interact with just one of them, selecting it randomly (round-robin).
- Increase `RECONCILE_REQUEUE_SECONDS` or set it to 0 - in this case, reconciliation is triggered only upon changes to GSLB, Ingress, DNSEndpoint, or during bootstrap.
- Revert to using `EXT_GSLB_CLUSTERS_GEO_TAGS`.

_We read ns records for **LoadBalancedZone** and parse available tags from that. The cycle goes through all **parentDNSServers**_

```shell
/var/lib/bind/zones # cat k8s.zone
$ORIGIN .
$TTL 30 ; 30 seconds
example.com             IN SOA  example.com. root.example.com. (
                                23         ; serial
                                30         ; refresh (30 seconds)
                                30         ; retry (30 seconds)
                                30         ; expire (30 seconds)
                                30         ; minimum (30 seconds)
                                )
                        NS      ns.example.com.
$ORIGIN example.com.
cloud                   NS      gslb-ns-eu-cloud    <---- finding these NS records and extract geotag 
                        NS      gslb-ns-us-cloud
gslb-ns-eu-cloud        A       172.18.0.6
                        A       172.18.0.7
gslb-ns-us-cloud        A       172.18.0.10
                        A       172.18.0.11

```


controllers/geotags/dynamic.go
```diff
@@ -0,0 +1,102 @@
...
+		if len(parts) != 2 {
+			continue
+		}
+		// parts[0] == prefix; parts[1] == <region>-test-cloud.example.com
+		tag := strings.Split(parts[1], "-")[0]
+		if tag != p.config.ClusterGeoTag {
+			extTags = append(extTags, tag)
+		}
+	}
+	sort.Strings(extTags)
```
k0da, line 100: IMO we should exclude `own` tag

controllers/providers/k8gbendpoint/applicationDNSEndpoint.go
```diff
@@ -187,7 +189,15 @@ func (d *ApplicationDNSEndpoint) GetDNSEndpoint() (*externaldns.DNSEndpoint, err
 
 func (d *ApplicationDNSEndpoint) GetExternalTargets(host string) (targets Targets) {
 	targets = NewTargets()
-	for tag, cluster := range d.config.DelegationZones.GetExternalClusterNSNamesByHostname(host) {
+	gt, err := geotags.GeoTag(d.config).GetExternalClusterNSNamesByHostname(host)
```
k0da, line 192: We need to pass `local` tag so we return only external ones

controllers/geotags/dynamic.go
kuritka: sure, we have it: https://github.com/k8gb-io/k8gb/pull/1914/files#diff-42234b232ea30a245f58cd1db685055d2b547cb01090dd25e4a8389befd048a3R96-R98

controllers/providers/k8gbendpoint/applicationDNSEndpoint.go
kuritka: it's there we pass whole *config so we know the current geotag

controllers/geotags/dynamic.go
k0da: I see

k0da (approved):

ytsarev:
@kuritka @k0da, are there any docs for this substantial change?

PR #1915, opened 2025-05-15
FIX: internal logr adapter panics

kuritka:
The controller-runtime uses logr internally. We're using zerolog. We have an adapter that bridges logr to call zerolog. However, after various upgrades, this adapter started to panic. Here's the fix.

### DIFF
--- a/controllers/logging/logr.go
+++ b/controllers/logging/logr.go
@@ -76,22 +76,37 @@ func (a *logrSinkAdapter) Error(err error, msg string, keysAndValues ...interfac
 }
 
 func (a *logrSinkAdapter) WithValues(keysAndValues ...interface{}) logr.LogSink {
+	newKV := make(map[string]string)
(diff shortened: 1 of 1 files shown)


controllers/logging/logr.go
```diff
@@ -76,22 +76,38 @@ func (a *logrSinkAdapter) Error(err error, msg string, keysAndValues ...interfac
...
+		keysAndValues: newKV,
+		name:          a.name,
 	}
-	return a
 }
 
 func (a *logrSinkAdapter) WithName(name string) logr.LogSink {
-	a.name = name
-	return a
+	// Vytvoř nový adapter s kopií hodnot
```
ytsarev, line 101: it's lovely to see Czech in the codebase, but can we translate this comment to English for the international community :)

ytsarev (approved):

PR #1928, opened 2025-05-28
Refactor depresolver to resolver

### LINKED ISSUE #1870: Refactor depresolver
We have the option to significantly slim down depresolver. Consider using libraries like kong.CLI for loading environment variables and go-playground/validator for custom validations. This would greatly reduce the amount of testing and overall maintenance.

Consider removing environment variables such as SplitBrain, EdgeDNSZone, and DNSZone. Also, consider introducing new environment variables as suggested here: https://github.com/k8gb-io/k8gb/pull/1864#pullrequestreview-2714838552

kuritka:
closes #1870

The previous depresolver was difficult to extend and suffered from poor test maintainability. This refactoring simplifies the depresolver and introduces a new module called resolver, which will be used to retrieve and validate configuration data outside of k8gb.

For loading ENV vars into the Config entity, we use the popular github.com/alecthomas/kong.

For validation, we use github.com/go-playground/validator/v10, although some validators (especially structural ones that validate rules across multiple fields) had to be implemented manually.

We have brand new tests that are much easier to maintain.

Field tags on the Config entity now contain all necessary information regarding validation and ENV var mapping.

Since the resolver interface is almost identical to the former depresolver interface, we could create a mock from the resolver and swap the implementation without significant refactoring outside the resolver module itself.

Signed-off-by: Michal Kuritka <kuritka@gmail.com>


ytsarev (dismissed):
Looks great, just needs rebase

k0da (approved):

ytsarev (approved):

PR #1929, opened 2025-05-28
FIX: HELM Disable validation - parentZones equal to extDNSZones

kuritka:
Skippimg this validation for infoblox

Signed-off-by: Michal Kuritka <kuritka@gmail.com>


k0da (dismissed):

abaguas (dismissed):
Good catch, thanks!

Issue #1921, opened 2025-05-18
Switch to EndpointSlice API


k0da:
Endpoint API is deprecated and 1.33 throws a warning. We must switch to EndpointSlice for ServiceHealth logic.

Details explained [here](https://kubernetes.io/blog/2025/04/24/endpoints-deprecation/)

ljluestc:
```

// main.go
package main

import (
	"fmt"
	"net/http"
)

func main() {
	// Health check endpoint
	http.HandleFunc("/health", func(w http.ResponseWriter, r *http.Request) {
		// Add custom health logic here, e.g., check database connection
		// For simplicity, always return healthy
		fmt.Fprintf(w, "OK")
	})

	// Main application endpoint
	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		fmt.Fprintf(w, "Hello, Kubernetes!")
	})

	fmt.Println("Starting server on :8080")
	if err := http.ListenAndServe(":8080", nil); err != nil {
		fmt.Printf("Server failed: %v\n", err)
	}
}
```

jkremser:
found these:
- https://github.com/k8gb-io/k8gb/blob/4f24a4ae5e30b03bdeca83468f0b86a1a13f00a5/controllers/gslb_controller_setup.go#L87
- https://github.com/k8gb-io/k8gb/blob/4f24a4ae5e30b03bdeca83468f0b86a1a13f00a5/controllers/gslb_controller_setup.go#L63 (name of the `Endpoint` resource is no longer the same as the name of the service, the link is now made via label on the `EndpointSlice`)
- https://github.com/k8gb-io/k8gb/blob/4f24a4ae5e30b03bdeca83468f0b86a1a13f00a5/chart/k8gb/templates/role.yaml#L20 (rbac)

PR #1940, opened 2025-06-16
Switch from Endpoints to EndpointSlice API (deprecation)

jkremser:
fix https://github.com/k8gb-io/k8gb/issues/1921

endpointslices now have the `.metadata.generation` so also removing that exception for endpoints in the `EventFilter`. This is actually a breaking change for older k8s versions that didn't have the EndpointSlice api present (before 1.21), so we may want to have some similar compatibility matrix as we had for the ingress v1 vs v2, or just bump the minimum k8s requirement in the [readme](https://github.com/k8gb-io/k8gb/blob/master/README.md?plain=1#L124) & [here](https://github.com/k8gb-io/k8gb/blob/master/chart/k8gb/Chart.yaml#L8)

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ k8gb is tested with the following environment options.
 
 | Type                             | Implementation                                                               |
 |----------------------------------|------------------------------------------------------------------------------|
-| Kubernetes Version               | >= `1.19`                                                                    |
+| Kubernetes Version               | >= `1.21`                                                                    |
 | Environment                      | Any conformant Kubernetes cluster on-prem or in cloud                        |
 | Ingress Controller               | NGINX, AWS Load Balancer Controller [*](#clarify)                            |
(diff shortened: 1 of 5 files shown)


k0da:
I don't think we need to keep backward compatibility here, as Endpoint api is long time deprecated and it is not something exposed to users to interact with

ytsarev:
we should specify `>=1.21` requirement in the compatibility matrix after this change

jkremser:
- [x] bump the minimum k8s requirement in the [readme](https://github.com/k8gb-io/k8gb/blob/master/README.md?plain=1#L124) & [here](https://github.com/k8gb-io/k8gb/blob/master/chart/k8gb/Chart.yaml#L8)

kuritka (approved):

ytsarev (approved):
Great stuff, thank you 👍

PR #1959, opened 2025-06-30
Describe breaking changes in v0.15.0

abaguas:
This PR is not meant to be merged as is, the goal is to keep track of the changes so that we can update the changelog accordinly when we do the v0.15.0 release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,100 @@
 # Changelog
 
+**Breaking changes:**
+
+v0.15.0 contains two breaking changes in the Helm chart. One relates to DNS zone configuration, which affects all users. The second one only affects users using the AWS Route53 integration.
+
+---
+
+DNS Zones
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,100 @@
...
+extdns:
+  env:
+  - name: AWS_DEFAULT_REGION
+    value: "us-east-1"
+```
+
+Finally, a couple of variables must be specified (there are helm validation function that make sure they are correct). Replace <GEOTAG> with the same value as k8gb.clusterGeoTag, and domainFilters with the same values as k8gb.dnsZones.parentZone:
+```
+extdns:
+  txtPrefix: "k8gb-<GEOTAG>"
```
k0da, line 91: JFYI: @kuritka is working on concurrent NS record updates from two or more clusters

ytsarev (commented):
@abaguas this is great initiative!  Maybe we should move it to the separate front-facing document. 

Arguments for the move:
* CHANGELOG.md is automatically generated
* We need to share the changes about release easily from both github and website

PR #1971, opened 2025-07-10
created docs site with mkdocs

itsfarhan:
1. Created MkDocs Configuration
2. Restructured Documentation
3. Created docs/tutorials.md with organized installation guides
4. Updated README.md 
5. Fixes from previous pull request -> https://github.com/k8gb-io/k8gb/pull/1851
  -   Fixed navigation
  -   Fixed Broken Links & Images
  -   Fixed image paths in components.md, deploy_windowsdns.md, local-kuar.md, deploy_cloudflare.md, local.md
  -   Fixed internal links in md files
6. All paths now work with MkDocs relative referencing

Live demo deployment on github pages -> https://itsfarhan.github.io/k8gb/

PR #1972, opened 2025-07-12
Revert "created docs site with mkdocs"

ytsarev:
Reverts k8gb-io/k8gb#1971


ytsarev:
CC @itsfarhan current mkdocs deploy removes chart repo from gh-pages, we need to find way to keep it around

k0da (approved):

PR #2029, opened 2025-08-20
Add LoadBalancer Service support via hostname annotation for LB service

sudhamshk:
### TLDR

This PR introduces first-class support for using Kubernetes Service resources of type LoadBalancer as backends for Gslb. It enhances the resolver logic, simplifies service targeting using label selectors, and automates Gslb creation via annotations.

### Technical Implementation
New Components

lbservice Resolver (controllers/refresolver/lbservice/)

- Resolves LoadBalancer service references in GSLB resources
- Handles hostname validation and IP address resolution
- Supports both name-based and label-based references

Enhanced Query Options (controllers/refresolver/lbservice/common/queryopts.go)

- Implements namespace fallback logic
- Consistent with Ingress resolver behavior

Integration Points

- Resolver Factory: Enhanced refresolver to support Service kind

### Testing
Unit Tests

- Tests GetServers and GetGslbExposedIPs

Integration

- Existing functionality: No breaking changes to existing Ingress integration
- Backward compatibility: Manual GSLB creation still works
- Error handling: Proper error messages for missing annotations

### Documentation
New Documentation

- docs/lbservice-integration.md - Manual GSLB creation for LoadBalancer services
- docs/examples/lbservice-gslb-example.yaml - Manual GSLB example

### Validation & Error Handling
Service Validation

- Only LoadBalancer services are processed

### Error Messages

- Clear error messages for missing required annotations
- Validation of hostname format and DNS zone compatibility
- Proper logging for debugging and troubleshooting

### Breaking Changes

- None: This is a purely additive feature
- Existing functionality: All existing Ingress and manual GSLB creation continues to work
- API compatibility: No changes to existing GSLB API

### Configuration

Constants

- hostnameAnnotation = "k8gb.io/hostname" - Centralized annotation constant
- Updated k8gbAnnotations slice to include hostname annotation
- Consistent annotation handling across handlers

Thanks @infa-dimukher for pairing

TODO in a follow up PR - add chainsaw tests


controllers/gslb_controller_reconciliation.go
```diff
@@ -122,7 +126,13 @@ func (r *GslbReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 	}
 
 	// == Reference resolution ==
-	refResolver, err := refresolver.New(gslb, r.Client)
+	var refResolver refresolver.GslbReferenceResolver
```
abaguas: Since the "config" is no longer used we can revert this block, and remove the `NewWithConfig` and `NewReferenceResolverWithConfig` functions.

controllers/handlers.go
```diff
@@ -200,7 +201,203 @@ func (g *IngressHandler) getGslb(obj client.Object) (*k8gbv1beta1.Gslb, bool, er
...
+		if r.Kind == "Gslb" {
+			return true
+		}
+	}
+	return false
+}
+
+func (g *ServiceHandler) isK8gbAnnotated(obj client.Object) bool {
+	annotations := obj.GetAnnotations()
+	_, found := annotations[strategyAnnotation]
```
abaguas: We are actually moving away from such a setup.

The ingress integration had this option where one could set annotations in the Ingress resource, which would then automatically create a GSLB resource. However, we had some issues where multiple controllers were trying to control the annotations on a resource.
We then decided to prefer a setup where a GSLB resource needs to be explicitly created, so that we don't annotate or watch annotations on resources we do not control.

So, we don't need to implement this feature for Services, we don't need to watch Service resources.

Is this alright for your setup?

abaguas (commented):
Thank you for the contribution. It makes sense to add the annotation for hostname now that we support multiple DNS zones

controllers/handlers.go
sudhamshk: this is a very important and outstanding feature

The platform team deploys and operates the k8gb controller; application teams merely add an annotation on their LoadBalancer Service to request a GSLB. This lowers the barrier for early adopters, keeps ownership clear, and avoids app teams having to learn or manage new CRDs on day one. In practice, this has driven faster experimentation and rollout with minimal platform intervention.

Could we please retain annotation-driven auto-GSLB for a while longer?

controllers/gslb_controller_reconciliation.go
sudhamshk: Done

controllers/handlers.go
abaguas: I fully agree with the separation of responsibilities. However, I think all the properties you want to achieve can be achieved with a GSLB resource.

The platform team deploys and operates the k8gb controller, application teams simply add a GSLB resource. That is the concept behind CRDs.
The ownership is clear, the application team manages the GSLB resource.

Why is managing annotations simpler than managing GSLB resources? How can a user find out which annotations are available? With a GSLB resource any user can do `kubect explain gslb`, see up-to-date documentation of all fields and profit from validation.

sudhamshk: Thanks, totally agree on the advantages of CRDs (discoverability with kubectl explain, validation, clear ownership).

When I first tried this with annotations, it really made my day easier and gave a strong sense of easy adoption lowering the barrier for new teams to try k8gb. I know the pros/cons have already been discussed, but I felt this experience is worth sharing as it can drive more people to experiment.

We don’t need full functionality via annotations just the basics with clear limitations. Once they see the benefits, they’ll be willing to go the extra mile to create and manage the Gslb resource (and adapt their Helm charts, etc.). A short note in the README would be perfect to highlight this quick-start path alongside the CRD approach.

As a middle ground, since Ingress annotations are already supported, could we keep Service annotations too and remove both together later when the community feels ready?

abaguas: Thank you for the feedback. It helps us learning your experience.

> As a middle ground, since Ingress annotations are already supported, could we keep Service annotations too and remove both together later when the community feels ready?

This is very unlikely to happen ;) If we add a feature to the codebase now, we don't want to remove it later.

The tradeoff here is functionality vs maintainability. We are a small team, so we need to choose wisely which features to implement and maintain. In this case, we would not be adding functionality, we would be providing an additional way of achieving what can already be done with the GSLB resource.

Let me share this with the other maintainers

controllers/gslb_controller_reconciliation.go
```diff
@@ -107,6 +109,8 @@ func (r *GslbReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		Msg("Resolved strategy")
 
 	// == Ingress ==========
+	// Only create embedded Ingress if ResourceRef is empty (embedded mode)
+	// Skip Ingress creation for LoadBalancer services as they don't need an Ingress
```
abaguas: This comment can be removed, after you removed the LoadBalancer logic

controllers/handlers.go
```diff
@@ -200,7 +201,203 @@ func (g *IngressHandler) getGslb(obj client.Object) (*k8gbv1beta1.Gslb, bool, er
...
+func (g *ServiceHandler) Handle(svc client.Object) []reconcile.Request {
+
+	// following lines filters only Services with strategyAnnotation which are not owned by GSLB
+	// filtering out Service without k8gb strategy
+	if !g.isK8gbAnnotated(svc) {
+		return nil
+	}
+
+	// filtering out Service which is owned by GSLB
+	if g.isOwnedByGSLB(svc) {
```
abaguas: This condition is not needed. With an embedded GSLB we were creating an Ingress resource from a GSLB. This functionality is deprecated in favor of resource references and will not be provided for services

```
kind: Gslb
metadata:
  name: roundrobin-ingress-embedded
spec:
  ingress:
    ingressClassName: nginx
    rules:
      - host: roundrobin-ingress-embedded.cloud.example.com
        http:
          paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: frontend-podinfo
                port:
                  name: http
  strategy:
    type: roundRobin
    dnsTtlSeconds: 5
```

controllers/handlers_test.go
```diff
@@ -226,3 +229,122 @@ func TestIngressHandler(t *testing.T) {
...
+	assert.NoError(t, err, "GSLB should be created")
+	assert.Equal(t, "test-service", gslb.Name)
+	assert.Equal(t, "default", gslb.Namespace)
+	assert.Equal(t, "Service", gslb.Spec.ResourceRef.Kind)
+	assert.Equal(t, "v1", gslb.Spec.ResourceRef.APIVersion)
+	assert.Equal(t, "test-service", gslb.Spec.ResourceRef.Name)
+	assert.Equal(t, "roundRobin", gslb.Spec.Strategy.Type)
+	assert.Equal(t, "test.example.com", gslb.Annotations["k8gb.io/hostname"])
+
+	// Test 5: Service owned by GSLB should not trigger new GSLB creation
```
abaguas: As mentioned above, no service will be owned by k8gb, so this test case can be removed

controllers/refresolver/lbservice/common/queryopts.go
```diff
@@ -0,0 +1,81 @@
+package common
```
abaguas: Do we need to duplicate this package?
Could we move the `common` package from the `controllers/refresolver/ingress` directory to the `controllers/refresolver`and use it in all integrations?

controllers/refresolver/lbservice/lbservice.go
```diff
@@ -0,0 +1,204 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+const (
+	hostnameAnnotation = "k8gb.io/hostname"
+)
+
+// LoadBalancerServiceReconciler reconciles a LoadBalancerService object
```
abaguas: I believe this new reconciler is not used, down to line 75

abaguas (commented):
I reviewed the code as if we would include the annotation option.
We will continue that discussion. In the meantime we can do a couple of changes to have this PR closer to merge

controllers/handlers.go
abaguas: @sudhamshk, we will not be able to support the annotations. I brought up this topic with the other maintainers and they clarified some of the problems we had in the past. You can find the reasoning bellow:

> Annotations create a fundamental problem: maintaining a single source of truth.
>  * If GSLB is the source of truth and someone tweaks an Ingress annotation, should that change be pushed back into the GSLB or should the annotation be overwritten?
>  * Conversely, if a GSLB was created from annotations, what takes precedence later — editing the GSLB or editing the annotations?

> In practice this led to drift and reconciliation loops (one change of ingreass auto-propagates, triggers another reconcile of GSLB , etc.), which became very hard to debug and support.

We will also deprecate the Ingress annotations in the next release.

This PR is still very valuable, and we are very interested in merging it, but we need to remove the annotations part

sudhamshk: Understood, I will remove the annotation section and update the PR accordingly

sudhamshk: Done, Please review the changes

controllers/refresolver/lbservice/testdata/lb_service_multiple_ips.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
```
abaguas, line 1: Thanks, looks good. Can you just add a `lbservice_test.go` where you use these testdata files. This file should test the `GetServers` and `GetGslbExposedIPs` functions. You can find examples in the other integrations

sudhamshk: Done

PR #2030, opened 2025-08-20
Refactor namespace initialization in Chainsaw

abaguas:
We had different initializations for istio and non-istio setups because of the istio setup requires the `istio-inject` label on the namespace. That created some differences between tests, which made the code harder to read.

This PR consolidates the initialization in a single step template that sets the label dynamically.

The validation of the Ingress resource is also removed to allow this consolidation. This was a redundant step since a valid DNSEndpoint guarantees a valid Ingress.


ytsarev (approved):
👍

PR #2031, opened 2025-08-20
remove failover playground terratest

abaguas:
The terratest setup is quite complex, which resulted in us having to give up on test parallelism. This increased the testing time, so we are moving away from it to Chainsaw.

The failover playground test was already implemented in Chainsaw a few months ago and it has been running well. We can now remove this test from the Terratest setup and some some precious e2e testing time (~27 to 22 minutes)

---

I am also pinning the version of helm to v3.18.4, instead of using the default version in Ubuntu 24.04, since v3.18.5 broke the installation of the istio-gateway helm chart: https://github.com/istio/istio/issues/57354


kuritka:
hi @abaguas, I understand your reasoning. Still, terratest is easier to debug directly from IDE. Maybe we could let it live for a little longer and then retire it later. Did you consider using `t.Skip("skipping test")` in terratest instead of removing it right away?

ytsarev (approved):

ytsarev:
Thanks for this change!

abaguas:
> hi @abaguas, I understand your reasoning. Still, terratest is easier to debug directly from IDE. Maybe we could let it live for a little longer and then retire it later. Did you consider using `t.Skip("skipping test")` in terratest instead of removing it right away?

@kuritka we need to sync on the direction then. I don't mind which tool we use, terratest or chainsaw. But the tests should be easy to write, easy to debug and run quickly (in parallel). Let's decide if we should invest in chainsaw or terratest.

When we started exploring Chainsaw we discussed that the Terratest setup was too complex. Could you elaborate on what you are able to do with Terratest from your IDE but cannot do with Chainsaw?

kuritka:
Hi @abaguas, thanks for elaborating. I agree with you — we need tests that are simple to write, fast to execute (in parallel), and still easy to debug. That’s exactly why we decided to move towards Chainsaw and invest our efforts there.

My earlier t.Skip() suggestion was more of a temporary idea to keep Terratest around while we figure out the strategy. Since the direction is already clear, I’m fine with removing the Terratest failover playground now and focusing on Chainsaw.

So from my side, I support this PR.

abaguas:
> Hi @abaguas, thanks for elaborating. I agree with you — we need tests that are simple to write, fast to execute (in parallel), and still easy to debug. That’s exactly why we decided to move towards Chainsaw and invest our efforts there.
> 
> My earlier t.Skip() suggestion was more of a temporary idea to keep Terratest around while we figure out the strategy. Since the direction is already clear, I’m fine with removing the Terratest failover playground now and focusing on Chainsaw.
> 
> So from my side, I support this PR.

Alright, thanks! If anything happens and we want to go back git history is our friend :)

PR #2032, opened 2025-08-22
Allow duplicit values in EXT_GSLB_CLUSTER_GEOTAGS

### LINKED ISSUE #720: Deployment configuration for geotags

kuritka:
closes #720

Maintaining extClusterGeoTags is error-prone, because someone might copy-paste the ext-cluster GeoTags between cluster configurations and accidentally leave a duplicate GeoTag that matches the clusterGeoTag.

This PR basically allows me to have a configuration like this:

```shell
cluster 1:
CLUSTER_GEO_TAG: eu
EXT_GSLB_CLUSTERS_GEO_TAGS: eu, us, cz, af, ap

cluster 2:
CLUSTER_GEO_TAG: us
EXT_GSLB_CLUSTERS_GEO_TAGS: eu, us, cz, af, ap

cluster 3:
CLUSTER_GEO_TAG: cz
EXT_GSLB_CLUSTERS_GEO_TAGS: eu, us, cz, af, ap
```

previously:

```shell
# cluster 1:
CLUSTER_GEO_TAG: eu
EXT_GSLB_CLUSTERS_GEO_TAGS: us, cz, af, ap

# cluster 2:
CLUSTER_GEO_TAG: us
EXT_GSLB_CLUSTERS_GEO_TAGS: eu, cz, af, ap

# cluster 3:
CLUSTER_GEO_TAG: cz
EXT_GSLB_CLUSTERS_GEO_TAGS: eu, us, af, ap
..
```

### DIFF
--- a/controllers/resolver/config_delegationZones.go
+++ b/controllers/resolver/config_delegationZones.go
@@ -56,6 +56,10 @@ func parseDelegationZones(config *Config) ([]*DelegationZoneInfo, error) {
 	extClusterNSNames := func(zone, edge string) map[string]string {
 		m := map[string]string{}
 		for _, tag := range config.ExtClustersGeoTagsRaw {
+			if tag == config.ClusterGeoTag {
+				// skip the cluster GeoTag, see: https://github.com/k8gb-io/k8gb/issues/720
+				continue
+			}
 			m[tag] = getNsName(tag, zone, edge)
 		}
 		return m
--- a/controllers/resolver/resolver_test.go
+++ b/controllers/resolver/resolver_test.go
@@ -507,6 +507,52 @@ func TestConfigurations(t *testing.T) {
 				 INFOBLOX_WAPI_PASSWORD="testpassword"`,
 			expectedError: true,
 		},
+		{
+			name: "valid configuration without clusterGeotag in EXT_GSLB_CLUSTERS_GEO_TAGS",
+			envvars: `CLUSTER_GEO_TAG=us;
+				 EXT_GSLB_CLUSTERS_GEO_TAGS=eu;
+				 DNS_ZONES=example.com:cloud.example.com:300;
+				 EDGE_DNS_SERVERS=local.test;
+				 EXTDNS_ENABLED=true`,
+			assert: func(t *testing.T, cfg *Config) {
+				assert.Equal(t, "gslb-ns-us-cloud.example.com", cfg.DelegationZones[0].ClusterNSName)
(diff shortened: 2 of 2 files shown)


ytsarev (approved):
Definitely more friendly user experience, thank you! 👍

abaguas (commented):
Very good!
We can also add helm validation to prevent the last 2 invalid cases

PR #1856, opened 2025-03-11
Use upstream chart for external-dns for AWS integration

abaguas:
The latest version of the external-dns image requires setting the AWS_DEFAULT_REGION environment variable.
The introduction of this variable demands changes in our chart as well as changes to the user configuration. Therefore, it is a good opportunity to do breaking changes and introduce the upstream external-dns helm chart.
By using the upstream chart we can profit from all the features of the external dns community out-of-the-box. We no longer have to develop a wrapper on our side, which is currently the bottleneck to supporting further upstream DNS providers. These features include both authentication options and new upstream providers.

The change deprecates only the route53 integration, all other integrations are not affected.

## Review tips

Checkout the first commit of this PR, fb11af4, and compare the route53 integration using the old route53 integration, or the new extdns integration:
`helm package -u . > /dev/null && helm template k8gb "k8gb-v0.14.0.tgz" --include-crds -f values.yaml -f values-route53.yaml > route53.yaml`
`helm package -u . > /dev/null && helm template k8gb "k8gb-v0.14.0.tgz" --include-crds -f values.yaml -f values-extdns.yaml > extdns.yaml`

The second commit removes the route53 configuration from the chart.

## Required user changes

The following configuration changes should be included in the release notes:

#### Auth configuration
1.
```
route53:
  enabled: true
  secret: credentials
```
becomes
```
extdns:
  enabled: true
  provider:
    name: aws
  extraVolumes:
  - name: aws-credentials
    secret:
      secretName: credentials
  extraVolumeMounts:
  - name: aws-credentials
    mountPath: /.aws
    readOnly: true
```

2.
```
route53:
  enabled: true
  irsaRole: arn:aws:iam::111111:role/external-dns
```
becomes:
```
extdns:
  enabled: true
  provider:
    name: aws
  serviceAccount:
    annotations:
      eks.amazonaws.com/role-arn: arn:aws:iam::111111:role/external-dns
```

3.
```
route53:
  enabled: true
  assumeRoleArn: role
```
becomes:
```
extdns:
  enabled: true
  provider:
    name: aws
  extraArgs:
    aws-assume-role: role
```

#### General configuration
In addition, the AWS_DEFAULT_REGION must be specified using an environment variable, for example:
```
extdns:
  env:
  - name: AWS_DEFAULT_REGION
    value: "us-east-1"
```
[...]

PR #2042, opened 2025-08-29
Move NS1 configuration to external-dns

abaguas:
Similarly to what we did with AWS Route53 (#1856), we are migrating the configuration of other DNS providers to the upstream helm chart of external dns. This PR does this migration for NS1.

The ns1 configuration is removed from the k8gb helm chart, and the examples are adapted with the equivalent configuration using the upstream external dns helm chart.

k8gb's helm chart also allowed to ignoreSSL, which can be achieved with the following configuration:
```
  extraArgs:
    ns1-ignoressl:
```

For additional options, please refer to: https://kubernetes-sigs.github.io/external-dns/v0.14.0/tutorials/ns1/

---

To check it for yourself, run the following commands on master and on this branch (the configuration is equivalent, only some labels differ):
```
cd chart/k8gb/
helm template . -f ../../docs/examples/ns1/k8gb-cluster-ns1-eu-west-1.yaml --include-crds > after.yaml
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ k8gb is tested with the following environment options.
 |----------------------------------|------------------------------------------------------------------------------|
 | Kubernetes Version               | >= `1.21`                                                                    |
 | Environment                      | Any conformant Kubernetes cluster on-prem or in cloud                        |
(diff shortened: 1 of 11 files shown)


ytsarev (approved):
Great stuff 👍

PR #2043, opened 2025-08-29
Deprecate configuration of GSLB resources via annotations

abaguas:
The introduction of an ingress integration with Service resources triggered the discussion about GSLB configuration (https://github.com/k8gb-io/k8gb/pull/2029#discussion_r2288799279). After discussion between the maintainers, we decided to deprecate this feature and remove it in k8gb v0.17.0

I also introduced an Architectural Decision Record folder to track such decisions. I believe our future selves will appreciate it one day.

### DIFF
--- a/adr/0000-template.md
+++ b/adr/0000-template.md
@@ -1,0 +1,40 @@
+# ADR-0000: [Title]
+
+## Status
+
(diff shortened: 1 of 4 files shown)


ytsarev (approved):
This is great, thanks a lot for such a constructive approach 👍

PR #2045, opened 2025-08-30
Move Cloudflare configuration to external-dns

abaguas:
Similarly to what we did with AWS Route53 (#1856) and NS1 (#2042), we are migrating the configuration of DNS providers to the upstream helm chart of external dns. This PR does this migration for Cloudflare.

The cloudfloare configuration is removed from the k8gb helm chart, and the examples are adapted with the equivalent configuration using the upstream external dns helm chart.

As linked in the docs update, any cloudflare options can now be used without any changes to k8gb's helm chart as long as they are supported by external dns (https://kubernetes-sigs.github.io/external-dns/latest/docs/tutorials/cloudflare/)

---

To check it for yourself, run the following commands on master and on this branch (the configuration is equivalent, only some labels differ):
```
cd chart/k8gb/
helm template . -f ../../docs/examples/cloudflare/k8gb-cluster-cloudflare-eu.yaml --include-crds > manifests.yaml
```


ytsarev (approved):
Great job 🚀

PR #1947, opened 2025-06-19
add coredns plugin and server bocks per zone

barmettlerl:
- Allow extraPlugins and extraServerBlocks to be defined per each zone instead for all zone entries in the Corefile. 
- Some refactoring to simplify stuff
- Removed the `k8gb.coredns.extraServerBlocks` and `extra_plugins` values from the values.yaml file
- Adjusted values.schema.json

Is related to my work to allow NS entries for coredns-crd-plugin, here the [issue](https://github.com/k8gb-io/coredns-crd-plugin/issues/74).

PR #2102, opened 2025-11-06
Dynamic ZoneDelegations

k0da:
..

<details>
---

  ### Heavy tests

  ### Debug tests

</details>


kuritka (dismissed):
excellent!

ytsarev (commented):
it looks super promising but requires documentation

k0da:
@ytsarev Documentation [added](https://github.com/AbsaOSS/k8gb/blob/c252ffb40205543f12c226e9015331c6fd8705e7/docs/dynamic_zones.md)

kuritka (dismissed):

chart/k8gb/templates/coredns/cm.yaml
```diff
@@ -8,18 +8,17 @@ metadata:
...
         health
         {{- if $.Values.coredns.corefile.reload.enabled }}
         reload {{ $.Values.coredns.corefile.reload.interval }} {{ $.Values.coredns.corefile.reload.jitter }}
         {{- end }}
-{{- if .extraPlugins }}
-{{- range .extraPlugins }}
-{{ . | nindent 8 }}
-{{- end }}
-{{- end }}
+        {{- if .extraPlugins }}
```
abaguas: This PR introduces a regression. It is no longer possible to configure extraPlugins per DNSZone.
This is a feature that was introduced by the community in https://github.com/k8gb-io/k8gb/pull/1947.

Ideally the new CRD also supports such overrides. If not, at least it shouldn't break the existing functionality.

controllers/zones/zone_service.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if coreDNSZones.Data == nil {
+		coreDNSZones.Data = make(map[string]string)
+	}
+
+	zoneList := &v1beta1.ZoneDelegationList{}
+	err = zs.client.List(ctx, zoneList)
```
abaguas, line 87: Nitpick: since you also instantiate a client.Reader, shouldn't you use it also for List operations?

abaguas (changes requested):
Thank you for the work. It looks good, except for a regression I pointed out

controllers/zones/zone_service.go
k0da: Client setups caches and initializes informer. ApiReader doesn't do that. We watch to Watch ZoneDelegations but we don't want to watch Configmaps

chart/k8gb/templates/coredns/cm.yaml
k0da: updated

docs/dynamic_zones.md
```diff
@@ -0,0 +1,59 @@
...
+kind: ConfigMap
+metadata:
+  annotations:
+    meta.helm.sh/release-name: k8gb
+    meta.helm.sh/release-namespace: k8gb
+  labels:
+    app.kubernetes.io/managed-by: Helm
+  name: coredns-dynamic
+```
+This configmap is mointed into CoreDNS pod and imported with import plugin:
```
donovanmuller:
```suggestion
This configmap is mounted into CoreDNS pod and imported with import plugin:
```

donovanmuller (commented):
@k0da Is it worth adding to the docs what problem this solves or how this improves things, using `ZoneDelegation`?

ytsarev (changes requested):
@k0da, thanks a lot for adding user-facing documentation. Please also extend the PR description or additional documentation with the developer-facing design decisions. It's a serious change introducing new CRD and we need to be crystal clear why we are adding more complexity.

In line with what @donovanmuller suggested above

controllers/zones/zone_service.go
kuritka: @abaguas  - if you use client.Client before manager start, there are no caches and controller runtime gets crazy when using client (log messages).  ApiReader is clean way to access resources out of cache.

k0da:
@donovanmuller Updated documentation with pursued use case

k0da:
@ytsarev it is backed by feature flag. And be enabled on explicit request. It is not meant to change default k8gb behavior.

ytsarev:
@k0da feature flag is great, we still need more detailed  documentation for substantial changes in the codebase. We can work on the docs together when I return from Atlanta 👍

controllers/zones/zone_service.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/k8gb-io/k8gb/api/v1beta1"
+	"github.com/k8gb-io/k8gb/controllers/resolver"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	zoneCM       = "coredns-dynamic"
+	zoneTemplate = `
+%s:5353 {
```
ytsarev, line 35: probably fine, but just to double check - is it intentionally hardcoded port in the config?

controllers/zones/zone_service.go
ytsarev, line 1: do we conver this logic by unit tests?

docs/dynamic_zones.md
ytsarev, line 1: which section of `mkdocs.yml` it will show up? we need to group it somewhere. `Advanced` section?

docs/dynamic_zones.md
```diff
@@ -0,0 +1,74 @@
...
+    }
+    import ../dynamic/*.conf
+```
+
+# ZoneDelegation Status
+Holds an info about all DNSServers participating into zone delegation.
+
+# ZoneDelegation Cleanup (WIP)
+`ZoneDelegation` is protected by Finalizer, where on object removal, controller is responsible to clean up own reference in zone delegation and delete delegation completely if
+current member is the last one standing.
```
ytsarev, line 74:
```suggestion
# Dynamic Zones

## High-Level Summary

Dynamic Zones allow k8gb to serve DNS zones only when applications are deployed, solving the problem of premature `NXDOMAIN` responses in anonymous, multi-tenant, or shared clusters.

By introducing a new cluster-scoped resource, `ZoneDelegation`, k8gb dynamically detects which zones a cluster should serve and configures CoreDNS automatically.

This provides:

- Accurate DNS behavior (no premature NXDOMAIN)
- Declarative onboarding
- Automatic CoreDNS reconfiguration without restarts
- Safe cleanup when zones are removed

---

# Dynamic Zones (Feature)

The feature introduces a cluster-scoped resource `ZoneDelegation` that contains:

- `loadBalancedZone`
- `parentZone`
- `dnsZoneNegTTL`

## Why Dynamic Zones?

Pre-configuring zones in anonymous or shared clusters makes CoreDNS authoritative too early, causing `NXDOMAIN` responses. `ZoneDelegation` lets applications ship their own zone definitions, enabling dynamic activation.

## Example

````yaml
kind: ZoneDelegation
metadata:
  name: test-zone
spec:
  loadBalancedZone: test-zone.cloud.example.com
  parentZone: cloud.example.com
  dnsZoneNegTTL: 30
````

---

# Setup

Enable via Helm:

````yaml
k8gb:
  feature:
    dynamicZones: true
````

Example generated ConfigMap entry:

````yaml
apiVersion: v1
data:
  test-zone-cloud-example-com.conf: |-
    test-zone.cloud.example.com:5353 {
      import k8gbplugins
    }
kind: ConfigMap
metadata:
  name: coredns-dynamic
````

CoreDNS loads dynamic zones via:

````yaml
import ../dynamic/*.conf
````

Full Corefile snippet:

````yaml
Corefile: |-
  (k8gbplugins) {
      errors
      health
      reload 30s 15s
      ready
      prometheus 0.0.0.0:9153
      forward . /etc/resolv.conf
      k8s_crd {
          negttl 30
          loadbalance weight
      }
  }

  static-zone.cloud.example.com:5353 {
      import k8gbplugins
  }

  import ../dynamic/*.conf
````

---

# ZoneDelegation Status

The `status` field lists all DNS servers currently participating in the delegation.

---

# ZoneDelegation Cleanup (WIP)

On deletion, the finalizer:

1. Removes this cluster from the delegation  
2. Deletes the delegation if this was the last participant  
3. Removes related ConfigMap entries  
4. Lets CoreDNS reload automatically  

Ensuring no abandoned zone configuration remains.

```

controllers/zones/zone_service.go
```diff
@@ -0,0 +1,95 @@
...
+		return delegationZones, nil
+	}
+	list := []v1beta1.ZoneDelegationSpec{}
+	for _, i := range zoneList.Items {
+		zoneInfo := i.Spec
+		zoneKey := fmt.Sprintf("%s.conf", strings.ReplaceAll(zoneInfo.LoadBalancedZone, ".", "-"))
+		coreDNSZones.Data[zoneKey] = getCoreDNSData(zoneInfo.LoadBalancedZone)
+		list = append(list, zoneInfo)
+	}
+	err = zs.client.Update(ctx, coreDNSZones)
```
ytsarev, line 102: The unconditional update might lead to resource update conflicts

chart/k8gb/values.yaml
```diff
@@ -12,6 +12,11 @@ k8gb:
   # -- whether it should also deploy the gslb and dnsendpoints CRDs
   deployCrds: true
   # -- whether it should also deploy the service account, cluster role and cluster role binding
+  feature:
```
ytsarev, line 15: Why do we need an additional `feature` hierarchy for this one?
why not just `dynamicZones: false` instead of 
```
feature:
  dynamicZones: false
```
?

chart/k8gb/crd/k8gb.absa.oss_gslbs.yaml
```diff
@@ -492,3 +492,98 @@ spec:
...
       status: {}
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.19.0
+  name: zonedelegations.k8gb.absa.oss
+spec:
+  group: k8gb.absa.oss
```
ytsarev:
```suggestion
  name: zonedelegations.k8gb.io
spec:
  group: k8gb.io
```

we might consider using new upcoming group for the new resource to avoid migrations

chart/k8gb/values.yaml
```diff
@@ -12,6 +12,11 @@ k8gb:
...
+    # -- whether to use Dynamic ZoneDelegation feature.
+    # If enabled, an extra volume and VolumeMounts must be specified
```
ytsarev, line 17: Does the user need to make any modifications to the extraVolumes configuration below?

ytsarev (commented):
@k0da awesome job, thank you!

I did e2e tests locally. Main functionality works, but i noticed some things worth discussing


On fresh install , after ZoneDelegation creation, the existing gslbs are not picking up geotags/config until the main k8gb pod restart
```
 k get gslb -A --context=k3d-test-gslb1
NAMESPACE         NAME                 STRATEGY     GEOTAG
test-gslb-istio   failover-istio       failover
test-gslb-istio   notfound-istio       roundRobin
test-gslb-istio   roundrobin-istio     roundRobin
test-gslb-istio   unhealthy-istio      roundRobin
test-gslb         failover-ingress     failover
test-gslb         roundrobin-ingress   roundRobin
➜  k8gb git:(dynamic_zones) ✗ k -n k8gb delete pod k8gb-6586fb7754-nn8rc --context=k3d-test-gslb1
pod "k8gb-6586fb7754-nn8rc" deleted
➜  k8gb git:(dynamic_zones) ✗ k get gslb -A --context=k3d-test-gslb1
NAMESPACE         NAME                 STRATEGY     GEOTAG
test-gslb-istio   failover-istio       failover     eu
test-gslb-istio   notfound-istio       roundRobin   eu
test-gslb-istio   roundrobin-istio     roundRobin   eu
test-gslb-istio   unhealthy-istio      roundRobin   eu
test-gslb         failover-ingress     failover     eu
test-gslb         roundrobin-ingress   roundRobin   eu
```

I do not see anything reported in `status` of ZoneDelegation
```
k get zonedelegations.k8gb.absa.oss test-zone -o yaml
kind: ZoneDelegation
metadata:
  annotations:
      {"apiVersion":"k8gb.absa.oss/v1beta1","kind":"ZoneDelegation","metadata":{"annotations":{},"name":"test-zone"},"spec":{"dnsZoneNegTTL":30,"loadBalancedZone":"cloud.example.com","parentZone":"example.com"}}
  creationTimestamp: "2025-11-19T07:27:53Z"
  generation: 1
  name: test-zone
  resourceVersion: "1049"
  uid: e437966d-f83c-492d-9607-cc20bb7fff63
spec:
  dnsZoneNegTTL: 30
  loadBalancedZone: cloud.example.com
  parentZone: example.com
```
Please also check the inline questions/suggestions.

Thank you!

controllers/zones/zone_service.go
k0da: Yes. It is similar to this https://github.com/k8gb-io/k8gb/blob/master/chart/k8gb/templates/coredns/cm.yaml#L12

docs/dynamic_zones.md
k0da: Sound like a good idea

controllers/zones/zone_service.go
k0da: This is configmap update. We want to ensure all ZoneDelegation objects are reflected as CoreDNS zone file.

chart/k8gb/values.yaml
k0da: This is a first conditional feature we have. I thought it would be nice to introduce such things via feature Flag as those are optional.

chart/k8gb/values.yaml
k0da: Good catch, I have to drop it.

controllers/zones/zone_service.go
ytsarev: I think it's good as it is for now, I just expect issues with the concurrent updates in the future.

chart/k8gb/values.yaml
ytsarev: I think it's a noble goal to reflect that, but in practice when this functionality becomes closer to default, it will lead to migration pain to the users.

I suggest to do a single level and reflect maturity in comment. Something like
`dynamicZones: false # Alpha feature`

sudhamshk:
Hi @k0da 


I tested the Dynamic Zones feature with the following setup:

**Clusters:** `cluster1`, `cluster2`, `cluster3`

**ZoneDelegation:**
```yaml
parentZone: app.example.com
loadBalancedZone: gslb.app.example.com
```

**Application placement:**
app1 → cluster1 and cluster2
app2 → cluster2 and cluster3

With Dynamic Zones enabled, the expectation I was having was that each cluster would only serve the zones it participates in.

**Observed behaviour:**

app1.gslb.app.example.com

- resolves correctly when NS points to cluster1 or cluster2
- returns NXDOMAIN when NS points to cluster3, since cluster3 does not host the Gslb for app1


app2.gslb.app.example.com

- resolves correctly when NS points to cluster2 or cluster3
- returns NXDOMAIN when NS points to cluster1, since cluster1 does not host the Gslb for app2

I think to match how Dynamic Zones is designed, each app (or service) needs its own delegated sub-zone. That way, only the clusters hosting that app receive NS delegation, and clusters that don’t host it won’t return NXDOMAIN.

k0da:
> Hi @k0da
> 
> I tested the Dynamic Zones feature with the following setup:
> 
> **Clusters:** `cluster1`, `cluster2`, `cluster3`
> 
> **ZoneDelegation:**
> 
> ```yaml
> parentZone: app.example.com
> loadBalancedZone: gslb.app.example.com
> ```
> 
> **Application placement:** app1 → cluster1 and cluster2 app2 → cluster2 and cluster3
> 
> With Dynamic Zones enabled, the expectation I was having was that each cluster would only serve the zones it participates in.
> 
> **Observed behaviour:**
> 
> app1.gslb.app.example.com
> 
>     * resolves correctly when NS points to cluster1 or cluster2
> 
>     * returns NXDOMAIN when NS points to cluster3, since cluster3 does not host the Gslb for app1
> 
> 
> app2.gslb.app.example.com
> 
>     * resolves correctly when NS points to cluster2 or cluster3
> 
>     * returns NXDOMAIN when NS points to cluster1, since cluster1 does not host the Gslb for app2
> 
> 
> I think to match how Dynamic Zones is designed, each app (or service) needs its own delegated sub-zone. That way, only the clusters hosting that app receive NS delegation, and clusters that don’t host it won’t return NXDOMAIN.

yes. That the intent. A team who deploys app to a particular cluster brings delegation along with the app. So cluster which have ZoneDelegation have app, be healthy or unhealthy but it would have gslb resource to participate into app queries.

Our internal setup works in a way a team owns subdomain exclusively and not shared with the other teams

elohmrow:
obviated by the ZoneDelegation series, e.g., https://github.com/k8gb-io/k8gb/pull/2335

PR #2110, opened 2025-11-12
feat: Add health checking for ingress controllers (#1754)

### LINKED ISSUE #1754: GSLB does not consider gateway or Ingress controller failure in multi-cluster setup
In a multi-cluster setup, k8gb currently determines whether an IP should be added to DNS based on the health of application pods tied to a specified service. However, this does not account for the health of centralized routing components like gateways (e.g., Istio Gateway) or Ingress controllers.

During local testing, I observed that k8gb did not remove the IP for a data center where the gateway or Ingress controller was scaled down to zero. It appears that k8gb doesn’t currently consider the health of these routing components in its failover decisions.

Would it be possible to enhance k8gb to include health checks for both the gateway or Ingress controller and the application pods? This could improve the accuracy of failover responses in scenarios where the availability of these routing components directly impacts traffic distribution across clusters.

sudhamshk:
## Summary

This PR implements health checking for ingress controllers/load balancers to ensure failover occurs when the ingress layer fails, even if application pods remain healthy.

Fixes #1754

## Problem Statement

Currently, k8gb only checks the health of application pods. If the ingress controller crashes or becomes unhealthy while application pods remain healthy, k8gb continues routing traffic to that cluster, resulting in failed requests.

## Solution

Add optional health checking for the ingress controller service. When enabled, k8gb considers both application health AND ingress controller health when making failover decisions.

**Key principle**: Service is healthy ONLY if BOTH are healthy.

### Helm Values

k8gb:
  ingressController:
    healthCheck:
      enabled: true                           # Feature flag
      serviceName: "ingress-nginx-controller" # Service to monitor
      serviceNamespace: "ingress-nginx"       # Service namespace### Example Configurations

### Architecture

1. **Configuration-Based Discovery**: Users explicitly configure the ingress controller service (name + namespace)
   - No label management required
   - Works with any ingress controller
   - Clear, predictable behavior

2. **Health Check Logic** (`getLoadBalancerHealthStatus`):
   - Fetches the configured service from Kubernetes
   - Queries EndpointSlices for ready endpoints
   - Returns: `Healthy`, `Unhealthy`, or `NotFound`

3. **Failover Logic** (`getServiceHealthStatus`):
   - First checks application pod health
   - If feature enabled AND app is healthy: also checks LB health
   - Marks service unhealthy if LB is not healthy
   - Result: Both must be healthy for overall health

<details>
---

  ### Heavy tests

  ### Debug tests

</details>


controllers/gslb_controller_reconciliation.go
```diff
@@ -144,11 +144,37 @@ func (r *GslbReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 	}
 	gslb.Status.LoadBalancer.ExposedIPs = loadBalancerExposedIPs
 
+	// Get load balancer service reference from configuration
+	var lbService *k8gbv1beta1.NamespacedName
+	if r.Config.IngressControllerServiceName != "" && r.Config.IngressControllerServiceNamespace != "" {
```
abaguas: Could we get the ingress controller service dynamically from the refResolver instead of requesting the user to explicitly configure it? Did you consider this approach?

sudhamshk: Yes, dynamic discovery via refResolver was considered and even implemented initially. The idea was to auto-detect the ingress controller Service (e.g., nginx-ingress-controller) without requiring users to explicitly configure it.

However, it was later removed due to several issues:

- Different ingress technologies (NGINX, Traefik, HAProxy, Istio, Contour, etc.) expose services differently refResolver might work for some but broke for others.
- Clusters may run multiple ingress controllers (internal vs external), leading to unreliable detection.
- May add a maintenance overhead

Given the challenges mentioned, I went with an explicit configuration for now.
Happy to revisit and make it dynamic if that’s the preferred direction.

abaguas: Thank you for sharing the reasoning

> Different ingress technologies (NGINX, Traefik, HAProxy, Istio, Contour, etc.) expose services differently refResolver might work for some but broke for others.

Which cases did you find where it is not possible to identify the service? In the case of istio configuration we really have to find it, otherwise we can't find the exposed ip address.
We can also request users to label their ingress controller service in a way that it is easy for us to find it.

> Clusters may run multiple ingress controllers (internal vs external), leading to unreliable detection.

This is actually a very interesting case. The current implementation only supports a single ingress gateway. This indicates if the ingress controller is statically configured then we should configure the ingress controller selection on the GSLB, not in a global config

ytsarev:
Hey @sudhamshk , do you plan to continue working on this change? Would love to push it forward!

sudhamshk:
Hey @ytsarev 

Yes, I am interested to work on it, actually I was looking for feedback

ytsarev (commented):
I took a pass through this. The idea is useful, but I do not think it is merge-ready yet.

Main concerns:
- LB service discovery runs even when `ingressController.healthCheck.enabled=false`, so discovery failures can break reconciliation for users who have not opted in.
- "Automatic discovery" depends on a manual label plus exact Ingress/Service status address matching, which seems fragile across cloud providers.
- Gateway API currently returns no LB service, so enabling the feature silently has no effect there.
- The health check validates EndpointSlices behind the ingress-controller Service, not the actual external LB/listener/routing path.

Given the PR is stale and conflicts with current `master`, I would rebase first and tighten the feature contract before deeper review.

elohmrow:
@sudhamshk  closing this PR due to inactivity. If you are still working on this and want to bring it back, just reply here and we can reopen it!

Issue #954, opened 2022-09-20
Add support for Gateway API


somaritane:

illrill:
Eager about this one before we can start considering K8GB for our company 👍

abaguas:
In the past few months we paved the way to enable this integration. I can't promise dates, but we may release this feature around March

larssb:
@abaguas that would be so wonderfully awesome. As we've already transitioned many of our workloads being exposed outside X Kubernetes cluster to be using the Gateway API.

angelbarrera92:
+1 looking for it!

Meanwhile im trying to understand if I can use `service` instead of `ingress` in the `Gslb.spec.resourceRef` attribute. That would help transitioning from ingress to gateway api in diferent clusters.

Service Example: https://github.com/k8gb-io/k8gb/blob/master/docs/examples/lbservice-gslb-example.yaml#L32

TL;DR

cluster-a having ingress for the same workload
cluster-b having gateway-api for the same workload

abaguas:
> +1 looking for it!
> 
> Meanwhile im trying to understand if I can use `service` instead of `ingress` in the `Gslb.spec.resourceRef` attribute. That would help transitioning from ingress to gateway api in diferent clusters.
> 
> Service Example: https://github.com/k8gb-io/k8gb/blob/master/docs/examples/lbservice-gslb-example.yaml#L32
> 
> TL;DR
> 
> cluster-a having ingress for the same workload
> cluster-b having gateway-api for the same workload
> 
> 

@angelbarrera92 
I never tested this setup, but I think it will work. We identify a workload by the hostname. So you only need to make sure the GSLBs on the different clusters refer to the same hostname, independently of how the app is exposed (istio, gateway api, ingress, lb service).

angelbarrera92:
I'll test it in the upcoming days :) (hopefully)

ytsarev:
Gateway support was released in https://github.com/k8gb-io/k8gb/releases/tag/v0.17.0

PR #2114, opened 2025-11-13
Add support for GatewayAPI's HTTPRoute and GRPCRoute

abaguas:
This is a first of a series of PRs implementing support for the GatewayAPI: #954

Gateway API quick intro to understand the CRDs: https://www.youtube.com/watch?v=6m6-VrQPXyY

This PR adds the Gateway API CRDs to the local setup, and adds support for [HTTPRoutes](https://gateway-api.sigs.k8s.io/api-types/httproute/) and [GRPCRoutes](https://gateway-api.sigs.k8s.io/api-types/grpcroute/).

For k8gb, the key components of any route resource are the `parentRefs` (linking to the Gateway resource where we can fetch the exposed IP addresses), the `backendRefs` (linking to the Services backends, from where we determine the health of the application), and the hostname of the application being exposed. Here is an example using HTTPRoute:
```
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: gatewayapi-httproute
  namespace: test-gslb
spec:
  parentRefs:
  - name: gatewayapi-gateway
  hostnames:
  - gatewayapi-httproute.cloud.example.com
  rules:
  - backendRefs:
    - name: gatewayapi-service
      port: 8080
```

A Gateway looks as follows:
```
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: gatewayapi-gateway
  namespace: test-gslb
spec:
  gatewayClassName: gatewayapi-gateway
  listeners:
  - name: http
    protocol: HTTP
    port: 80
status:
  addresses:
  - type: IPAddress
    value: 10.0.0.1
```

The only slightly involved logic in this PR is in the `adapters.go` file. There, we define two interfaces: `routeSpec` and a `routeRuleSpec`. These interfaces are used to wrap the different route resources, since the fields we are interested in are defined in different structs for each Route type (GRPCRoute, HTTPRoute, ...). Our business logic can then use it to fetch `parentRefs`, `backendRefs` and `hostnames`, independently of the underlying route type.

The rest of the code is simply fetching the right field from the Custom Resources, similar to what we do with Ingress or VirtualService resources.

---

The planned course of action for follow up PRs are:
- support for TCPRoute, UDPRoute and TLSRoute
- E2E testing in chainsaw
- integration in the local setup


abaguas:
> I went through code and looks good 👍.
> 
> Please add documentation in followup PR.

Thanks! Will do!

abaguas:
> > I went through code and looks good 👍.
> > Please add documentation in followup PR.
> 
> Thanks! Will do!

As promised: #2119

PR #2121, opened 2025-11-20
fix: restore extraServerBlocks as global setting

### LINKED ISSUE #2120: `extraServerBlocks` feature broken after refactoring - location mismatch between schema and template
## Problem

The `extraServerBlocks` feature is not working due to a refactoring issue. The feature was originally added in PR #1709 (commit 518a2ce1) as a **global setting** at `k8gb.coredns.extraServerBlocks`, but was later moved to be **per-zone** in the schema (`k8gb.dnsZones[].extraServerBlocks`) without updating the template accordingly.

### Original Implementation (PR #1709 - Sept 2024)
The feature was added as a **global setting**:
- **Location**: `k8gb.coredns.extraServerBlocks` (root level)
- **Schema**: Defined in `k8gbCoredns` object
- **Template**: Accessed as `.Values.k8gb.coredns.extraServerBlocks`
- **Use case**: Add glue records for other clusters in k8gb's CoreDNS for cluster discovery

### Current State (Broken)
After refactoring to support multiple DNS zones, the location changed:

**Template** (`chart/k8gb/templates/coredns/cm.yaml` lines 38-40):
```yaml
{{- end }}
    {{- with .extraServerBlocks -}}
    {{- tpl . $ | nindent 4 }}
    {{- end }}
{{- end }}
```
This is **outside** the `{{- range .Values.k8gb.dnsZones }}` loop and tries to access `.extraServerBlocks` (which would be `.Values.extraServerBlocks` - doesn't exist).

**Schema** (`values.schema.json` lines 532-564):
Defines `extraServerBlocks` as a property **inside** the `k8gbDnsZone` definition:

```json
"k8gbDnsZone": {
    "type": "object",
    "properties": {
        "loadBalancedZone": {...},
        "parentZone": {...},
        "dnsZoneNegTTL": {...},
        "extraPlugins": {...},
        "extraServerBlocks": {
            "type": "string"
        }
    },
    ...
}
```

This causes:
1. **Schema validation errors** when trying to define it at root level: `additional properties 'extraServerBlocks' not allowed`
2. **Silent failure** when defined correctly inside DNS zones (the template can't access it from the wrong context)

## Design Analysis: Per-Zone vs Global?

After analyzing the original PR #1709, the intended design was clearly **global**:
- The use case was to add glue records for cluster discovery
- Extra server blocks define **different DNS zones** not managed by k8gb
- Having it per-zone would cause duplicate server blocks with multiple k8gb zones
- The original configuration was at `k8gb.coredns.extraServerBlocks` (global level)

During the refactoring to support multiple DNS zones (`dnsZone` → `dnsZones[]`), the setting was mistakenly moved to per-zone in the schema without updating the template.

## Steps to Reproduce

1. Add `extraServerBlocks` to a DNS zone in `values.yaml`:
   ```yaml
   k8gb:
     dnsZones:
       - parentZone: "example.com"
         extraServerBlocks: |
           test.example.com:5353 {
               forward . 8.8.8.8
           }
   ```

2. Run `helm template .`

3. Observe that the `extraServerBlocks` content is **not** included in the rendered Corefile

### Historical Timeline
1. **PR #1709 (Sept 2024)**: Feature added as `k8gb.coredns.extraServerBlocks` (global)
2. **Later refactoring**: DNS zones structure changed from single `dnsZone` to array `dnsZones[]`
3. **Migration issue**: `extraServerBlocks` was moved to per-zone in schema but template wasn't updated
4. **Current state**: Broken - neither global nor per-zone works

This suggests **global** intent - one extra server block for glue records that applies to all k8gb zones.

### Current Documentation Issues
- The README.md documents `extraServerBlocks` as being inside the `dnsZones` array (line 95)
- The default `values.yaml` shows it with a test value but provides no clear example
- No clear guidance on whether it should be global or per-zone

## Workaround

Currently there is **no workaround** - the feature is broken because:
- Cannot define at root level (schema validation fails)
- Cannot use per-zone (template doesn't access it)

Users must wait for this fix to use `extraServerBlocks`.

angelbarrera92:
Fixes #2120

## Summary

The `extraServerBlocks` feature was broken after refactoring to support multiple DNS zones. This PR restores it to its original global design as intended in PR #1709.

## Problem

The feature was originally added in PR #1709 (commit 518a2ce1) as a **global setting** at `k8gb.coredns.extraServerBlocks` to enable adding glue records for cluster discovery. During the refactoring that changed from single `dnsZone` to `dnsZones[]` array, `extraServerBlocks` was mistakenly moved to per-zone in the schema but the template was never updated, causing it to break.

## Changes

This PR restores the original global behavior:

1. **Schema** (`values.schema.json`): 
   - Moved `extraServerBlocks` from `k8gbDnsZone` definition to root-level `All` definition
   - Added proper description

2. **Template** (`templates/coredns/cm.yaml`):
   - Fixed template to access `.Values.extraServerBlocks` instead of `.extraServerBlocks`

3. **Values** (`values.yaml`):
   - Removed `extraServerBlocks` from per-zone example
   - Added it at root level with comprehensive documentation and example
   - Documented the use case (glue records, cluster discovery)

4. **Documentation** (`README.md`):
   - Updated to reflect new root-level location
   - Removed from dnsZones array documentation

## Testing

```bash
# Schema validation now passes
helm template .

# Feature works correctly with custom server blocks
helm template . --set 'extraServerBlocks=example.com:5353 { errors; forward . /etc/resolv.conf }'
```

## Use Case

As documented in the original PR #1709, this feature allows k8gb to configure glue records for other clusters in CoreDNS, enabling cluster discovery without depending on infrastructure outside the cluster.

Example:
```yaml
extraServerBlocks: |
  example.com:5353 {
    errors
    forward . /etc/resolv.conf
    k8s_crd {
      filter k8gb.absa.oss/dnstype=glue
    }
  }
```

## Breaking Changes

None - the feature was previously broken. This restores it to working state with the original intended behavior.

### DIFF
--- a/chart/k8gb/README.md
+++ b/chart/k8gb/README.md
@@ -79,6 +79,7 @@ Note: k8gb is architected to run on top of any compliant Kubernetes cluster and 
 | extdns.sources[0] | string | `"crd"` |  |
 | extdns.txtOwnerId | string | `"k8gb-<GEOTAG>"` |  |
 | extdns.txtPrefix | string | `"k8gb-<GEOTAG>-"` |  |
+| extraServerBlocks | string | `""` | Extra CoreDNS server blocks to be added after all k8gb zones. Useful for adding glue records or additional DNS zones for cluster discovery. |
 | global.imagePullSecrets | list | `[]` | Reference to one or more secrets to be used when pulling images ref: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/ |
 | infoblox.dnsView | string | `"default"` | DNS view to use for zone operations |
 | infoblox.enabled | bool | `false` | infoblox provider enabled |
@@ -92,7 +93,7 @@ Note: k8gb is architected to run on top of any compliant Kubernetes cluster and 
 | k8gb.clusterGeoTag | string | `"eu"` | Unique geotag for this K8GB instance. This tag identifies the cluster's location or role (e.g., "eu", "us-east", "dc1" or "primary"). This tag should be present in all clusters’ extGslbClustersGeoTags |
 | k8gb.deployCrds | bool | `true` | whether it should also deploy the gslb and dnsendpoints CRDs |
 | k8gb.deployRbac | bool | `true` | whether it should also deploy the service account, cluster role and cluster role binding |
-| k8gb.dnsZones | list | `[{"dnsZoneNegTTL":30,"extraPlugins":[],"extraServerBlocks":"","geoDataField":"","geoDataFilePath":"","loadBalancedZone":"cloud.example.com","parentZone":"example.com"}]` | array of dns zones controlled by gslb |
+| k8gb.dnsZones | list | `[{"dnsZoneNegTTL":30,"extraPlugins":[],"geoDataField":"","geoDataFilePath":"","loadBalancedZone":"cloud.example.com","parentZone":"example.com"}]` | array of dns zones controlled by gslb |
 | k8gb.edgeDNSServers[0] | string | `"1.1.1.1"` | use this DNS server as a main resolver to enable cross k8gb DNS based communication |
 | k8gb.exposeMetrics | bool | `false` | Exposing metrics |
 | k8gb.extGslbClustersGeoTags | string | `"eu,us"` | Comma-separated list of geotags for external K8GB clusters. These are arbitrary, user-defined identifiers (e.g., "eu,us" or "dc2,dc3") used for coordination between K8GB instances If the value remains empty, dynamic geotags extracted from the NS records on the edge DNS will be used. |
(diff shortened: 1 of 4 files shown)


abaguas:
Hi @barmettlerl, as you contributed #1947, can you explain how you are using `extraServerBlocks`?

Since an `extraServerBlock` defines a complete new domain, it makes sense that it is defined on a global level, not per dnsZone. But I would like to make sure I am not missing anything

angelbarrera92:
I was initially troubleshooting a configuration issue at the CoreDNS level. During this process, I later discovered the root cause was actually a bug within the CoreDNS plugin itself, which is tracked here: https://github.com/k8gb-io/coredns-crd-plugin/issues/106.

Although we are not currently planning to use this specific feature, I took the opportunity to fix the bug while I was already working in that area (the Boy Scout Principle).

elohmrow:
@angelbarrera92 thanks for this - is this still relevant / something you want to work on?

elohmrow:
@angelbarrera92 closing this PR due to inactivity. If you are still working on this and want to bring it back, just reply here and we can reopen it!

Issue #794, opened 2021-12-09
Add versioning to k8gb.io documentation


somaritane:
As our releasing cadence increases, and we bring more breaking changes,
it might be useful to have a versioning system for our documentation.

This might help us to keep track of what's new in each release, especially when it comes to k8gb configuration values.

P.S. This might require changing of Jekyll theme to something that supports versioning,
or changing to some other static web-page authoring framework, which has that versioning support built-in, like [Hugo](https://gohugo.io/)

Assignees:
Labels: documentation

jkremser:
there is a [docsy](https://github.com/google/docsy) hugo template used by a lot of cncf projects:
https://www.docsy.dev/docs/examples/ and also by k8s itself

somaritane:
need to check how it supports versioning

jkremser:
I think the version support is not there in the template or hugo itself.. well it can be done using the directories, like:
- `content/en/docs/v0.0.1/something.md`
- `content/en/docs/v0.0.2/something.md`

But this is not a good approach, imho. What I've seen is this:
1. have the main template for the landing page in 1 branch (be it `gh-pages` or anything else)
2. have the version specific docs stored in a separate branch using some template, like `v0.x-docs` (often this scheme corresponds with the strategy when breaking changes are introduced -> need for another parallel docs)
3. use the netlify feature for this: https://docs.netlify.com/domains-https/custom-domains/multiple-domains/#branch-subdomains together with auto builds
4. link those versioned docs from the landing page from some combo box

examples using this approach are:
- https://www.kubeflow.org/
- https://docs.k8ssandra.io/
- https://kubernetes.io/
(check the links when switching the versions)

somaritane:
Found this in docsy's documentation:
https://www.docsy.dev/docs/adding-content/versioning/

jkremser:
ah cool, yes, that looks like 4) from my description. So the combo box (/dropdown) is a built-in feature in the template, but they are not opinionated about the deployment of those versions

somaritane:
yep, so we can use branch subdomains for publishing and then refer to these subdomains

ytsarev:
@itsfarhan looks like something in your area of expertise :)

itsfarhan:
@ytsarev Yes, currently we don't have versioning in our docs. I will implement this and create a PR.

PR #2178, opened 2026-01-17
[Enhancement]: Added versioning to Mkdocs

itsfarhan:
# Versioning Mkdocs site with Mike
I have implemented versioning to our current mkdocs site.

Here are few major/minor changes done while implementing this enhancement:
- Using `mike` instead of mkdocs build. Here mike act as mkdocs to deploy and store each version as directory.
- Using `latest` for recent release tag.
- Completely modified CSS file and added comments for easy modifications in future.
- Added a banner to advertise K8GB and fixed **Star K8GB**. Previously it gave lot of trouble and broked the header and tabs structure.

Fixes: #794 

<details>
---

  ### Heavy tests

  ### Debug tests

</details>


itsfarhan:
Screenshot from 2026-01-18 16-37-16

Screenshot of running in local

.github/workflows/gh-pages.yaml
```diff
@@ -49,34 +51,36 @@ jobs:
           if git checkout gh-pages; then
             echo "Backing up existing files..."
             test -d charts && cp -r charts /tmp/preserved/
-            test -f artifacthub-repo.yml && cp artifacthub-repo.yml /tmp/preserved/
-            test -f index.yaml && cp index.yaml /tmp/preserved/
-            test -f CNAME && cp CNAME /tmp/preserved/
+            test -f artifacthub-repo.yml && cp /tmp/preserved/artifacthub-repo.yml .
+            test -f index.yaml && cp /tmp/preserved/index.yaml .
+            test -f CNAME && cp /tmp/preserved/CNAME .
```
ytsarev: Could you please elaborate on this preservation logic change? 
It's not immediately obvious to me how `artifacthub-repo.yml` , `index.yaml` and CNAME are getting copied around

.github/workflows/gh-pages.yaml
```diff
@@ -2,6 +2,8 @@ name: Github Pages
 
 on:
   push:
+    tags:
+      - 'v*'
```
ytsarev, line 6: Will it publish documentation for all existing tags or only new tags?
How do we control which documentation versions we expose and maintain?

.github/workflows/gh-pages.yaml
itsfarhan: I initially removed test logic thinking that _mike_ would handle things but later after going through more in the docs I understood that it is used instead of mkdocs build to only to establish version as alias and deploy. So i rewrote the logic. Let me revert this back.

.github/workflows/gh-pages.yaml
itsfarhan: This would use new tags only! 
In my local this is how `mike deploy --push --update-aliases v0.18.0 latest` I used for latest and to deploy previous tags `mike deploy --push --update-aliases v0.17.0`
Maybe if we can add manually in action that would deploy all versions and keep v0.18.0 as latest.??

.github/workflows/gh-pages.yaml
```diff
@@ -53,30 +55,96 @@ jobs:
...
-          # Clean everything except git files and preserved files
-          find . -maxdepth 1 -not -name '.git*' -not -name '.' -not -name '..' -exec rm -rf {} +
+          # Only deploy on tags (releases)
+          if [[ "${{ github.ref }}" == refs/tags/* ]]; then
+            VERSION=${GITHUB_REF#refs/tags/}
+            echo "Deploying version: $VERSION"
+            
+            # Version configuration
+            KEEP_LAST_N_VERSIONS=4 
+            PINNED_VERSIONS=("v0.16.0" "v0.15.0" "v0.12.0")
```
ytsarev: Could you elaborate why we are pinning exactly these versions? 

Can we also move PINNED_VERSIONS to env variable on top of the workflow for visibility?

.github/workflows/gh-pages.yaml
```diff
@@ -53,30 +55,96 @@ jobs:
...
-
-          # Clean everything except git files and preserved files
-          find . -maxdepth 1 -not -name '.git*' -not -name '.' -not -name '..' -exec rm -rf {} +
+          # Only deploy on tags (releases)
+          if [[ "${{ github.ref }}" == refs/tags/* ]]; then
+            VERSION=${GITHUB_REF#refs/tags/}
+            echo "Deploying version: $VERSION"
+            
+            # Version configuration
+            KEEP_LAST_N_VERSIONS=4
```
ytsarev: Can we move this variable to env on top of the workflow for visibility?

.github/workflows/gh-pages.yaml
```diff
@@ -53,30 +55,96 @@ jobs:
...
+            mike set-default --push latest
+            
+            # Final verification
+            mike list
+            FINAL_VERSIONS=$(mike list | grep -v latest | awk '{print $1}' | sort -V -r)
+            for version in $FINAL_VERSIONS; do
+            done            
+          else
+            echo "Skipping deployment.. only deploying tagged releases"
+          fi
```
ytsarev: It looks like this job workflow has accumulated quite a lot of logic. Can we encapsulate it into a separate `make` target and invoke it in the workflow? It would be great to be able to test it locally in isolation

.github/workflows/gh-pages.yaml
itsfarhan: These are example versions. If you're not including previous versions in sequence and want to cherry pick versions.

.github/workflows/gh-pages.yaml
itsfarhan: Sure I will put this into a script and test in local.

.github/workflows/gh-pages.yaml
```diff
@@ -53,30 +55,101 @@ jobs:
...
         run: |
-          # Copy built site to temp location
-          cp -r site /tmp/mkdocs-site
+          # Only deploy on tags (releases)
+          if [[ "${{ github.ref }}" == refs/tags/* ]]; then
+            VERSION=${GITHUB_REF#refs/tags/}
+            echo "Deploying version: $VERSION"
+            
+            # Get all git tags from v0.14.0 onwards
+            ALL_VERSIONS=$(git tag -l 'v*' | sort -V | awk '/^v0\.1[4-9]\.|^v0\.[2-9][0-9]\.|^v[1-9]\./' | sort -V)
```
ytsarev: @itsfarhan is it durable logic long term? Can you explain this regexp in the comment?

.github/workflows/gh-pages.yaml
```diff
@@ -53,30 +55,101 @@ jobs:
...
+                mike deploy --push --update-aliases $LATEST_VERSION latest
+                mike set-default --push latest
+              fi
+            fi
+            echo "Final state:"
+            mike list
+            
+          else
+            echo "Skipping deployment.. only deploying tagged releases"
+          fi
```
ytsarev: I am still concerned with putting advanced shell logic into the CI step.

Can we move it to make target that is testable locally?

itsfarhan: Sure, I will put it into makefile.

.github/workflows/gh-pages.yaml
itsfarhan: I have added regex to support anything from 0.14.0 onwards!
^v0\.1[4-9]\. - Matches versions from v0.14.x up to v0.19.x.
|^v0\.[2-9][0-9]\. - Matches versions from v0.20.x up to v0.99.x.
|^v[1-9]\. - Matches any version starting with v1.x.x through v9.x.x and above.

itsfarhan:
Screenshot from 2026-02-01 10-48-42
Screenshot from 2026-02-01 10-49-26

Tested this in local. Also added logic to move forward to other version(chronologically) if the version doesn't have mkdocs.yaml.

.github/workflows/gh-pages.yaml
ytsarev: Awesome, it clarifies. The same could go in the code comment too! 👍

itsfarhan: Sure I will add in the code comment too.

Also just incase - 
image

ytsarev (approved):

Issue #2183, opened 2026-01-20
GSLB reconciliation should not fail on hostnames outside delegated zones in referenced resources


Piroddi:
When using an HTTPRoute that contains multiple hostnames, GSLB reconciliation fails if any hostname is outside the delegated zones, even if one of the hosts is in a delegated zone/

Example HTTPRoute:

```
spec:
  hostnames:
    - test.hypersync.xyz
    - test.global.hypersync.xyz
```

If only global.hypersync.xyz has been delegated to k8gb, reconciliation fails with:

```
error="ingress host test.hypersync.xyz does not match delegated zone [global.hypersync.xyz]"
```

k8gb should filter out non-delegated hostnames and only manage the ones within its delegated zones. This allows a single HTTPRoute to serve both k8gb-managed and non-k8gb-managed hosts without requiring separate resources. If no hosts match any delegated zone, reconciliation should then still fail with a clear error (to catch misconfiguration).

One option could be to filter servers:

Filter servers after GetServers() returns in the reconciler:

https://github.com/k8gb-io/k8gb/blob/master/controllers/gslb_controller_reconciliation.go#L134

```
filteredServers := []*k8gbv1beta1.Server{}
for _, server := range servers {
    if r.Config.DelegationZones.ContainsZone(server.Host) {
        filteredServers = append(filteredServers, server)
    } else {
        log.Debug().
            Str("host", server.Host).
            Strs("delegationZones", r.Config.DelegationZones.ListZones()).
            Msg("Skipping host - does not match any delegated zone")
    }
}

if len(filteredServers) == 0 {
    return result.RequeueError(fmt.Errorf("no hosts match delegated zones %v", r.Config.DelegationZones.ListZones()))
}

gslb.Status.Servers = filteredServers
```

This approach can apply to all referenced resource types (HTTPRoute, Ingress, VirtualService)

ytsarev:
The approach sounds good to me 👍 k8gb should fail only if 0 matching hosts found

PR #2184, opened 2026-01-20
feat(gslb): Add support to filter out non delegated zones

Piroddi:
After fetching the servers from either Ingress, HttpRoute or VirtualService the gslb reconciliation will now filter only the hosts that have been delegated to k8gb.

This allows users to use a single resource to define both k8gb hosts and non k8gb hosts.

Fixes: #2183


Piroddi:
> Thanks a lot for the contribution!
> 
> High level looks great, can you please cover the impementation with unit test(s)?

Thanks @ytsarev 

Currently the changes are inline of the reconciler, so a test would require a full integration test and mocking of the  	Delegation Zones in the config.

I could move this filtering logic out into a separate function and add a unit test on the function itself?

ytsarev:
@Piroddi 
> I could move this filtering logic out into a separate function and add a unit test on the function itself?

yes, i think isolated unit test is enough for this change 👍

PR #2246, opened 2026-02-18
fix(docs): deploy mike docs on master for latest 3 tags

ytsarev:
Run docs publishing for both release-tag pushes and docs updates on master. Tag pushes keep version-specific deployment, while master pushes now deploy the latest 3 release tags so versioned docs stay updated even when no new tag is cut.

Harden mike deployment targets in the Makefile to fail fast and restore the original git ref via trap-based cleanup. Remove error-swallowing behavior and only update the `latest` alias when deploying the newest deployable version.

..

<details>
---

  ### Heavy tests

  ### Debug tests

</details>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -725,42 +725,72 @@ endef
 
 # Documentation with Mkdocs
 
-.PHONY: docs-deploy docs-list
+.PHONY: docs-deploy docs-deploy-last-3 docs-list
 
 docs-list: ## List deployed versions
 	mike list
 
-docs-deploy: ## Deploy docs with mike (requires VERSION)
+docs-deploy: ## Deploy docs with mike for VERSION (requires VERSION)
 	@if [ -z "$(VERSION)" ]; then \
 		echo "ERROR: VERSION is required"; \
 		exit 1; \
 	fi
-	@echo "Deploying documentation for $(VERSION)"
-	@ALL_VERSIONS=$$(git tag -l 'v*' | sort -V | awk '/^v0\.1[4-9]\.|^v0\.[2-9][0-9]\.|^v[1-9]\./' | sort -V); \
-	DEPLOYED_VERSIONS=$$(mike list | grep -v latest | awk '{print $$1}' | sort -V); \
-	echo "All versions from v0.14.0 onwards: $$ALL_VERSIONS"; \
-	echo "Currently deployed: $$DEPLOYED_VERSIONS"; \
-	for version in $$ALL_VERSIONS; do \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -725,42 +725,72 @@ endef
...
-		echo "$(VERSION) is not latest, keeping existing alias"; \
 	fi; \
-	mike list
\ No newline at end of file
+	mike list
+
+docs-deploy-last-3: ## Deploy docs for latest 3 release tags
+	@set -eu; \
+	git fetch --force --tags; \
+	VERSIONS=$$(git tag -l 'v*' | sort -V | awk '/^v0\.1[4-9]\.|^v0\.[2-9][0-9]\.|^v[1-9]\./' | tail -n 3); \
```
itsfarhan, line 766: awk filter will fetch versions as per regex. but do we need to exclude rc tags. I mean is it okay to deploy v0.x.x-rc1?

elohmrow:
@ytsarev when `docs-deploy-last-3` runs, it checks out each tag (v0.18.0, v0.17.0, etc.) and deploys from that tag's state. so if I fix a typo on master but haven't released v0.19.0 yet, the typo fix wouldn't appear in the redeployed v0.18.0 docs, right?

is the benefit here more about refreshing shared infra rather than propagating content changes to old versions? i don't know how mike works ...

itsfarhan:
> @ytsarev when `docs-deploy-last-3` runs, it checks out each tag (v0.18.0, v0.17.0, etc.) and deploys from that tag's state. so if I fix a typo on master but haven't released v0.19.0 yet, the typo fix wouldn't appear in the redeployed v0.18.0 docs, right?
> 
> is the benefit here more about refreshing shared infra rather than propagating content changes to old versions? i don't know how mike works ...

@elohmrow Yes, you're right. So if you fix a typo on master but haven't released v0.19.0 yet, that fix won't appear in the redeployed v0.18.0 docs because v0.18.0's tag still has the old typo.

Basically how mike works is, it builds each version into a separate directory in the gh-pages branch (like v0.18.0/, v0.17.0/). Each version's content is frozen from its tag, but the site structure regenerated each time you deploy. Basically each version deployed act as directory.

If we have to fix typos in old version docs i guess we can cherry pick and create patch tags like v0.18.1. Also the logic would still point to v0.19.0 as `latest` (alias) even though we release a patch after new release.

elohmrow:
thanks for the explanation @itsfarhan! makes sense now 👍

PR #2323, opened 2026-04-08
ZoneDelegation (3.3/4) feat: Create ZoneDelegation reconciliation

kuritka:
## Summary
This PR updates the reconciliation flow and runtime behavior for delegated zones.

It introduces `ZoneDelegation` as the single source of truth and creates these resources during bootstrap reconciliation.

## Key changes
- bootstrap `ZoneDelegation` resources during reconciliation
- preserve backward compatibility by creating them from existing `delegationZones` config
- move IP resolution into the managed reconciliation flow
- create delegated zones during initial reconciliation
- skip GSLB reconciliation until exposed IPs are available
- make `DelegationEndpoints` use `ZoneDelegation` as the source of truth

## Backward compatibility
Existing installations remain compatible, since `ZoneDelegation` can be created from the existing config derived from `DNS_ZONES` / installation `values.yaml`, for example:

```yaml
dnsZones:
  - parentZone: "example.com"
    dnsZoneNegTTL: 30
```


controllers/providers/k8gbendpoint/delegationDNSEndpoint.go
```diff
@@ -36,15 +38,15 @@ type DelegationDNSEndpoint struct {
 	client       client.Client
 	config       resolver.Config
 	logger       *zerolog.Logger
-	info         resolver.DelegationZoneInfo
+	info         zones.ZoneDelegationDetail
```
k0da: Can we be consistent and keep type name closer to what we had?

`ZoneDelegationinfo` or `ZoneDelegationStatus`

kuritka: the object ZoneDelegationinfo is extended ZoneDelegation (contains exposed IP's etc). I renamed to ExtendedZoneDelegation to be more clear.

ZoneDelegation = resource 
ExtendedZoneDelegation = resource + additional properties.

k0da: Isn't "properties" just a status? ZoneDelegationSpec should have zone/parent... and its status backing coredns servers

kuritka: It's analogy to DynamicZone and DynamicZoneInfo (first is plain configuration, second is object we internally work with).  It just wraps (extends) ZoneDelegation - single source of truth - itself.

```go
type ZoneDelegation struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`

	Spec   ZoneDelegationSpec   `json:"spec,omitempty"`
	Status ZoneDelegationStatus `json:"status,omitempty"`
}



type ExtendedZoneDelegation struct {
	wrapper           *ZoneDelegationWrapper
	LoadBalancedZone  string
	ParentZone        string
	NegativeTTL       int
	ClusterNSName     string
	ExtClusterNSNames map[string]string
	IPs               ZoneDelegationIPs
}


// GetNSServerList returns a sorted list of all NS servers for the delegation zone
func (d *ExtendedZoneDelegation) GetNSServerList() []string {
	list := []string{d.ClusterNSName}
	for _, v := range d.ExtClusterNSNames {
		list = append(list, v)
	}
	sort.Strings(list)
	return list
}

// GetExternalDNSEndpointName returns name of endpoint sitting in k8gb namespace
func (d *ExtendedZoneDelegation) GetExternalDNSEndpointName() string {
	var suffix = strings.Trim(strings.ReplaceAll(d.LoadBalancedZone, ".", "-"), " ")
	return fmt.Sprintf("k8gb-ns-extdns-%s", suffix)
}

func (d *ExtendedZoneDelegation) GetNSName(tag string) string {
	return getNsName(tag, d.LoadBalancedZone, d.ParentZone)
}

```

k0da (approved):
LGTM

PR #2327, opened 2026-04-10
ZoneDelegation (3.5/4) feat: ZoneDelegation reconciliation

kuritka:
The ZoneDelegation reconciliation ensures that the status of both newly created and existing ZoneDelegation resources is correctly updated.

As part of this process, Glue A records must be resolved. In the past, this caused excessive DNS queries; however, since the number of ZoneDelegation resources is deterministic, the implementation now minimizes DNS requests by resolving all required data efficiently.

The reconciliation logic also optimizes status updates — the ZoneDelegation status is updated only when an actual change is detected, preventing unnecessary API calls and reconcile loops.

example:

```yaml
apiVersion: k8gb.io/v1beta1
kind: ZoneDelegation
metadata:
  name: cloud-example-com.conf
spec:
  dnsZoneNegTTL: 30
  loadBalancedZone: cloud.example.com
  parentZone: example.com
status:
  dnsServers:
  - address: 172.18.0.6
    name: gslb-ns-eu-cloud.example.com
  - address: 172.18.0.7
    name: gslb-ns-eu-cloud.example.com
  - address: 172.18.0.11
    name: gslb-ns-us-cloud.example.com
  - address: 172.18.0.12
    name: gslb-ns-us-cloud.example.com
```


controllers/zones/zone_service.go
```diff
@@ -209,33 +199,56 @@ func equalStatus(a, b v1beta1.ZoneDelegationStatus) bool {
...
-
 	return true
 }
 
+func (z *ZoneDelegationImpl) buildDesiredStatus(ctx context.Context, exzd ExtendedZoneDelegation) (v1beta1.ZoneDelegationStatus, error) {
+	status := v1beta1.ZoneDelegationStatus{
+		DNSServers: make([]v1beta1.DNSServer, 0),
+	}
+
+	glueaList, err := z.ipresolver.GetGlueAInfo(ctx, exzd.LoadBalancedZone, exzd.ParentZone)
```
k0da: glueAList pls

main.go
```diff
@@ -180,6 +180,20 @@ func run() error {
...
+	}
+
+	if err = zoneDelegationReconciler.SetupWithManager(mgr); err != nil {
+		log.Err(err).Msg("Unable to create ZoneDelegation reconciler")
+		return err
+	}
+
 	if err = gslbReconciler.SetupWithManager(mgr); err != nil {
 		log.Err(err).Msg("Unable to create Gslb reconciler")
 		return err
```
k0da, line 199: x.yaml shouldn't be there

k0da (changes requested):

main.go
kuritka: oops. good catch!

k0da (approved):

PR #2333, opened 2026-04-16
feat: Pipeline doesn't rely on init-ingress but service type LoadBalancer

kuritka:
Our pipeline no longer depends on init-ingress, which previously served as one of the sources of IP addresses. In our case, it addressed an undocumented chicken-and-egg problem, where we had to wait for k8gb to obtain IP addresses to prevent failures during bootstrap. This issue no longer occurs.

We intentionally expose two IP addresses (even though a single one would be sufficient) to maintain consistency with tests that expect and validate two IP addresses.

(im enabling loadbalancer for servers and agents in us/eu regions )

### DIFF
--- a/k3d/test-gslb1.yaml
+++ b/k3d/test-gslb1.yaml
@@ -42,6 +42,11 @@ ports:
       - loadbalancer
 options:
   k3s:
+    nodeLabels:
+      - label: svccontroller.k3s.cattle.io/enablelb=true
+        nodeFilters:
+          - agent:*
+          - server:*
     extraArgs:
       - arg: --disable=traefik,metrics-server,local-storage
         nodeFilters:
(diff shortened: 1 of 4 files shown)


k3d/test-gslb1.yaml
```diff
@@ -42,6 +42,11 @@ ports:
       - loadbalancer
 options:
   k3s:
+    nodeLabels:
+      - label: svccontroller.k3s.cattle.io/enablelb=true
+        nodeFilters:
+          - agent:*
+          - server:*
```
k0da, line 49: why server?

kuritka: I wanted explicitly say, that we are using 2 IPs. If I switch into one, I'll need to do larger changes in terratest / Chainsaw, they count with 2 IPs unfortunatelly :(

kuritka: maybe I can create followup issue on that.

ytsarev (approved):
Nice! 🚀

k0da (approved):

PR #2338, opened 2026-04-24
ZoneDelegation (3.8/4) feat: use ZoneDelegations for authoritative DNS resolution

kuritka:
Main motivation of this PR is to optimize how ApplicationDNSEndpoint resolves authoritative DNS servers.

Instead of repeatedly querying edgeDNS servers for Glue A records, ApplicationDNSEndpoint now reads authoritative server IPs from existing ZoneDelegation resources. This significantly reduces the number of DNS queries sent to edgeDNS.

As part of this change:

Tests were updated to reflect the new behavior
ZoneService was extended to resolve authoritative CoreDNS endpoints (IP + GeoTag) from ZoneDelegation status

ZoneDelegation reconciliation remains responsible for periodically querying edgeDNS and updating its status with authoritative DNS server information.

The key optimization is that the number of ZoneDelegation resources is significantly lower than the number of GSLB resources. This shifts DNS resolution from a per-GSLB model to a shared, cached model, reducing load on edgeDNS.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


controllers/zone_delegation_reconciliation.go
```diff
@@ -79,7 +78,7 @@ func (r *ZoneDelegationReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return result.RequeueError(err)
 	}
 
-	return result.Stop()
+	return result.Requeue()
```
k0da, line 81: Is there default re queue interval?

k0da (dismissed):
Otherwise LGTM

controllers/zone_delegation_reconciliation.go
kuritka: sure: https://github.com/k8gb-io/k8gb/pull/2338/changes#diff-d64647b890b854d268321e363dce69112714abb482b878bd998a438a3b185ba6R51

Default is default value from envvars: https://github.com/k8gb-io/k8gb/blob/master/controllers/resolver/config.go#L44

k0da (approved):

PR #2339, opened 2026-04-26
Fix: Sync blogs across all versions

### LINKED ISSUE #2330: Blog posts invisible on landing page due to versioning strategy
We recently added a blog to the k8gb documentation site, but there's a fundamental mismatch between how docs and blog content are deployed. So basically blog posts merged to master after v0.19.0 won't be visible on the landing page until v0.20.0 is released. 

**Current Setup**
- We use mike for versioning MkDocs documentation
- Blog posts live in docs/blog/posts/
- Landing page is set to the latest stable release
- Blog is configured in mkdocs.yml

itsfarhan:
All blog posts now available across all versions immediately, not just when new release cuts

Fixes: #2330 

<details>
---

  ### Heavy tests

  ### Debug tests

</details>

### DIFF
--- a/.github/workflows/gh-pages.yaml
+++ b/.github/workflows/gh-pages.yaml
@@ -84,6 +84,14 @@ jobs:
           test -f /tmp/preserved/index.yaml && cp /tmp/preserved/index.yaml .
           test -f /tmp/preserved/CNAME && cp /tmp/preserved/CNAME .
 
+      - name: Sync blog posts to all versions
+        if: github.ref == 'refs/heads/master'
+        run: make docs-sync-blog
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Commit and push
+        run: |
           # Commit and push if there are changes
           git add .
           if ! git diff --cached --quiet; then
--- a/.github/workflows/gh-pages.yaml
+++ b/.github/workflows/gh-pages.yaml
@@ -77,6 +77,7 @@ jobs:
 
       - name: Restore preserved files
         run: |
+          git fetch origin gh-pages:gh-pages
           git checkout gh-pages
           # Restore preserved files
           test -d /tmp/preserved/charts && cp -r /tmp/preserved/charts .
--- a/.github/workflows/gh-pages.yaml
+++ b/.github/workflows/gh-pages.yaml
@@ -77,27 +77,29 @@ jobs:
 
       - name: Restore preserved files
         run: |
-          git fetch origin gh-pages:gh-pages
-          git checkout gh-pages
+          git fetch --force origin gh-pages:gh-pages
+          GHPAGES_DIR="${RUNNER_TEMP}/gh-pages"
+          echo "GHPAGES_DIR=${GHPAGES_DIR}" >> "${GITHUB_ENV}"
+          git worktree add "${GHPAGES_DIR}" gh-pages
           # Restore preserved files
-          test -d /tmp/preserved/charts && cp -r /tmp/preserved/charts .
-          test -f /tmp/preserved/artifacthub-repo.yml && cp /tmp/preserved/artifacthub-repo.yml .
-          test -f /tmp/preserved/index.yaml && cp /tmp/preserved/index.yaml .
-          test -f /tmp/preserved/CNAME && cp /tmp/preserved/CNAME .
+          for path in charts artifacthub-repo.yml index.yaml CNAME; do
+            if [ -e "/tmp/preserved/${path}" ]; then
+              cp -r "/tmp/preserved/${path}" "${GHPAGES_DIR}/"
+            fi
+          done
 
       - name: Sync blog posts to all versions
         if: github.ref == 'refs/heads/master'
         run: make docs-sync-blog
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Commit and push
         run: |
           # Commit and push if there are changes
-          git add .
-          if ! git diff --cached --quiet; then
-            git commit -m "Deploy documentation from ${{ github.sha }}"
-            git push origin gh-pages
+          git -C "${GHPAGES_DIR}" add .
+          if ! git -C "${GHPAGES_DIR}" diff --cached --quiet; then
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -797,11 +797,31 @@ endef
...
 
-.PHONY: docs-list docs-deploy docs-deploy-master docs-deploy-last-3
+.PHONY: docs-list docs-deploy docs-deploy-master docs-deploy-last-3 docs-sync-blog
 
 docs-list: ## List deployed versions
 	mike list
 
+docs-sync-blog: ## Sync blog posts to all deployed documentation versions
+	@set -eu; \
+	git fetch origin gh-pages:gh-pages --quiet; \
```
ytsarev: This target is invoked after the workflow has already checked out `gh-pages`. Fetching `origin gh-pages:gh-pages` at that point tries to update the currently checked-out branch, which Git rejects. Suggest removing this fetch/checkout from the target and making the workflow call it only while `gh-pages` is already checked out, or fetch into a temporary ref before switching branches.

Makefile
```diff
@@ -797,11 +797,31 @@ endef
...
+	if ! git checkout gh-pages 2>/dev/null; then \
+		echo "WARN: gh-pages not found, skipping blog sync"; \
+		exit 0; \
+	fi; \
+	BLOG_DIR="docs/blog"; \
```
ytsarev: After `git checkout gh-pages`, this is no longer the source tree. The published branch has rendered paths like `blog/...` and `v0.18.1/blog/...`; it does not have `docs/blog/posts`. This check makes the target a no-op in the real workflow. Suggest using the rendered root `blog/` output from the current deploy as the source, or stage the source blog content before switching to `gh-pages`.

Makefile
```diff
@@ -797,11 +797,31 @@ endef
...
+	if [ ! -d "$$BLOG_DIR/posts" ]; then \
+		echo "No blog posts to sync"; \
+		exit 0; \
+	fi; \
+	for version in */; do \
+		if [ -d "$$version/$$BLOG_DIR" ]; then \
+			echo "Syncing blog to $$version"; \
+			cp -n "$$BLOG_DIR/posts"/*.md "$$version/$$BLOG_DIR/posts/" 2>/dev/null || true; \
```
ytsarev: Copying Markdown into `v*/docs/blog/posts` will not update the served pages, archive/category pages, or search index because MkDocs has already rendered the site by this point. `cp -n` also leaves existing posts stale. Suggest replacing the rendered `v*/blog` subtree from the current rendered `blog/` output, and separately validating any assets/search data that need to move with it.

ytsarev (changes requested):
I like the overall idea, but the implementation needs changes. 

Requesting changes because the new sync target currently does not operate on the published site layout. The sync should run against the rendered `gh-pages` tree and replace generated blog output, rather than fetching the checked-out branch or copying source Markdown into version directories. See the inline comments for the concrete failure points and a safer direction.

Makefile
itsfarhan: Local gh-pages branch is not updated with the new content from mike deploy. therefore, I will be adding this line in gh-pages.yml file.

Makefile
itsfarhan: Oh right. modifying this to /blog/...

.github/workflows/gh-pages.yaml
```diff
@@ -77,13 +77,22 @@ jobs:
...
           git checkout gh-pages
           # Restore preserved files
           test -d /tmp/preserved/charts && cp -r /tmp/preserved/charts .
           test -f /tmp/preserved/artifacthub-repo.yml && cp /tmp/preserved/artifacthub-repo.yml .
           test -f /tmp/preserved/index.yaml && cp /tmp/preserved/index.yaml .
           test -f /tmp/preserved/CNAME && cp /tmp/preserved/CNAME .
 
+      - name: Sync blog posts to all versions
+        if: github.ref == 'refs/heads/master'
+        run: make docs-sync-blog
```
ytsarev, line 93: This step runs after the previous step checks out `gh-pages`. That branch does not contain the repository `Makefile`, so `make docs-sync-blog` will fail before the sync runs. Keep the sync logic available after switching branches, for example by inlining it here, calling a copied script, or running it before leaving the source checkout.

Makefile
```diff
@@ -797,11 +797,33 @@ endef
...
+docs-sync-blog: ## Sync blog posts to all deployed documentation versions
+	@set -eu; \
+	echo "Syncing blog posts across all deployed versions..."; \
+	if [ ! -d "blog" ]; then \
+		echo "ERROR: blog/ directory not found. This target must be run from gh-pages branch after deployment."; \
+		exit 1; \
+	fi; \
+	SYNCED_COUNT=0; \
+	for version_dir in v*/; do \
+		if [ -d "$$version_dir" ] && [ -d "$$version_dir/blog" ]; then \
```
ytsarev: This condition skips deployed versions that do not already have a `blog/` directory. The published tree includes versions without `blog/`, so this will not sync to all deployed versions. If every version directory should get the blog, remove this check and let the copy create the directory.

Makefile
```diff
@@ -797,11 +797,33 @@ endef
...
+			echo "Syncing blog to $$version_dir"; \
+			rm -rf "$$version_dir/blog"; \
+			cp -r blog "$$version_dir/blog"; \
```
ytsarev: The source here is top-level `blog/` on `gh-pages`, not necessarily the freshly deployed master blog. The published layout can have newer posts under `master/blog` than root `blog/`, so this may copy stale content into each version. Use the freshly built master blog as the source, or update root `blog/` before copying.

ytsarev (commented):
Review comments inline.

ytsarev (changes requested):
@itsfarhan i've made inline comments. Can you please test the changes in your fork before the next round of review? That should catch most of the issues. Thanks a lot for your contribution!

ytsarev:
Tested in my fork https://github.com/ytsarev/k8gb/actions/runs/25763886968 

https://github.com/ytsarev/k8gb/tree/gh-pages contains the blogs in the latest stable tags within the fork 

image
image

ytsarev (approved):

PR #2342, opened 2026-04-26
fix: require all services healthy for server health (#2326)

### LINKED ISSUE #2326: Service health calculation doesn't take multiple services into account
We're trying to set up K8GB to replace our custom build LB setup. However, we're running into issues when an `Ingress` maps to multiple services. For example, the default Harbor [ingress](https://github.com/goharbor/harbor-helm/blob/main/templates/ingress/ingress.yaml#L70-L126) has routes to multiple services. If one of these services is down, we would like to do a failover to our secondary Harbor instance.

K8GB determines service health in [`calculateServiceHealth`](https://github.com/k8gb-io/k8gb/blob/master/controllers/runtime_shared.go#L50-L95) and it seems it will always return a [*Healthy*](https://github.com/k8gb-io/k8gb/blob/master/controllers/runtime_shared.go#L83-L91) status if at least one service per server is up. Is this intended behavior and if so; would it be possible to control this behavior?

In the current state we can't use K8GB for a number of of-the-shelf applications because they expose multiple services in their ingresses.

itsfarhan:
When an Ingress maps to multiple Kubernetes services k8gb only checked the last service in the loop. So basically, the problem was if service A was healthy and service B was down, the server could still be marked Healthy depending on loop iteration order.

`calculateServiceHealth` now:
1. Returns \`NotFound\` if **any** service is not found
2. Returns \`Unhealthy\` if **any** service has no ready endpoints (and none are not found)
3. Returns \`Healthy\` only when **all** services have ready endpoints

Changes:

- `api/v1beta1/gslb_types.go\`: Same
- `controllers/runtime_shared.go\`: Rewrite health logic with all-service aggregation
- `controllers/runtime_shared_test.go\`: 13 unit tests covering multi-service scenarios

Closes #2326


api/v1beta1/gslb_types.go
```diff
@@ -131,6 +131,7 @@ const (
 	Healthy   HealthStatus = "Healthy"
 	Unhealthy HealthStatus = "Unhealthy"
 	NotFound  HealthStatus = "NotFound"
+	Partial   HealthStatus = "Partial"
```
ytsarev: We should not add any new features to the legacy api group

api/v1beta1io/gslb_types.go
```diff
@@ -131,6 +131,7 @@ const (
 	Healthy   HealthStatus = "Healthy"
 	Unhealthy HealthStatus = "Unhealthy"
 	NotFound  HealthStatus = "NotFound"
+	Partial   HealthStatus = "Partial"
```
ytsarev: This status is effectively not exposed anywhere

ytsarev:
@itsfarhan thanks a lot for your contribution!

I added https://github.com/k8gb-io/k8gb/pull/2342/commits/a85b9f17e5e6ad13320b797c6c62277c830ee308 to this PR:

>Introduce spec.serviceHealthPolicy for canonical k8gb.io GSLB resources so multi-service hosts can choose how backing services are evaluated.

>The default Any policy keeps the previous compatibility behavior where one healthy service makes the host healthy. The All policy requires every referenced service to exist and have ready endpoints, enabling failover for applications that expose multiple required services through one Ingress.

>Update service health calculation tests for both policies and add local deploy-full-local-setup demo resources for multiservice-gslb-all and multiservice-gslb-any so the behavior is visible in the playground.

What do you think?

controllers/runtime_shared_test.go
```diff
@@ -0,0 +1,395 @@
...
+		Spec: k8gbv1beta1io.GslbSpec{
+			ServiceHealthPolicy: serviceHealthPolicy,
+		},
+		Status: k8gbv1beta1io.GslbStatus{
+			Servers: servers,
+		},
+	}
+}
+
+func ptr[T any](v T) *T { return &v }
```
kuritka: Check if we have such function in utils package

controllers/runtime_shared_test.go
```diff
@@ -0,0 +1,395 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctx := context.Background()
+			objs := buildTestObjects(tt.services, tt.name)
+			cl := fake.NewClientBuilder().
+				WithScheme(testScheme).
+				WithObjects(objs...).
+				Build()
```
kuritka, line 304: I'm very happy with this test! 

just for your information: we have Client between our gomocks as well. I'm completelly fine with using `fake`; I just want you to know gomocks  works better in complex scenarios, and it's there.

kuritka (commented):
Good Job @itsfarhan  👍  please check my comments

ytsarev:
@kuritka thanks a lot for the review, I addressed the feedback, PTAL :)

kuritka (approved):
👍 LGTM, good job!

PR #2350, opened 2026-04-30
feat: ZoneDelegation (3.11/4) ipresolver  returns status per GlueARecord

kuritka:
This PR improves the handling of Glue A record resolution failures.

Previously, resolution would stop on the first failed Glue A record. With this change, the ipresolver returns all Glue A records, including those that cannot be resolved, and filters out only the valid ones. As a result, the ZoneDelegation status will always contain the maximum number of IPs for all alive clusters, even if some records fail.

Additionally, by preserving the status of unresolved records, this enables proper finalization of delegation zones.


controllers/zones/zone_service.go
```diff
@@ -230,9 +230,10 @@ func (z *ZoneDelegationImpl) buildDesiredStatus(ctx context.Context, exzd Extend
...
 	}
 
-	glueAList, err := z.ipresolver.GetGlueAInfo(ctx, exzd.LoadBalancedZone, exzd.ParentZone)
-	if err != nil {
-		return v1beta1.ZoneDelegationStatus{}, err
+	glueAList := z.ipresolver.GetClusterGlueAResults(ctx, exzd.LoadBalancedZone, exzd.ParentZone).FilterResolvedRecords()
+
+	if glueAList.LocalClusterError() != nil {
+		return v1beta1.ZoneDelegationStatus{}, glueAList.LocalClusterError()
 	}
```
ytsarev: `FilterResolvedRecords()` drops the local error record before `LocalClusterError()` has a chance to inspect it. That means a `GetExposedIPs` failure can be silently converted into a status update without local DNS server addresses. Keep the raw results for error handling, then filter only after local errors have been checked.

```suggestion
	glueAResults := z.ipresolver.GetClusterGlueAResults(ctx, exzd.LoadBalancedZone, exzd.ParentZone)
	if err := glueAResults.LocalClusterError(); err != nil {
		return v1beta1.ZoneDelegationStatus{}, err
	}

	glueAList := glueAResults.FilterResolvedRecords()
```

controllers/ipresolver/ipresolver.go
```diff
@@ -76,37 +70,66 @@ func (b *ResolverImpl) GetExposedIPs(ctx context.Context) (*Resolved, error) {
...
+					GeoTag:  tag,
+					IP:      glueARecord,
+					Cluster: extClusterNSName,
+					Status:  utils.DNSQueryStatusResolved,
+					IsLocal: false})
 			}
 			continue
 		}
-		gainfo = append(gainfo, &GlueAInfo{GeoTag: tag, IP: extClusterNSName, Cluster: extClusterNSName})
+		gainfo = append(gainfo, &GlueAInfo{
```
ytsarev: This fallback now treats every non-error DNS response with no A records as `Resolved` and uses the NS hostname as the `IP`. With the new `DNSQueryResult` contract, `NoAnswer`, `NXDOMAIN`, `SERVFAIL`, and `REFUSED` can all arrive with `Err == nil`; those should not be converted into resolved glue records or written into `ZoneDelegationStatus.DNSServers[].address`, which is documented as an IP address. Consider preserving `dnsResult.Status` here and only emitting resolved records when the query status is actually `DNSQueryStatusResolved`.

controllers/ipresolver/ipresolver_test.go
```diff
@@ -160,17 +174,19 @@ func TestResolveNSNames(t *testing.T) {
...
-			info, err := NewResolver(test.config, cl, qs).GetGlueAInfo(context.TODO(), test.loadBalancedZone, test.parentZone)
-			if test.expectedError {
-				assert.Nil(t, info)
-				assert.Error(t, err)
-				return
-			}
+			info := NewResolver(test.config, cl, qs).GetClusterGlueAResults(context.TODO(), test.loadBalancedZone, test.parentZone).FilterResolvedRecords()
+			err := info.LocalClusterError()
 			assert.NoError(t, err)
 			for i, glueAInfo := range test.expectedResult {
```
ytsarev: This assertion loop is currently comparing each `expectedResult` item back to itself; `info` is not used after line 177 except for `LocalClusterError()`. That makes the test pass even if `GetClusterGlueAResults()` returns the wrong records, statuses, or length. Please assert `len(info)` and compare against `info[i]` here so the new resolver behavior is actually covered.

ytsarev (commented):
Added inline comments, please consider the suggestions

controllers/ipresolver/ipresolver.go
kuritka: After testing against real DNS, the whole `append` is removed (congrats: you pointed issue line!! )

```yaml
status:
  dnsServers:
  - address: 1.2.3.4
    name: gslb-ns-eu-app.cloud.example.com
  - address: 7.8.9.10
    name: gslb-ns-us-app.cloud.example.com
```

on eu cluster: when record is not found for us,   status MUST change:

```yaml
status:
  dnsServers:
  - address: 1.2.3.4
    name: gslb-ns-eu-app.cloud.example.com    
```

consequence: the only record says - the rest of ZD on other clusters are removed. Now you can remove whole LoadBalancedZone

controllers/zones/zone_service.go
kuritka: fixed, thx

controllers/ipresolver/ipresolver_test.go
kuritka: nice one! thx, will amend fixes

kuritka:
@ytsarev fixed and thx for review. Nice cacthes!

controllers/utils/dns_query_service.go
```diff
@@ -37,19 +44,81 @@ func NewDNSQueryService() *DNSQueryServiceImpl {
...
 	dnsMsg := new(dns.Msg)
-	fqdn := fmt.Sprintf("%s.", host) // Convert to true FQDN with dot at the end
+	fqdn := dns.Fqdn(host)
 	dnsMsg.SetQuestion(fqdn, dns.TypeA)
-	dnsMsgA, err := Exchange(dnsMsg, nameservers)
-	return dnsMsgA, err
+	msg, err := Exchange(dnsMsg, nameservers)
+
+	if err != nil {
+		if isTimeout(err) {
```
ytsarev, line 54: `DNSQueryStatusTimeout` looks unreachable here. `Query()` checks `errors.As(err, &netErr)`, but the error returned by `Exchange()` is currently formatted with `%s` after all DNS servers fail, which drops the original `net.Error` type. In practice a timeout will be classified as generic `DNSQueryStatusError`. Consider preserving the original error with wrapping (`%w`) or classifying timeout before `Exchange()` replaces it.

ytsarev (dismissed):
Approving this round. The new timeout-status comment is non-blocking: the query is still treated as failed, it only loses the more specific `Timeout` classification.

Please still address the two previous review threads that remain open in the current patch:

- `zone_service.go`: local cluster errors are checked after `FilterResolvedRecords()`, so a local error record can be dropped before `LocalClusterError()` sees it.
- `ipresolver_test.go`: the assertion loop still compares `expectedResult` values back to themselves instead of comparing against `info`, so the resolver output is not actually covered.

controllers/utils/dns_query_service.go
kuritka: Good catch — you're right. `Exchange()` was formatting the final error with `%s` which dropped the original error chain and made timeout detection unreachable. I’ve switched it to `%w` so the wrapped `net.Error` is preserved and `DNSQueryStatusTimeout` can be detected correctly.

controllers/ipresolver/ipresolver_test.go
```diff
@@ -160,17 +166,19 @@ func TestResolveNSNames(t *testing.T) {
...
 				assert.Equal(t, test.expectedResult[i].IP, glueAInfo.IP)
 				assert.Equal(t, test.expectedResult[i].Cluster, glueAInfo.Cluster)
 				assert.Equal(t, test.expectedResult[i].GeoTag, glueAInfo.GeoTag)
+				if test.expectedResult[i].Err != nil {
+					assert.Error(t, glueAInfo.Err)
+				} else {
+					assert.NoError(t, glueAInfo.Err)
+				}
+				assert.Equal(t, test.expectedResult[i].Status, glueAInfo.Status)
 			}
```
ytsarev: Suggestion for the remaining test assertion issue. This checks local errors on the raw result, then compares the filtered records against the actual resolver output.

```suggestion
			results := NewResolver(test.config, cl, qs).GetClusterGlueAResults(context.TODO(), test.loadBalancedZone, test.parentZone)
			err := results.LocalClusterError()
			assert.NoError(t, err)

			info := results.FilterResolvedRecords()
			assert.Len(t, info, len(test.expectedResult))

			for i, expected := range test.expectedResult {
				actual := info[i]

				assert.Equal(t, expected.IP, actual.IP)
				assert.Equal(t, expected.Cluster, actual.Cluster)
				assert.Equal(t, expected.GeoTag, actual.GeoTag)
				assert.Equal(t, expected.Status, actual.Status)
				assert.Equal(t, expected.IsLocal, actual.IsLocal)

				if expected.Err != nil {
					assert.Error(t, actual.Err)
				} else {
					assert.NoError(t, actual.Err)
				}
			}
```

ytsarev (dismissed):
@kuritka Just last suggestion bit in testing, ones it's there ready to merge!

controllers/ipresolver/ipresolver_test.go
```diff
@@ -160,17 +166,19 @@ func TestResolveNSNames(t *testing.T) {
...
```
ytsarev, line 185: This is very close now. One remaining issue: the loop ranges over `info.FilterResolvedRecords()`, so missing records still make the test pass. Store the filtered records, assert the length, then compare expected to actual.

```suggestion
			results := NewResolver(test.config, cl, qs).GetClusterGlueAResults(context.TODO(), test.loadBalancedZone, test.parentZone)
			err := results.LocalClusterError()
			assert.NoError(t, err)

			info := results.FilterResolvedRecords()
			assert.Len(t, info, len(test.expectedResult))

			for i, glueAInfo := range info {
				assert.Equal(t, test.expectedResult[i].IP, glueAInfo.IP)
				assert.Equal(t, test.expectedResult[i].Cluster, glueAInfo.Cluster)
				assert.Equal(t, test.expectedResult[i].GeoTag, glueAInfo.GeoTag)
				if test.expectedResult[i].Err != nil {
					assert.Error(t, glueAInfo.Err)
				} else {
					assert.NoError(t, glueAInfo.Err)
				}
				assert.Equal(t, test.expectedResult[i].Status, glueAInfo.Status)
			}
```

controllers/ipresolver/ipresolver_test.go
```diff
@@ -160,17 +166,19 @@ func TestResolveNSNames(t *testing.T) {
...
-			if test.expectedError {
-				assert.Nil(t, info)
-				assert.Error(t, err)
-				return
-			}
+			info := NewResolver(test.config, cl, qs).GetClusterGlueAResults(context.TODO(), test.loadBalancedZone, test.parentZone)
+			err := info.LocalClusterError()
 			assert.NoError(t, err)
-			for i, glueAInfo := range test.expectedResult {
+			for i, glueAInfo := range info.FilterResolvedRecords() {
```
ytsarev: GitHub is rejecting the larger suggestion because it spans a changed/deleted hunk. Here is the same required test strengthening as a smaller applyable suggestion anchored only to the current `for` line:

```suggestion
			info = info.FilterResolvedRecords()
			assert.Len(t, info, len(test.expectedResult))

			for i, glueAInfo := range info {
```

ytsarev (approved):
@kuritka awesome, thanks for the great implementation and additional fixes 💪

PR #2367, opened 2026-05-20
feat: Zone delegation (3.15/4) finalization/remove zone delegation for Infoblox

kuritka:
_review after rebasing https://github.com/k8gb-io/k8gb/pull/2350_ 

This PR ensures that zone finalization is handled correctly for the Infoblox provider.

If a ZoneDelegation resource is deleted from one of the external clusters, the corresponding Glue A record is removed from Infoblox. The remaining clusters detect this change and remove the related Glue A reference from the status of their own ZoneDelegation objects. Once ZoneDelegation is removed from all external clusters, only the local cluster reference remains in the current cluster state.

If the ZoneDelegation is also removed from the local cluster, the delegated zone reference is removed from Infoblox as well. The delegated zone in Infoblox acts as a container for Glue A, NS, and SOA records, so deleting it also removes all associated objects. This ensures that Infoblox is fully cleaned up.


controllers/zones/wrapper.go
```diff
@@ -113,6 +113,13 @@ func (d *ExtendedZoneDelegation) GetNSName(tag string) string {
 	return getNsName(tag, d.LoadBalancedZone, d.ParentZone)
 }
 
+func (d *ExtendedZoneDelegation) IsLastZoneDelegationResource() bool {
+	if len(d.wrapper.zoneDelegation.Status.DNSServers) == 1 {
+		return d.wrapper.zoneDelegation.Status.DNSServers[0].Name == d.ClusterNSName
+	}
+	return false
+}
```
ytsarev, line 129: `status.dnsServers` has one entry per IP, so a single remaining cluster with two IPs never satisfies `len == 1`. Counting the remaining server names keeps the last-cluster check aligned with the status shape.

```suggestion
func (d *ExtendedZoneDelegation) IsLastZoneDelegationResource() bool {
	servers := d.wrapper.zoneDelegation.Status.DNSServers
	if len(servers) == 0 {
		return false
	}

	for _, server := range servers {
		if server.Name != d.ClusterNSName {
			return false
		}
	}

	return true
}
```

controllers/providers/dns/infoblox.go
```diff
@@ -150,18 +161,63 @@ func (p *InfobloxProvider) getZoneDelegated(o ibcl.IBObjectManager, fqdn string)
...
+	if zone == nil || zone.Ref == "" {
+		return fmt.Errorf("delegated zone %s not found or ref is empty", fqdn)
+	}
+
+	if _, err := o.DeleteZoneDelegated(zone.Ref); err != nil {
+		return err
+	}
+	m.InfobloxObserveRequestDuration(start, metrics.DeleteZoneDelegated, err == nil)
+	return nil
+}
```
ytsarev, line 202: Finalizers should be idempotent. If the delegated zone is already gone, the controller can safely remove its finalizer instead of requeueing forever.

```suggestion
func (p *InfobloxProvider) deleteZoneDelegated(o ibcl.IBObjectManager, fqdn string) error {
	start := time.Now()

	zone, err := o.GetZoneDelegated(fqdn)
	if err != nil {
		if errors.IsNotFound(err) {
			log.Info().Msgf("Delegated zone %s is already absent from edge DNS", fqdn)
			return nil
		}
		return fmt.Errorf("failed to get delegated zone %s: %w", fqdn, err)
	}
	if zone == nil {
		log.Info().Msgf("Delegated zone %s is already absent from edge DNS", fqdn)
		return nil
	}
	if zone.Ref == "" {
		return fmt.Errorf("delegated zone %s ref is empty", fqdn)
	}

	_, err = o.DeleteZoneDelegated(zone.Ref)
	if err != nil && errors.IsNotFound(err) {
		log.Info().Msgf("Delegated zone %s is already absent from edge DNS", fqdn)
		err = nil
	}
	m.InfobloxObserveRequestDuration(start, metrics.DeleteZoneDelegated, err == nil)
	if err != nil {
		return fmt.Errorf("failed to delete delegated zone %s: %w", fqdn, err)
	}
	return nil
}
```

controllers/providers/dns/infoblox.go
```diff
@@ -150,18 +161,63 @@ func (p *InfobloxProvider) getZoneDelegated(o ibcl.IBObjectManager, fqdn string)
...
+		filtered = append(filtered, ns)
+	}
+
+	zone.DelegateTo.NameServers = filtered
+	_, err = p.updateZoneDelegated(o, zone.Ref, filtered)
+	if err != nil {
+		return fmt.Errorf("failed to update delegated zone %s: %w", zoneFQDN, err)
+	}
+	return nil
+}
```
ytsarev: Partial finalization should remove all records for the local NS name. Matching on the current LB IPs can leave stale Glue A records behind after an IP rotation.

```suggestion
func (p *InfobloxProvider) removeGlueAFromDelegatedZone(o ibcl.IBObjectManager, zoneFQDN string, nsName string, _ []string) error {
	zone, err := o.GetZoneDelegated(zoneFQDN)
	if err != nil {
		if errors.IsNotFound(err) {
			log.Info().Msgf("Delegated zone %s is already absent from edge DNS", zoneFQDN)
			return nil
		}
		return fmt.Errorf("failed to get delegated zone %s: %w", zoneFQDN, err)
	}
	if zone == nil {
		log.Info().Msgf("Delegated zone %s is already absent from edge DNS", zoneFQDN)
		return nil
	}
	if zone.Ref == "" {
		return fmt.Errorf("delegated zone %s ref is empty", zoneFQDN)
	}

	filtered := make([]ibcl.NameServer, 0, len(zone.DelegateTo.NameServers))
	removed := false
	for _, ns := range zone.DelegateTo.NameServers {
		if ns.Name == nsName {
			removed = true
			continue
		}
		filtered = append(filtered, ns)
	}
	if !removed {
		return nil
	}

	if _, err = p.updateZoneDelegated(o, zone.Ref, filtered); err != nil {
		return fmt.Errorf("failed to update delegated zone %s: %w", zoneFQDN, err)
	}
	return nil
}
```

ytsarev (commented):
Added concrete implementation suggestions for the finalization issues flagged in review.

kuritka:
Thanks @ytsarev , really nice catches 👍 I’ve already applied the fixes and updated the implementation accordingly.

ytsarev (approved):
Looks ready from my side. The updated PR addresses the three issues I raised:

- Last-zone detection now handles multiple IP rows for the same NS.
- Infoblox full-zone deletion is idempotent, so finalization can complete if the zone is already gone.
- Partial finalization now removes all records for the local NS name, not just records matching the current IPs. That avoids leaving stale glue records behind after IP rotation.

No remaining blocking findings from my side.

PR #2369, opened 2026-05-22
fix: typos and wrong default INFOBLOX_WAPI_PORT in config.go

immanuwell:
Few small bugs in `controllers/resolver/config.go` that slipped through linting.

## What's broken

1. `INFOBLOX_WAPI_PORT` defaults to 53 instead of 443

The Helm chart and README both say the default WAPI port is 443 (HTTPS), but the Go fallback is 53 (DNS port). Anyone running k8gb without the Helm chart and without setting `INFOBLOX_WAPI_PORT` explicitly gets port 53, which just doesn't work.

Reproduce:
```go
// INFOBLOX_WAPI_PORT not set
cfg, _ := NewResolver().ResolveOperatorConfig()
fmt.Println(cfg.Infoblox.Port) // prints 53, should be 443
```

chart/k8gb/values.yaml and README both document `wapiPort: 443`, so this is clearly the intended default.

2. Struct tag typo: `optionl` instead of `optional`

`TracingSamplingRatio` has `optionl:""` - missing the `a`. The tag gets silently ignored by kong. All other optional fields use `optional:""` correctly.

3. Two help text typos

- `"dnsZone ad edgeDNSZone"` - "ad" should be "and"
- `"clusters gek tags"` - "gek" should be "geo"

## Changes

- `config.go`: fix struct tag typo, fix both help strings, fix default port 53 -> 443
- `resolver_test.go`: update assertion that was testing the wrong default (53 -> 443)

### DIFF
--- a/controllers/resolver/config.go
+++ b/controllers/resolver/config.go
@@ -31,7 +31,7 @@ import (
 type Config struct {
 	ClusterGeoTag string `env:"CLUSTER_GEO_TAG" required:"" validate:"geotag" help:"ClusterGeoTag to determine specific location"`
 
-	DNSZones string `env:"DNS_ZONES" required:"" help:"pairs of dnsZone ad edgeDNSZone, eg: DNS_ZONES=example.com:cloud.example.com;example.io:cloud.example.io"`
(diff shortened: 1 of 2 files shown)


elohmrow (approved):
thanks for this contribution @immanuwell! 

LGTM 🚀

ytsarev:
@immanuwell nice catch, I appreciate it!

PR #2386, opened 2026-06-17
feat: Zone delegation (3.16/4) finalization for ExternalDNS

kuritka:
Delay ZoneDelegation finalization until the last cluster removes its
delegation entry from k8gb-zone-delegation ConfigMap.

When the last ZoneDelegation is deleted, finalize it immediately and
remove the DNSEndpoint so the delegated zone is removed. Previously
delayed ZoneDelegations then observe that the finalized cluster no longer
has active DNS server IPs and complete their own cleanup together with
the DNSEndpoint removal.

Also adjust ClusterRole and Role permissions for k8gb-zone-delegation
ConfigMap to allow operator caching while keeping write access restricted.


ytsarev (commented):
Review findings with suggested patches.

controllers/zone_delegation_reconciliation.go
```diff
@@ -82,12 +85,18 @@ func (r *ZoneDelegationReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	if zone.IsInDeletion() {
 		err := r.finalize(ctx, zone)
 		if err != nil {
 			return result.RequeueError(err)
 		}
-		return result.Stop()
+		return result.Requeue()
 	}
 
 	err = r.ZoneService.UpdateStatus(ctx, zone)
```
ytsarev: [P1] `UpdateStatus` fetches and writes a separate object, so the local `zone` still has the pre-update status. The following `ExtendedZoneDelegation` therefore feeds stale or empty NS targets into the DNSEndpoint. Please return the object whose status was persisted (or mutate the input) and use it for endpoint generation.

```suggestion
	zone, err = r.ZoneService.UpdateStatus(ctx, zone)
```

controllers/providers/dns/external.go
```diff
@@ -66,13 +70,80 @@ func (p *ExternalDNSProvider) CreateZoneDelegation(zoneInfo *zones.ExtendedZoneD
...
 	return strings.ToUpper(externalDNSTypeCommon)
 }
+
+func (p *ExternalDNSProvider) deleteZoneDelegated(ctx context.Context, zone *zones.ExtendedZoneDelegation) error {
+	delegationDNSEndpoint := &externaldnsApi.DNSEndpoint{}
+	selector := types.NamespacedName{Name: zone.GetExternalDNSEndpointName(), Namespace: p.config.K8gbNamespace}
+	if err := p.client.Get(ctx, selector, delegationDNSEndpoint); err != nil {
+		return err
+	}
+	return p.client.Delete(ctx, delegationDNSEndpoint)
```
ytsarev: [P1] Finalizer cleanup must be idempotent. If the endpoint deletion succeeds but removing the ZoneDelegation finalizer fails, the retry currently receives `NotFound` here and can never finish. Treat an already absent endpoint as successful cleanup, including a delete race.

```suggestion
	if err := p.client.Get(ctx, selector, delegationDNSEndpoint); err != nil {
		return client.IgnoreNotFound(err)
	}
	return client.IgnoreNotFound(p.client.Delete(ctx, delegationDNSEndpoint))
```

chart/k8gb/templates/role.yaml
```diff
@@ -0,0 +1,23 @@
...
+    resources:
+      - configmaps
+    resourceNames:
+      - k8gb-zone-delegation
+    verbs:
+      - get
+      - update
+      - patch
+      - delete
+      - create
```
ytsarev: [P2] Kubernetes authorization cannot constrain `create` by `resourceNames`, because the object name is not available to the authorizer for create requests. The [upstream Kubernetes RBAC documentation](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources) explicitly calls out this limitation. Consequently `CreateOrUpdate` cannot recreate a missing ConfigMap with this rule. Grant namespace-scoped create separately and retain name-restricted permissions for existing objects.

```suggestion
  - apiGroups:
      - ""
    resources:
      - configmaps
    verbs:
      - create
  - apiGroups:
      - ""
    resources:
      - configmaps
    resourceNames:
      - k8gb-zone-delegation
    verbs:
      - get
      - update
      - patch
      - delete
```

kuritka: makes sense, true. Thx @ytsarev

controllers/zone_delegation_reconciliation.go
kuritka: good catch! amended 👍

ytsarev (approved):
@kuritka thank you so much for the implementation and follow-up fixes 👍

PR #2387, opened 2026-06-17
Extending ZoneDelegation documentation

kuritka:


k0da (dismissed):
LGTM

ytsarev (approved):
@kuritka thanks a lot for the great docs, i removed whitespaces and rephrased slightly confusing part

```
   114 -Use `doFinalize: false` when the delegated zone is managed externally or when k8gb should only stop serving the zone locally.
    114 +Use `doFinalize: false` when k8gb should not modify the parent DNS delegation during deletion; the controller only removes this cluster's loc
         al CoreDNS configuration.
```

PR #2396, opened 2026-06-26
fix(dns): use dot separator for localtargets DNS prefix

### LINKED ISSUE #2366: Validate localtargets DNS label length and use a separate label instead of localtargets-<host>

AruneshDwivedi:
The localtargets-<host> naming scheme could produce DNS labels exceeding the 63-character limit when the hostname's first label was long enough. Changed the separator from '-' to '.' so 'localtargets' becomes its own DNS label, keeping the original hostname labels intact.

Fixes #2366

### DIFF
--- a/controllers/providers/k8gbendpoint/applicationDNSEndpoint.go
+++ b/controllers/providers/k8gbendpoint/applicationDNSEndpoint.go
@@ -233,6 +233,18 @@ func (d *ApplicationDNSEndpoint) GetExternalTargets(host string) (targets Target
 				Msg("can't resolve FQDN using nameservers")
 			continue
 		}
+		// Fall back to legacy naming scheme for backward compatibility during migration
+		if len(d.queryService.ExtractARecords(dnsResult.Msg)) == 0 {
+			lHostLegacy, err := getLocalTargetsHostLegacy(host)
+			if err != nil {
+				continue
+			}
+			dnsResultLegacy := d.queryService.Query(lHostLegacy, nameServersToUse)
+			if dnsResultLegacy.Err != nil {
+				continue
+			}
+			dnsResult = dnsResultLegacy
+		}
 		clusterTargets := d.queryService.ExtractARecords(dnsResult.Msg)
 		if len(clusterTargets) > 0 {
 			targets[authServer.GeoTag] = &Target{clusterTargets}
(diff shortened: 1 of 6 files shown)


ytsarev (changes requested):
Thanks a lot for the contribution. Conceptually, it looks like a correct implementation but as described in #2366 we need to keep extra attention to the backward compatibility.

Most probably we will have to support both old and new notations for a time being.

AruneshDwivedi:
Closing this PR — the branch has diverged significantly from upstream and includes many unrelated changes. Will reopen with a clean branch containing only the DNS prefix fix and backward compatibility fallback.

PR #2402, opened 2026-06-29
feat: override discovered exposed IPs via CLUSTER_EXPOSED_IPS

### LINKED ISSUE #2360: Allow overriding discovered CoreDNS exposed IPs for bare-metal clusters behind static NAT
## Problem

k8gb's IP discovery assumes the CoreDNS Service / Ingress LB status reflects publicly routable IPs. On bare-metal clusters behind 1:1 static NAT — a common colo / on-prem topology — this assumption breaks:

- The cluster's CoreDNS Service (or fallback Ingress with `k8gb.io/ip-source=true`) only ever sees the **private node IPs** assigned by the cluster's L4 LB (e.g. klipper-lb on k3s, MetalLB in L2 mode, kube-vip).
- The colo perimeter does 1:1 SNAT/DNAT between private node IPs and a set of public IPs. The cluster itself has no awareness of those public IPs.
- k8gb publishes the private IPs as the NS glue records (`gslb-ns-<geoTag>-<loadBalancedZone>` A records in EdgeDNS) and as `localtargets-<host>` records inside each Gslb's DNSEndpoint.
- External resolvers follow the NS delegation, hit the private IPs in the glue, and time out.

## Reproduction

1. Stand up any bare-metal-style k8s with a node-IP-based LoadBalancer controller (k3d, kind, k3s with klipper-lb).
2. Install k8gb with `coredns.serviceType=LoadBalancer` and any EdgeDNS provider (Cloudflare / Route53).
3. Configure 1:1 NAT between node IPs and public IPs at the network perimeter (or simulate by treating node IPs as "private").
4. Observe `gslb-ns-<geoTag>-<loadBalancedZone>` A records in EdgeDNS contain node IPs.

## Why existing knobs don't solve it

| Approach                                                                     | Why it doesn't work on bare-metal NAT                                                                       |
| ---------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------- |
| `coredns.serviceType: LoadBalancer` reading `Status.LoadBalancer.Ingress[].IP`   | Bare-metal LBs assign **private** node IPs to that field                                                        |
| `Status.LoadBalancer.Ingress[].Hostname` (FQDN lookup path in `extractIPFromLB`) | klipper-lb doesn't populate Hostname; no annotation to inject one                                           |
| Ingress fallback with `k8gb.io/ip-source=true`                                 | Same root cause — bare-metal Ingress status only has private IPs                                            |
| Per-Gslb `k8gb.io/exposed-ip-addresses` annotation                             | Only overrides the **per-Gslb localtargets / A records**; does NOT fix the NS glue records published to EdgeDNS |

The last point matters: even if every Gslb annotates itself with public IPs, the NS delegation in EdgeDNS still points to private IPs, so the whole chain is broken before resolvers ever reach the cluster's CoreDNS.

## Proposed enhancement

Add a cluster-level override:

- New chart value `k8gb.edgeDNSPublicIPs` (`[]string`, default empty)
- New env var `EDGE_DNS_PUBLIC_IPS` (comma-separated)
- When set, k8gb uses these IPs in place of the discovered Service / Ingress IPs at the **zone-delegation boundary** so both the EdgeDNS NS glue records AND per-Gslb `localtargets-*` / final A records use the override.

## Suggested insertion point

`controllers/zones/wrapper.go` `ZoneDelegationWrapper.GetDetail()`. This covers both ExternalDNS and Infoblox providers, since both consume `ExtendedZoneDelegation.LocalCoreDNSExposedIPs`. Doing it here keeps the IPResolver unchanged and avoids special-casing provider code.

## Alternatives considered

- **Per-Gslb annotation only** — partial fix; doesn't reach EdgeDNS glue records.
- **External tooling to rewrite records post-publish** — fights k8gb's TXT-registry ownership and creates reconcile churn.
- **Custom CoreDNS Service hostname** — requires the bare-metal LB to support hostname annotations; klipper-lb doesn't, and patching this in-cluster is brittle.

## Use case

Bare-metal / colo BYOC clusters where the cluster sees only RFC1918 IPs but is exposed via static 1:1 NAT on the perimeter. Common in datacenter deployments where MetalLB / Cilium LB-IPAM with a public CIDR isn't an option (e.g. Cilium migration in progress, or perimeter team owns the NAT).

I have a working patch and can open a PR if the approach is acceptable.

aryasoni98:
Bare-metal / colo clusters behind 1:1 static NAT only see private node IPs in the CoreDNS Service / Ingress status. k8gb published those private IPs as the zone-delegation NS glue records (`gslb-ns-<geoTag>-<loadBalancedZone>`), so external resolvers following the delegation timed out.

This PR adds a cluster-level override: when `CLUSTER_EXPOSED_IPS` / Helm value `k8gb.clusterExposedIPs` is set, k8gb uses these operator-provided public IPs instead of discovering them from Service / Ingress status. The override covers **both**:

- ZoneDelegation NS glue / CoreDNS addresses (via `ipresolver.readIPs`), so every NS-glue consumer, including ExternalDNS and Infoblox, receives the public IPs without provider-specific code
- Application `localtargets-*` / final A records (via `refresolver.ResolveGslbExposedIPs`)

Configured IPs are expected to route DNS traffic to CoreDNS and application traffic to the cluster ingress (typically the same 1:1 NAT address with port-based forwarding). Precedence for application records:

> per-Gslb annotations (`k8gb.io/exposed-hostnames`, `k8gb.io/exposed-ip-addresses`) > `CLUSTER_EXPOSED_IPS` > discovered service/ingress IPs

If DNS and application traffic use different public IPs, keep `k8gb.clusterExposedIPs` for CoreDNS/NS glue and set application addresses with the per-Gslb annotations.

Closes #2360

<details>

---

### Heavy tests

### Debug tests

</details>


controllers/resolver/config.go
```diff
@@ -51,6 +51,8 @@ type Config struct {
 
 	CoreDNSServiceType corev1.ServiceType `env:"COREDNS_SERVICE_TYPE" required:"" enum:"ClusterIP,LoadBalancer,NodePort,ExternalName" default:"ClusterIP" help:"k8gb requires the cluster IP addresses to operate. If the service type is LoadBalancer, the IP addresses are read from the CoreDNS service otherwise, they are read from the ingress labeled with 'k8gb.io/ip-source=true'."`
 
+	EdgeDNSExposedIPs []string `env:"EDGE_DNS_EXPOSED_IPS" optional:"" validate:"omitempty,unique,dive,ip" help:"cluster-level override of the publicly routable IP addresses this cluster's CoreDNS is exposed on; comma-separated. When set, these IPs are published as the zone-delegation NS glue records instead of the IPs discovered from the CoreDNS service/ingress. Use on bare-metal/colo clusters that sit behind 1:1 static NAT and only see private node IPs. e.g: '203.0.113.10,203.0.113.11'"`
```
ytsarev: IPv6 passes this validation, but both delegation paths still publish these values as DNS A records. For example, `2001:db8::1` renders successfully into `EDGE_DNS_EXPOSED_IPS` and would later become an invalid A-record target. Until AAAA glue is supported, constrain this input to IPv4.

```suggestion
	EdgeDNSExposedIPs []string `env:"EDGE_DNS_EXPOSED_IPS" optional:"" validate:"omitempty,unique,dive,ipv4" help:"cluster-level override of the publicly routable IP addresses this cluster's CoreDNS is exposed on; comma-separated. When set, these IPs are published as the zone-delegation NS glue records instead of the IPs discovered from the CoreDNS service/ingress. Use on bare-metal/colo clusters that sit behind 1:1 static NAT and only see private node IPs. e.g: '203.0.113.10,203.0.113.11'"`
```

controllers/ipresolver/ipresolver.go
```diff
@@ -131,13 +134,28 @@ func (b *Resolved) HasIngress() bool {
...
 	return fmt.Sprintf("IPResolverService type LoadBalancer %s/%s %s", b.svc.Namespace, b.svc.Name, b.IPs)
 }
 
 func readIPs(ctx context.Context, cl client.Client, config *resolver.Config) (*Resolved, error) {
+	// Cluster-level override for bare-metal / colo clusters behind 1:1 static NAT.
+	// When the operator provides the publicly routable CoreDNS IPs explicitly, use them
+	// in place of the IPs discovered from the CoreDNS service/ingress, so the zone-delegation
+	// NS glue records published to EdgeDNS reference publicly routable addresses.
+	// See https://github.com/k8gb-io/k8gb/issues/2360
+	if len(config.EdgeDNSExposedIPs) > 0 {
```
ytsarev: This changes `ipresolver.GetExposedIPs`, which feeds zone delegation, but application targets are populated independently in `gslb_controller_reconciliation.go` via `refResolver.GetGslbExposedIPs` and then emitted as `localtargets-*` and final A records. Setting only this cluster-level override therefore still leaves application answers on the discovered private addresses.

Issue #2360 explicitly asks for the cluster override to cover both paths. Please either propagate it into Gslb target resolution—with explicit precedence relative to the per-Gslb annotations—or narrow the issue/PR contract instead of closing #2360.

ytsarev (changes requested):

aryasoni98:
Thanks for the review-both points addressed in 6da6c21d:

**1. ``IPv6 → IPv4 constraint.** Applied your suggestion: `validate:"omitempty,unique,dive,ipv4"` " Both delegation paths publish A records, so IPv6 is now rejected at config validation (added a config test covering `2001:db8::1`). Also mirrored as `"format": "ipv4" "in."

**2. Override only covered NS glue, not application records.** Fixed—the cluster override now feeds Gslb target resolution too, so `localtargets-*` / final A record uses the public IPs instead of the discovered private ones. The new `resolveGslbExposedIPs "helper" is wired into both the reconciliation and legacy controllers with explicit precedence:

> per-Gslb annotations (`k8gb.io/exposed-hostnames`, `k8gb.io/exposed-ip-addresses`) > `EDGE_DNS_EXPOSED_IPS` > discovered service/ingress IPs

So both delegation paths now agree, satisfying #2360's both-path contract, while individual Gslbs can still opt out via annotation. Added precedence unit tests and updated chart values/README``.

controllers/gslb_controller_reconciliation.go
```diff
@@ -230,6 +230,35 @@ func (r *GslbReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
+//
+// Precedence, from highest to lowest:
+//  1. Per-Gslb override annotations (k8gb.io/exposed-hostnames, k8gb.io/exposed-ip-addresses)
+//  2. Cluster-level EDGE_DNS_EXPOSED_IPS override
+//  3. IPs discovered from the CoreDNS service / ingress status
+//
+// The cluster-level override mirrors the NS-glue override applied in ipresolver, so
+// bare-metal / colo clusters behind 1:1 static NAT publish the same publicly routable
+// IPs on both delegation paths. See https://github.com/k8gb-io/k8gb/issues/2360
+func resolveGslbExposedIPs(config *resolver.Config,
```
ytsarev: This function owns application-address resolution and precedence rather than reconciliation state, and it is shared by both the current and legacy controllers. Could we move `resolveGslbExposedIPs` and `hasExposedIPOverrideAnnotation` into `controllers/refresolver/exposed_ips.go`, with the tests alongside them in `controllers/refresolver/exposed_ips_test.go`?

I would keep the zone-delegation override in `ipresolver.readIPs`: CoreDNS/NS-glue addresses and application targets are separate resolution domains even though they consume the same configured fallback. To keep dependencies narrow, the exported helper could accept `GslbReferenceResolver`, annotations, the cluster override slice, and `utils.DNSList` rather than the complete controller config.

controllers/gslb_controller_reconciliation.go
```diff
@@ -230,6 +230,35 @@ func (r *GslbReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 	return result.Requeue()
 }
 
+// resolveGslbExposedIPs returns the application target IPs (localtargets-*/final A
+// records) for a Gslb, honoring the cluster-level EDGE_DNS_EXPOSED_IPS override.
+//
+// Precedence, from highest to lowest:
+//  1. Per-Gslb override annotations (k8gb.io/exposed-hostnames, k8gb.io/exposed-ip-addresses)
+//  2. Cluster-level EDGE_DNS_EXPOSED_IPS override
+//  3. IPs discovered from the CoreDNS service / ingress status
```
ytsarev: The fallback comes from the resource referenced by the `Gslb` (Ingress, Service, Gateway, or Istio), not from the CoreDNS exposure source.

```suggestion
//  3. IPs discovered from the Gslb's referenced application resource
```

chart/k8gb/values.schema.json
```diff
@@ -422,6 +422,14 @@
...
                     }
                 },
+                "edgeDNSExposedIPs": {
+                    "type": "array",
+                    "items": {
+                        "type": "string",
+                        "minLength": 1,
+                        "format": "ipv4"
+                    }
+                },
```
ytsarev: Runtime configuration validates this slice with `unique`, but the Helm schema currently accepts duplicates. A duplicated value can therefore pass Helm validation and then prevent the controller from starting.

```suggestion
                "edgeDNSExposedIPs": {
                    "type": "array",
                    "uniqueItems": true,
                    "items": {
                        "type": "string",
                        "minLength": 1,
                        "format": "ipv4"
                    }
                },
```

ytsarev (changes requested):
Thanks for addressing the earlier IPv4 and both-path consistency findings. I’m requesting two focused follow-ups before approval: keep application exposed-IP precedence in the reference-resolution layer rather than reconciliation, and keep the Helm schema synchronized with runtime uniqueness validation. Please also update the branch onto current `master` so CI validates the final integration.

ytsarev:
Manual two-cluster e2e passed using `K8GB_LOCAL_VERSION=test make deploy-full-local-setup`:

- The cluster override propagated through both the current and legacy controllers.
- ZoneDelegation status, NS-glue DNSEndpoints, live glue, `localtargets-*`, and final A records used the configured one-IP values (EU `.6`, US `.11`).
- The per-Gslb `k8gb.io/exposed-ip-addresses` annotation correctly overrode application records while NS glue remained on the cluster override.

Both releases were restored to `edgeDNSExposedIPs: []` afterward; baseline four-address DNS behavior returned and all k8gb pods are healthy.

Local setup note: RFC2136 ExternalDNS has AXFR disabled, so stale glue values had to be removed manually before validating live EdgeDNS. The generated k8gb resources were already correct.

ytsarev:
Thanks a lot, @aryasoni98 !

The implementation looks sound, but the configuration name `k8gb.edgeDNSExposedIPs` is confusing because the value is not limited to EdgeDNS. It overrides both the ZoneDelegation NS glue/CoreDNS addresses and the application `localtargets-*` / final A records. `k8gb.clusterExposedIPs` (`CLUSTER_EXPOSED_IPS`) better reflects that cluster-wide behavior. The configured IPs are expected to route DNS traffic to CoreDNS and application traffic to the cluster ingress; per-Gslb exposed-IP/hostname annotations still take precedence for exceptions.

I pushed refactoring commit to this PR with the rename.

@kuritka, please double-check my changes here before we merge, thanks!

aryasoni98:
Thanks @ytsarev for the rename and the two-cluster e2e validation!

Agreed - `k8gb.clusterExposedIPs` / `CLUSTER_EXPOSED_IPS` is much clearer now that the override covers both ZoneDelegation NS glue and application `localtargets-*` / final A records. I also updated the PR title and description to match.

All follow-ups from the earlier review are in place:
- IPv4-only validation
- both-path override (ipresolver + refresolver)
- precedence moved into `refresolver`
- schema `uniqueItems` + `ipv4`

Happy to address anything else @kuritka finds. Thanks again!

kuritka (dismissed):

kuritka:
thx @aryasoni98 , looks good 👍  would be also nice to cover by simple chainsaw (GSLB override test).

ytsarev (dismissed):

ytsarev (approved):
@kuritka thanks for the cross-review, good feedback - I covered this feature with the chainsaw test and amended the PR 👍 

@aryasoni98 thanks a lot for this contribution!

PR #2409, opened 2026-06-30
fix: Skip GSLB reconciliation when ZoneDelegation is removed

kuritka:
Skip GSLB reconciliation when ZoneDelegation is removed instead of requeing error.

### DIFF
--- a/controllers/gslb_controller_reconciliation.go
+++ b/controllers/gslb_controller_reconciliation.go
@@ -148,10 +148,16 @@ func (r *GslbReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		m.IncrementError(gslb)
 		return result.RequeueError(fmt.Errorf("getting delegation zones (%s)", err))
 	}
-	filteredServers := filterServersByZoneDelegations(r.Logger, servers, list)
+	filteredServers := filterServersByZoneDelegations(servers, list)
 
 	if len(filteredServers) == 0 {
-		return result.RequeueError(fmt.Errorf("no hosts match delegated zones %v", list.ListZones()))
+		r.Logger.
+			Info().
+			Str("gslb", gslb.Name).
+			Str("namespace", gslb.Namespace).
+			Strs("server hosts", listHosts(servers)).
+			Msgf("Skipping delegation DNS endpoint update: ZoneDelegation is not present")
+		return result.Requeue()
 	}
 
 	gslb.Status.Servers = filteredServers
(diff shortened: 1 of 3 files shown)


controllers/gslb_controller_reconciliation.go
```diff
@@ -149,10 +149,16 @@ func (r *GslbReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
 	if len(filteredServers) == 0 {
-		return result.RequeueError(fmt.Errorf("no hosts match delegated zones %v", list.ListZones()))
+		r.Logger.
+			Info().
+			Str("gslb", gslb.Name).
+			Str("namespace", gslb.Namespace).
+			Strs("server hosts", listHosts(servers)).
+			Msgf("Skipping delegation DNS endpoint update: ZoneDelegation is not present")
+		return result.Requeue()
 	}
```
ytsarev, line 161: This condition also covers non-empty ZoneDelegation lists where no server host matches, which should remain a configuration error. Limit the successful requeue to the actual deletion/absence case.

```suggestion
	if len(filteredServers) == 0 {
		if len(list.Items) == 0 {
			r.Logger.
				Info().
				Str("gslb", gslb.Name).
				Str("namespace", gslb.Namespace).
				Strs("server hosts", listHosts(servers)).
				Msg("Skipping delegation DNS endpoint update: ZoneDelegation is not present")
			return result.Requeue()
		}
		return result.RequeueError(fmt.Errorf("no hosts match delegated zones %v", list.ListZones()))
	}
```

controllers/gslb_legacy_controller.go
```diff
@@ -119,9 +119,15 @@ func (r *LegacyGslbReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 	if len(filteredServers) == 0 {
-		return result.RequeueError(fmt.Errorf("no hosts match delegated zones %v", list.ListZones()))
+		r.Logger.
+			Info().
+			Str("gslb", gslb.Name).
+			Str("namespace", gslb.Namespace).
+			Strs("server hosts", listHosts(servers)).
+			Msgf("Skipping delegation DNS endpoint update: ZoneDelegation is not present")
+		return result.Requeue()
 	}
```
ytsarev, line 131: This condition also covers non-empty ZoneDelegation lists where no server host matches, which should remain a configuration error. Limit the successful requeue to the actual deletion/absence case.

```suggestion
	if len(filteredServers) == 0 {
		if len(list.Items) == 0 {
			r.Logger.
				Info().
				Str("gslb", gslb.Name).
				Str("namespace", gslb.Namespace).
				Strs("server hosts", listHosts(servers)).
				Msg("Skipping delegation DNS endpoint update: ZoneDelegation is not present")
			return result.Requeue()
		}
		return result.RequeueError(fmt.Errorf("no hosts match delegated zones %v", list.ListZones()))
	}
```

ytsarev (changes requested):
Please consider the condition enhancement in both controllers

controllers/gslb_controller_reconciliation.go
kuritka: Here is filtering:


```go
func filterServersByZoneDelegations(
	servers []*k8gbv1beta1io.Server,
	zoneDelegations *k8gbv1beta1io.ZoneDelegationList) []*k8gbv1beta1io.Server {
	var filtered []*k8gbv1beta1io.Server
	for _, server := range servers {
		if zoneDelegations.ContainsZone(server.Host) {
			filtered = append(filtered, server)
		}
	}
	return filtered
}

```

basically, if list.Items (ZoneDelegationList) == 0 , than filter returns empty filtered anyway.

but in terms of how it works: We are skipping rest of reconciliation if there is missing zoneDelegation for that glsb

ytsarev: You’re right. Checking `len(list.Items) == 0` would be too narrow because other ZoneDelegations may still exist while none applies to this GSLB. The filtered result is the correct condition here

ytsarev (approved):

PR #2427, opened 2026-07-13
feat: dynamic geotags refactor for ZoneDelegation

kuritka:
Dynamic geotags are available only for IBX when ext GSLB cluster geotag is empty. for more information, see: https://github.com/k8gb-io/k8gb/blob/bc4ec7cfaf91f757d1d8ae30202a969316ce8e24/docs/dynamic_geotags.md


k0da (approved):
LGTM

PR #2429, opened 2026-07-16
ci: avoid cancelled Chainsaw checks

ytsarev:
## Summary

- run Chainsaw push validation only for master and tags
- remove the duplicate-skip pre-job
- harden the actual Chainsaw job before checkout and execution

## Context

image

Same-repository pull requests trigger both push and pull_request. They previously shared the same concurrency group, so the pull request run cancelled the push run and left cancelled checks on the head commit even when Chainsaw passed.

Scoping branch pushes to master keeps pull request validation in pull_request context, including heavy-tests labels, without the false red status. Direct pushes to other branches now rely on pull request validation; tag coverage is preserved.

## Validation

- actionlint passed with the installed versions stale ubuntu-24.04 label diagnostic ignored
- workflow trigger and job structure verified with yq
- git diff --check

### DIFF
--- a/.github/workflows/chainsaw.yaml
+++ b/.github/workflows/chainsaw.yaml
@@ -2,6 +2,10 @@ name: Chainsaw
 
 on:
   push:
+    branches:
+      - master
+    tags:
+      - '**'
     paths-ignore:
       - '**.md'
       - '**.svg'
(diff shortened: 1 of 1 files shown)


kuritka (approved):
Awesome @ytsarev !

PR #2467, opened 2026-07-31
feat: refactor ExtClusterNSNames

kuritka:
in config we had `map[string]string {us: gslb-ns-us-cloud.example.com}`, in zone delegations we used `ipresolver.ClusterNSNames {gslb-ns-us-cloud-example.com: us}`.

This caused issues and complexity grew because we had multiple sources of truth for ExternalClusterGeotags I unified implementation into single palce: `resolver.ClusterNSNames(..)`, `resolver.ClusterNSNames(..).GetExtclusterGeotags` as single point of access into ClusterNsNames  and geotags etc.


k0da (approved):
🚀

PR #2470, opened 2026-08-01
refactor: simplify ResolveExternalAuthoritativeServers

kuritka:
We use a single structure to resolve the corresponding Zonedelegation from the hostname and returns the authoritative server details for the `applicationDNSEndpoint`. This PR doesnt change any functionality. It only clarifies the expected input for `applicationDNSEndpoint` targets.


kuritka:
for keeping things simple we will extend ZoneDelegation status for extra field `geoTag` or `Tag`. Than we don't need to doing extraparsing tags back from the ZD statuses

PR #2494, opened 2026-08-06
fix: restore global CoreDNS server blocks

### LINKED ISSUE #2120: `extraServerBlocks` feature broken after refactoring - location mismatch between schema and template

MikeTomlin19:
## What problem this solves

`extraServerBlocks` is currently documented under each item in
`k8gb.dnsZones`, while the CoreDNS ConfigMap reads it from the template's
root context. As a result, users cannot add custom server blocks to the
generated Corefile.

## Why this change was made

This restores the original public value path,
`k8gb.coredns.extraServerBlocks`, and renders it through `tpl` so Helm
expressions in custom blocks continue to work. It also removes the unusable
per-zone field from the chart defaults, schema, and generated documentation.

The regression was introduced while migrating to multi-zone configuration in
#1947. #2121 identified the same underlying problem but closed due to
inactivity; credit to @angelbarrera92 for that earlier investigation. This
version preserves the original value path introduced in #1709 and adds an
executable regression test.

## User impact

Users can once again append global custom server blocks with:

```yaml
k8gb:
  coredns:
    extraServerBlocks: |
      delegated.example:5353 {
        forward . 192.0.2.53
      }
```

The default remains empty, so existing Corefiles are unchanged unless the
setting is configured.

## Evidence and testing

- `make check` (license injection, Go lint, generated API/CRDs, unit tests)
- `make test-chart` with Helm 3.17.3 and Helm 4.2.3
- `helm lint chart/k8gb` with Helm 3.17.3 and Helm 4.2.3
- `values.yaml` validation against `values.schema.json`
- `actionlint .github/workflows/chainsaw.yaml`
- Pre-fix mutation verified the regression test fails when the template reads
  the old root-context `.extraServerBlocks` lookup

Fixes k8gb-io/k8gb#2120

<details>

---

By default, all checks run automatically. For website-related changes, the
Netlify bot also generates a preview.

### Heavy tests

Add the [`heavy-tests`](/k8gb-io/k8gb/issues?q=is%3A*+label%3Aheavy-tests)
label to run full-blown tests that include more than two-cluster scenarios.

### Debug tests

If the test suite fails, use `Re-run all jobs` with
[debug logging](https://docs.github.com/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging)
enabled. This also makes the
[print debug](/k8gb-io/k8gb/blob/master/.github/actions/print-terratest-debug/action.yaml)
action more verbose.

</details>


k0da:
Thanks for the PR.  Could you please elaborate on the use case you're pursuing. To me this looks like an abuse of k8gb functionality and hijack  custom serverBlocks to k8gb's server. I couldn't find a clear use case in original PRs.

We moved into ZoneDelegation which expands an ability to mange zones and decouples it from single controller configuration. Perhaps your usecase could be implemented as ZoneDelegation CR ZoneDelegation drop-in?

MikeTomlin19:
I’m not pursuing a separate deployment-specific use case; I picked this up because #2120 reports a documented feature regression. The historical use case from #1709 was adding CoreDNS glue/forwarding blocks for peer-cluster discovery without relying on external DNS.

This PR restores that behavior and adds regression coverage. If `ZoneDelegation` is now the supported way to handle that use case, I’m happy to rework or close this PR—could you clarify how the equivalent configuration should be expressed?

MikeTomlin19:
Rebased onto current master (f11e9515). Verification: go test ./... passed.

MikeTomlin19:
Thanks for the guidance. I agree that ZoneDelegation is the better direction for this use case now, so I’m closing this draft rather than restoring and maintaining a parallel global CoreDNS server-block path.

PR #2498, opened 2026-08-14
ci: add AWS Route53 E2E integration test setup

nXtCyberNet:
- Automated E2E Workflow: Added .github/workflows/aws-route53-e2e.yaml to run automated AWS Route53 integration tests on k3d clusters with concurrency locking and maintainer authorization.
- Go Integration Harness: Created route53_test.go to automate OpenTofu infrastructure provisioning, k3d cluster creation, AWS credential secret injection, and async DNS record assertions (A and NS records) with defer cleanup.
- OpenTofu Setup: Updated opentofu/main.tf with force_destroy = true and output "zone_id". Removed legacy IAM user/policy creation so existing environment credentials are used directly without requiring IAM user creation permissions (iam:CreateUser / iam:CreateAccessKey).
Makefile Target: Added make test-route53 to execute the integration test suite.


nXtCyberNet:
please also add the label - `run-aws-e2e` as i think the best setup since the integration test are not always free(because of its 12 hour billing cycle ) and needed some amount even the cost is very less in case of aws its gonna be free but for azure and gcp idont think it will gonna cost something , so its better to run these tests after the maintainers approval

PR #2500, opened 2026-08-18
feat: ZONE_DELEGATION_RECONCILE_REQUEUE_SECONDS as dedicated value fo…

kuritka:
…r ZD reconciliation

For zoneDelegation we need dedicated reconciliation requeue seconds (till now we used GSLB requeue seconds). In our environment we are fine if ZD if refreshed every 5 minutes. Also, when IPs are not yet ready for k8gb or ZD reconciliation fails, we need different intervals than 5 minutes.

For that reason we bring `ZONE_DELEGATION_RECONCILE_REQUEUE_SECONDS` and `ZONE_DELEGATION_RECOVERY_SECONDS` which distinguish between ZD  reconciliation and recovery time.

### DIFF
--- a/chart/k8gb/templates/deployment.yaml
+++ b/chart/k8gb/templates/deployment.yaml
@@ -78,6 +78,10 @@ spec:
               value: {{ include "k8gb.dnsZonesString" . }}
             - name: RECONCILE_REQUEUE_SECONDS
               value: {{ quote .Values.k8gb.reconcileRequeueSeconds }}
+            - name: ZONE_DELEGATION_RECONCILE_REQUEUE_SECONDS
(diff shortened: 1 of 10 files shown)


k0da (dismissed):

k0da (dismissed):

k0da (dismissed):

deploy/helm/stable.yaml
```diff
@@ -1,5 +1,6 @@
 k8gb:
   reconcileRequeueSeconds: 10
+  zoneDelegationReconcileRequeueSeconds: 20
```
k0da: I think that should go to `next`

k0da (approved):

